PR #4, opened 2021-07-08
Fill in missing ffi parts, and stub transitive dep

ribasushi:
grep filecoin-ffi go.sum | wc -l
       0


whyrusleeping (approved):
fantastic. Thank you!

PR #5, opened 2021-07-08
Feat/port lotus dm selector retrieval 6393 to estuary

ribasushi:
@whyrusleeping needs more work on the edges, but you should be able to get it from here


filclient/filclient.go
```diff
@@ -760,7 +761,7 @@ func (fc *FilClient) CheckOngoingTransfer(ctx context.Context, miner address.Add
 
 }
 
-func (fc *FilClient) RetrievalQuery(ctx context.Context, maddr address.Address, pcid cid.Cid) (*retrievalmarket.QueryResponse, error) {
+func (fc *FilClient) RetrievalQuery(ctx context.Context, maddr address.Address, pcid cid.Cid, optionalSelector ipld.Node) (*retrievalmarket.QueryResponse, error) {
```
whyrusleeping, line 764: is this even used?

replication.go
```diff
@@ -1842,47 +1852,65 @@ func (cm *ContentManager) retrieveContent(ctx context.Context, contentToFetch ui
...
 
-	if err := cm.runRetrieval(ctx, contentToFetch); err != nil {
+	contentRoot, err := cm.runRetrieval(ctx, contentToFetch)
+	if err != nil {
 		prog.endErr = err
 		return err
 	}
 
+	// FIXME need to do something with this
+	panic(contentRoot)
```
whyrusleeping, line 1862: eh?

ribasushi: When you do a partial retrieval - you get a new CID for your sub-dag. The client doesn't know what it is, just the path to it. The code pulls it out for you by re-running the selector, and then you do... something with it.

filclient/filclient.go
ribasushi: Not at present: https://github.com/filecoin-project/go-fil-markets/blob/32e5ccef50d88d8a19b79f71ea98157c4a0047d2/retrievalmarket/types.go#L163

But in the future they very well might be able to respond intelligently, so I wired it through into the signature.

replication.go
ribasushi: @whyrusleeping an alternative would be to run the selector-over-your BS lazily, when the user requests their data back. Then changing the functions args won't be needed.

It is currently done this way to minimize the amount of time for GC ( since you can't pin the partial dag )...

Dunno - thoughts welcome.

ribasushi:
@elijaharita that last commit wires in https://github.com/application-research/estuary/pull/5/commits/c3292cbafbe455a8acefd1bedb2c6e5888bec811#diff-101af5442d7541ffa58c8cda4f725ff032482209abe78f90545af14996457020R347-R350

You can use it like so: https://github.com/filecoin-project/lotus/pull/6393#issue-661783290

ribasushi:
@whyrusleeping this bad boy is brought up to master again. Do you want to merge it, `FIXME`s included, or wanna think more?

ribasushi:
I will keep this out until we land the graphsync upgrade in lotus, to minimize churn. Will reopen in couple weeks when the time is right.

PR #9, opened 2021-07-16
full event-based rewrite of `FilClient.RetrieveContent`

elijaharita:
This is a pretty big update, and I don't think I missed anything but there's always the possibility, so probably a good idea to check it well, and test it out! I'm quite happy with the new retrieval progress logging because it doesn't clog up the terminal, which was becoming an issue for me when I went back to check previous runs during testing. Of course, it's a change from the previous behavior, so let me know if there are any objections.

Unfortunately, this update does not fix the previously encountered issue with getting stuck in the responder paused state for some files. However, I'm not sure if the other, original issue has changed at all. Need to do more testing.


filclient/filclient.go
```diff
@@ -853,221 +857,113 @@ func (fc *FilClient) RetrieveContent(ctx context.Context, miner address.Address,
...
-	// NB: data transfer will propose the retrieval, and if the miner accepts
-	// it, will start sending data via graphsync to us. This happens behind the
-	// scenes, and we dont get much introspection into this
+	defer unsubscribe()
 
-	// Now, we poll the data transfer channel status and wait until it says its accepted
-	if err := fc.waitForDealAccepted(ctx, chanid); err != nil {
-		return nil, err
+	// Wait for the retrieval to finish before exiting the function
+	if err := <-dtRes; err != nil {
```
whyrusleeping, line 946: probably want to respect the context here as well, so putting this into a select like:
```go
select {
  case err := <-dtRes:
    if err != nil {
      return err
    }
  case <-ctx.Done():
    return ctx.Err()
}
```


That way, if the caller cancels their request, this exits more quickly

filclient/filclient.go
```diff
@@ -853,221 +857,113 @@ func (fc *FilClient) RetrieveContent(ctx context.Context, miner address.Address,
...
+				case retrievalmarket.DealStatusFundsNeededUnseal:
+					dtRes <- xerrors.Errorf("received unexpected payment request for unsealing data")
+				default:
+					log.Debugf("unrecognized voucher: %v", resType)
+				}
+			default:
+				log.Error("unrecognized voucher response type")
+			}
+		case datatransfer.DataReceived:
+			if time.Since(lastReceivedUpdate) >= time.Millisecond*100 {
```
whyrusleeping, line 934: Honestly, we could probably just take this opportunity to remove the progress printing, or maybe have some way of configuring it?

filclient/filclient.go
```diff
@@ -853,221 +857,113 @@ func (fc *FilClient) RetrieveContent(ctx context.Context, miner address.Address,
...
+				fmt.Printf("received: %v\r", state.Received())
+				lastReceivedUpdate = time.Now()
+			}
+		case datatransfer.FinishTransfer:
```
whyrusleeping, line 938: Probably worth having a 'default' case to this outer switch statement , not sure what else could happen here, but i figure it might be worth paying attention to.

filclient/filclient.go
```diff
@@ -853,221 +857,113 @@ func (fc *FilClient) RetrieveContent(ctx context.Context, miner address.Address,
...
+
+	// The next nonce (incrementing unique ID starting from 0) for the next voucher
+	var nonce uint64 = 0
+
+	// Set up incoming events handler
+
+	// dtRes receives either an error (failure) or nil (success) which is waited
+	// on and handled below before exiting the function
+	dtRes := make(chan error)
+	unsubscribe := fc.dataTransfer.SubscribeToEvents(func(event datatransfer.Event, state datatransfer.ChannelState) {
```
whyrusleeping, line 887: So, part of the reason I initially didnt use the subscribe code was because it sends events for *all* data transfers in progress. For the filc case, we dont notice any issues, but in the case of estuary this will be clogged with thousands of other things going on all at the same time.

To get around this, I think we need to filter handling events based on `state.ChannelID == chanid`.

whyrusleeping: Looking at this a bit more, I think theres a race condition (which absolutely exists in the current code too, but its not clear if its actually an issue) where some events for our transfer might fire after we call `OpenPullDataChannel` but before we call subscribe.

This is a little tricky, since in order to do the filtering i mentioned, we need the channel id, so we cant call subscribe first.
Theres no atomic way to subscribe and open a new channel at the same time... How does go-fil-markets do this?

filclient/filclient.go
```diff
@@ -853,221 +857,113 @@ func (fc *FilClient) RetrieveContent(ctx context.Context, miner address.Address,
...
+
+					if err := fc.dataTransfer.SendVoucher(ctx, chanid, &retrievalmarket.DealPayment{
+						ID:             proposal.ID,
+						PaymentChannel: pchAddr,
+						PaymentVoucher: vres.Voucher,
+					}); err != nil {
+						dtRes <- xerrors.Errorf("failed to send payment voucher: %w", err)
+					}
+
+					nonce++
```
whyrusleeping, line 924: Are calls to this event handler potentially happening in parallel to eachother? Its probably not an issue in practice, but we could have data consistency issues if we ran this event code twice in parallel

filclient/filclient.go
elijaharita: Hmm, that's tricky, I wasn't aware. That could be a solution, I'll think about other ways we could solve this more effectively. Might have to come next week.

filclient/filclient.go
elijaharita: You mean pushing to dtRes? Or the entire callback in SubscribeToEvents?

filclient/filclient.go
elijaharita: the progress printing actually doesn't show up in logs after the fact, because the message gets overwritten by the next line in the console. i agree with you that it's not ideal for a real program like estuary, although it is very helpful to have at the moment. it shouldn't have any sort of performance penalty, either. anyways, that's my argument. i'll execute whatever you decide on.

filclient/filclient.go
elijaharita: added

filclient/filclient.go
whyrusleeping: it does end up in the logs when you pipe the logs to a file (the overwriting effect is just a terminal thing, the bytes are still written out).

How about we make it a config field in the filclient struct? just a simple "RetrivalProgressLogging" boolean to toggle it on or off?

filclient/filclient.go
whyrusleeping: the entire callback really, im just looking at this nonce incrementing and thinking that we could end up with things out of order, or having the wrong values if we (somehow) end up doing these in parallel.

filclient/filclient.go
elijaharita: smart idea

filclient/filclient.go
elijaharita: is this still a problem? i'm not sure quite exactly what you're getting at. the whole callback is guarded by a mutex now, does that help?

filclient/filclient.go
```diff
@@ -853,221 +860,137 @@ func (fc *FilClient) RetrieveContent(ctx context.Context, miner address.Address,
...
-	for {
-		// Now that the deal has been accepted, we sit around and wait for data
-		// to come in and for the miner to ask for more money
-		amt, nvc, done, err := fc.waitForPaymentNeeded(ctx, chanid, voucherCount)
-		if err != nil {
-			return nil, xerrors.Errorf("waitForPayment needed failed: %w", err)
-		}
+	unsubscribe := fc.dataTransfer.SubscribeToEvents(func(event datatransfer.Event, state datatransfer.ChannelState) {
+		// Lock chanid for the entire callback
+		chanidMu.Lock()
```
whyrusleeping, line 890: instead of locking around the entire function, we could just lock around checking if this is the right channel. Something like:
```go
chanidMu.Lock()
match := state.ChannelID() == chanid
chanidMu.Unlock()

if !match {
  return
}
...
```

filclient/filclient.go
```diff
@@ -853,221 +860,137 @@ func (fc *FilClient) RetrieveContent(ctx context.Context, miner address.Address,
...
+						Lane:        pchLane,
+						Nonce:       nonce,
+						Amount:      totalPayment,
+					})
+					if err != nil {
+						dtRes <- err
+					}
+
+					if types.BigCmp(vres.Shortfall, big.NewInt(0)) > 0 {
+						dtRes <- xerrors.Errorf("not enough funds remaining in payment channel (shortfall = %s)", vres.Shortfall)
```
whyrusleeping, line 923: on these error channel sends, there are a few things to consider. First is a thing Ive caught myself doing so many times, you replaced a return error with a channel send, and now we are no longer returning from the function (meaning that here, SendVoucher will still get called).

Second, if we get one error, then we move on in the rest of our code, what happens when the next event comes in and another error happens? It will send on the same channel, but nobody will be receiving on the other end, and this will block. I think that these callbacks happen in their own goroutine anyways, so it wont likely cause a deadlock, but it is likely to cause a resource leak, with a bunch of goroutines hanging on error sends for a long time. To get around this, we can use a select statement to wait for multiple conditions, like so:
```go
select {
case dtRes <- err:
  return
case <-doneChannel:
  return
}
```

That way, if we are 'done' we dont block forever. To implement that done channel, we can use `ctx.Done()` on our functions context, and do some cancellation on it at the top of the function:

(either before or after the tracer injection):
```go
ctx, cancel := context.WithCancel(ctx)
defer cancel()
```

whyrusleeping: To make that cleaner, we could turn it into a helper function to wrap around the select statement:
```go
sendErr := func(e error) {
  select {
  case dtRes <- e:
  case <-ctx.Done():
  }
}
```

And then just use it like:
```
if err != nil {
  sendErr(err)
  return
}
```

filclient/filclient.go
elijaharita: chanid actually gets used later in the function, so i can't keep the locked section this short. i alternatively considered copying chanid as chanidLocked or something along those lines and then replacing uses of chanid later in the function with that, but i realized that it's not really useful because the only time outside of this function that the mutex gets locked is to set chanid initially. 

should i go back to that way?

filclient/filclient.go
elijaharita: how about something like this?

```
finish := func(err error) {
  select {
  case dtRes <- err:
  default:
  }
}
```

and then, just

```
if err != nil {
  finish(err)
  return
}
```

that way it would try to push to the channel, or if it couldn't, just do nothing

filclient/filclient.go
whyrusleeping: Ah, good point about using it later down. Taking a copy under the lock should work well enough then. Its an annoying but technically correct approach.

filclient/filclient.go
whyrusleeping: Yeah, That works too, but you need to make sure you buffer the channel so you never accidentally hit the default.
```go
dtRes := make(chan error)
```

elijaharita: ok, will do. how does this change the behavior, if i may ask? i tried doing some research just now but i don't think i get what you're talking about here.

whyrusleeping: Ah, so this is a fun one.

Channels are unbuffered by default, which means that sends will block until the other side is receiving synchronously. In the unbuffered mode, if the other side is not receiving at the exact moment of the send, a default case in a select statement would be taken. even if the receive happens a millisecond later, the case evaluation happens as we enter the select.

Now, with a buffered channel, values sent can just go into the buffer (up to the number specified) and can be received whenever. A send into a buffered channel with space *always* happens immediately, ensuring we never race to take the default path.

whyrusleeping (approved):
LGTM, Thank you!

PR #12, opened 2021-07-22
`pinStatuses`

elijaharita:
I think this should work. But I haven't tested it yet.


pinning.go
```diff
@@ -21,47 +21,79 @@ import (
...
+	for i, id := range contentIDs {
+		cm.pinLk.Lock()
+		operation, ok := cm.pinJobs[id]
+		cm.pinLk.Unlock()
+
+		// If the status already exists in memory, use that
+		if ok {
+			statuses[i] = operation.PinStatus()
+		} else {
+			// Otherwise, find the content result from the database query...
```
whyrusleeping, line 45: We could just sort the list and do it that way, Will the caller be surprised if the response order is different than the input?

elijaharita: what do you mean @whyrusleeping?

pinning.go
```diff
@@ -375,18 +402,29 @@ func (s *Server) handleListPins(e echo.Context, u *User) error {
 
 	}
 
+	var contents []Content
+	if err := q.Scan(&contents).Error; err != nil {
+		return err
+	}
+
+	contentIDs := make([]uint, len(contents))
```
whyrusleeping, line 410: oh hey, we already have the slice of contents. Maybe we just make pinStatus accept a Content type? that way we dont have to go load it from the db again?

whyrusleeping (approved):
Very clean, that looks like its gonna make things a wee bit faster.

PR #13, opened 2021-07-23
`estuary setup`

### LINKED ISSUE #11: Estuary should help the host with creating an initial admin account

elijaharita:
Adds `estuary setup` which sets up an initial auth token. 

Closes #11

### DIFF
--- a/main.go
+++ b/main.go
@@ -207,7 +207,8 @@ func main() {
 	}
 	app.Commands = []*cli.Command{
 		{
-			Name: "setup",
+			Name:  "setup",
+			Usage: "Creates an initial auth token under new user \"admin\"",
 			Action: func(cctx *cli.Context) error {
 				db, err := setupDatabase(cctx)
 				if err != nil {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -204,6 +205,48 @@ func main() {
...
+					PassHash: passHash,
+					Perm:     100,
+				}
+				if err := db.Create(newUser).Error; err != nil {
+					return fmt.Errorf("admin user creation failed: %w", err)
+				}
+
+				authToken := &AuthToken{
+					Token:  "EST" + uuid.New().String() + "ARY",
+					User:   0,
```
whyrusleeping, line 236: 0 should be `newUser.ID` instead.

main.go
```diff
@@ -204,6 +205,48 @@ func main() {
...
+				}
+				if err := db.Create(authToken).Error; err != nil {
+					return fmt.Errorf("admin token creation failed: %w", err)
+				}
+
+				fmt.Printf("Auth Token: %v", authToken.Token)
+
+				return nil
+			},
+			Usage: "Creates an initial auth token under new user \"admin\"",
```
whyrusleeping, line 247: nit: put the usage above the action function definition

elijaharita: LOL whoops, that's not a nit at all, i committed a crime

main.go
elijaharita: ok, fixes applied

whyrusleeping (approved):

whyrusleeping:
Great, this makes setting up your own estuary deployment a good deal easier!

PR #20, opened 2021-08-13
Update filclient and apply new error handling changes

elijaharita:
Updates Filclient to the latest version and replaces uses of `filclient.ApiError` with the new generic `fliclient.Error`


whyrusleeping (approved):
LGTM, maybe we make helper functions for this in the future, but good enough for now

PR #25, opened 2021-08-28
Database manager

elijaharita:
This code doesn't actually do anything yet - it's just a framework which will be used to replace current raw gorm calls around the codebase in the future.

Each table has a corresponding `type InsertTableNameQuery struct`, containing a pointer to the underlying database backend instance. Instances of these are spawned from the main `DBMgr` struct, like `dbMgr.InsertTableName()`, and modifier and action functions are chained onto them in order to do something.

A typical call might look like `item, err := dbMgr.InsertTableName().WithID(5).Get()`. The first function, `.InsertTableName()`, selects which table to operate on, each of which has its own set of structured, typed operations. The function at the end of a dbMgr chain, `.Get()` in this case, is the action, and is usually named as a verb. Between the table selector and the action, are any amount of modifiers, usually named with a preposition, like `.WithID(...)`. 

A lot of this is pretty similar to how Gorm works already, but it makes use of typed function parameters and return types to reduce the chance of mistakes and pave the way to improve performance by avoiding Gorm's expensive runtime introspection (Gorm is still currently used as the backend, but it will be far less painful to change that after this update). Also, this system opens up the possibility to add complex special actions (e.g. `ObjRefsQuery.GetStats()` in dbmgr.go) which will be much more nicely kept in one spot, rather than strewn about the codebase.

Right now, there is a ton of code duplication because I had to re-type a bunch of effectively identical functions that had different return types. However supposedly, generics are coming to Go in 1.18, which means this could potentially be slimmed down a ton soon! Excited!!


dbmgr/dbmgr.go
```diff
@@ -0,0 +1,985 @@
...
+func (q *UsersQuery) Count() (int64, error) {
+	var count int64
+	if err := q.DB.Count(&count).Error; err != nil {
+		return 0, err
+	}
+	return count, nil
+}
+
+func (q *UsersQuery) Exists() (bool, error) {
+	count, err := q.Count()
```
whyrusleeping, line 273: for exists, i discovered recently its *much* faster to do a select * from X where Y limit 1 and just see if we got anything back, counts in sql are really slow for some reason

elijaharita: Ah, that makes sense, cuz count is still going to run through all the rows even if it found the item. Will update when I get the chance

dbmgr/dbmgr.go
```diff
@@ -0,0 +1,985 @@
...
+	if err != nil {
+		return false, err
+	}
+	return count > 0, nil
+}
+
+// Errors if none were deleted
+func (q *UsersQuery) ExpectDelete() error {
+	res := q.DB.Delete(&User{})
```
whyrusleeping, line 282: it might be worth throwing some safety checks in here to make sure nobody just calls this by itself. I think that would delete all users...

dbmgr/dbmgr.go
ribasushi: If you *really* want to press it, the mostest-performantly-optimizable way is:
`SELECT EXISTS ( SELECT 42 FROM {{src}} WHERE {{cond}} )`

Not having a `LIMIT` to contend with matters

dbmgr/dbmgr.go
elijaharita: I was thinking about that. My idea was to put a boolean in relevant query struct isSingle or something along those lines. Queries that fundamentally select one item would get the single flag turned on after being called, and certain actions could throw an error if they weren't called on a single query. 

Even nicer to use would be to have an additional struct, TableNameHereQuerySingle, which would embed TableNameHereQuery and would expose additional functions that should only be called for single queries and not multi line queries. That would make issues show up at compile time. It adds a bit more complexity though, so might want to wait for generics for this approach. I think it sounds super cool though.

dbmgr/dbmgr.go
```diff
@@ -0,0 +1,985 @@
...
+	if order == OrderDescending {
+		q.DB = q.DB.Order("id DESC")
+	} else {
+		q.DB = q.DB.Order("id ASC")
+	}
+	return q
+}
+
+func (q *ContentsQuery) CreateAll(contents []Content) error {
+	return q.DB.Create(&contents).Error
```
whyrusleeping, line 419: passing a pointer to an array here doesnt seem quite right. I dont think that will do what we expect

dbmgr/dbmgr.go
```diff
@@ -0,0 +1,985 @@
...
+func NewDealsQuery(db *gorm.DB) *DealsQuery {
+	return &DealsQuery{DB: db.Model(Deal{})}
+}
+
+func (q *DealsQuery) WithID(id uint) *DealsQuery {
+	q.DB = q.DB.Where("id = ?", id)
+	return q
+}
+
+func (q *DealsQuery) WithSuccessful(valid bool) *DealsQuery {
```
whyrusleeping, line 571: maybe `WithDealOnChain` ?

whyrusleeping (approved):
A few nits, but overall this LGTM. With us moving the structs into this new file, does this commit actually build on its own?

dbmgr/dbmgr.go
elijaharita: my mistake, will fix. although we don't actually even create multiple contents anywhere in the program anyway, right? not sure why i added this.

elijaharita:
@whyrusleeping the new structs are actually just duplicates of the current structs with the json tags stripped out. dbmgr/ currently exists in its own little bubble and hasn't affected the outside program in any way yet. in the future, gorm tags and irrelevant fields can be removed from the current struct definitions, to decouple them from the database and make them more flexible for changes and additions.

dbmgr/dbmgr.go
whyrusleeping: Yeah, i think just adding a flag that says whether *any* condition has been set would be good

PR #26, opened 2021-08-30
Shuttle: Cache auth tokens

arajasek:
Should work, but idk how to test (either manually or some unit test thing)

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -14,6 +14,8 @@ import (
 	"sync"
 	"time"
 
+	lru "github.com/hashicorp/golang-lru"
+
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/urfave/cli/v2"
 	"go.opentelemetry.io/otel"
@@ -223,6 +225,12 @@ func main() {
 			return err
 		}
 
+		// TODO: Paramify this?
+		cache, err := lru.New2Q(1000)
+		if err != nil {
+			return err
+		}
+
 		d := &Shuttle{
 			Node:       nd,
 			Api:        api,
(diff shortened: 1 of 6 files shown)


cmd/estuary-shuttle/main.go
```diff
@@ -462,9 +473,25 @@ type User struct {
...
 	AuthToken       string `json:"-"` // this struct shouldnt ever be serialized, but just in case...
 	StorageDisabled bool
+	AuthExpiry      time.Time
 }
 
 func (d *Shuttle) checkTokenAuth(token string) (*User, error) {
+
+	val, ok := d.authCache.Get(token)
+	if ok {
+		usr, ok := val.(User)
```
whyrusleeping, line 483: probably makes sense to put pointers to users in the cache, i think they get silently converted to pointers under the hood, and this might make things a little weird

arajasek: seems reasonable, done

cmd/estuary-shuttle/main.go
```diff
@@ -493,13 +520,17 @@ func (d *Shuttle) checkTokenAuth(token string) (*User, error) {
 		return nil, err
 	}
 
-	return &User{
+	usr := &User{
 		ID:              out.ID,
 		Username:        out.Username,
 		Perms:           out.Perms,
 		AuthToken:       token,
 		StorageDisabled: out.Settings.ContentAddingDisabled,
```
whyrusleeping, line 528: gotta set that new field here

handlers.go
```diff
@@ -2263,6 +2263,11 @@ func (s *Server) handleGetViewer(c echo.Context, u *User) error {
 		return err
 	}
 
+	var at AuthToken
+	if err := s.DB.Find(&at, "\"user\" = ? AND token = ?", u.ID, c.Param("key")).Error; err != nil {
```
whyrusleeping, line 2267: this wont work AFAIK. we will have to augment checkTokenAuth to store the auth token for the user it returns in that user object. Putting it in an unexported field seems to make sense to me. Then we can just get `u.authToken.Expiry` here

cmd/estuary-shuttle/main.go
arajasek: yes

whyrusleeping (approved):

PR #27, opened 2021-08-31
Small fix for janky `estuary setup` output

elijaharita:
Silenced an expected GORM error and added a missing newline


whyrusleeping (approved):
Ah, so thats how you make gorm be quiet. nice

PR #28, opened 2021-09-01
feat: modify Makefile/go.mod so lotus is dep

lanzafame:
This PR makes estuary more self-contained in terms of lotus/filecoin-ffi deps.


whyrusleeping (approved):
Woo! Thank you!

PR #31, opened 2021-09-02
proof-of-concept support for encrypted data

b5:
so, don't merge this. Mainly just to get a discussion going. I had to butcher your `go.mod`. Deps are hard.

Also, **WARNING IMPLEMENTATION NOT FINISHED**. I _know_ this doesn't produce secure data yet. This is up to make sure we can actually plug it in the right way when it's at least in a usable state


parkan:
love that this is happening!

re: go.mod, I had similar issues/had to run tidy when my estuary repo was slightly out of sync vs what master on this repo was expecting, getting everything in sync removed the need to do anything (may have had a slightly out of date go version too), may be worth trying?

marshall:
heya @b5, looks like this PR has been stale for a while. is this still relevant? some of my thoughts below:

WRT securing data at rest for barge to upload, this looks relatively straight forward but I see [qri-io/wnfs-go] is still in an experimental state. what do you think is needed there to make it usable for such a use case?

for estuary, I think you'd want to also be able to display some info about `wnfs` content in estuary's webui, and also support retrieving content back using some of the newer tools such as [filclient] and [autoretrieve]. wdyt?

LMK if you think this has a chance of being brought up to speed soon, otherwise maybe we can close and re-open when the technical details are worked out more

[qri-io/wnfs-go]: https://github.com/qri-io/wnfs-go
[filclient]: https://github.com/application-research/filclient
[autoretrieve]: https://github.com/application-research/autoretrieve

PR #33, opened 2021-09-06
Systemd service for primary estuary node

lanzafame:
This PR contains a generic systemd service file for a primary estuary node. The `go-log` package was upgraded so make use of the extra configurability that it provides.


whyrusleeping (approved):
LGTM, thanks!

PR #34, opened 2021-09-06
avoid copying contentStagingZone lock

iand:
The copy is only needed for marshaling to JSON but that has a race anyway since the lock is not held during the marshal process. 
Took a simple approach of adding a DeepCopy method which could be error prone if new fields are added. There are several alternatives: (a) create a new snapshot type which is simply a struct with the same fields that is intended for serialization or (b) move the data to an embedded type guarded by the mutex or even (c) treat contentStagingZone as a value object and let ContentManager manage the locking.

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -133,6 +133,26 @@ type contentStagingZone struct {
 	lk sync.Mutex
 }
 
+func (cb *contentStagingZone) DeepCopy() *contentStagingZone {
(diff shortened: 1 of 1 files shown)


whyrusleeping (approved):
Oh nice, that makes sense. Even if we change fields later and forget to update deep copy, it shouldnt break anything since those are just for introspection of things.

PR #35, opened 2021-09-06
replace cid by multihash in objects table

iand:
Adds a migration that manually adds the hash column, converts all cids to hashes and then removes the cid column. 

Fix up various places where cid is referenced but this is pretty superficial at the moment. There could be places where we could be using multihash instead of cid and avoid any conversion.

### DIFF
--- a/main.go
+++ b/main.go
@@ -145,7 +145,8 @@ type Content struct {
 
 type Object struct {
 	ID         uint        `gorm:"primarykey"`
-	Hash       util.DbHash `gorm:"index"`
+	Cid        util.DbCID  `gorm:"index"` // Cid is deprecated
+	Hash       util.DbHash `gorm:"index"` // Hash will replace Cid in the future
 	Size       int
 	Reads      int
 	LastAccess time.Time
@@ -484,51 +485,7 @@ func setupDatabase(cctx *cli.Context) (*gorm.DB, error) {
 	}
 
 	db.AutoMigrate(&Content{})
-
-	// Check if we need to migrate cid to hash.
-	if db.Migrator().HasColumn(&Object{}, "hash") {
-		db.AutoMigrate(&Object{})
-	} else {
-		// Migration from cid to hash is needed
-		fmt.Println("migrating object cids to multihashes")
-		type cidrow struct {
-			ID  uint
-			Cid util.DbCID
-		}
-		rows, err := db.Raw("select id, cid from objects").Rows()
-		if err != nil {
-			return nil, err
-		}
-		defer rows.Close()
-
-		if err := db.Transaction(func(tx *gorm.DB) error {
-			if err := tx.Migrator().AddColumn(&Object{}, "hash"); err != nil {
-				return err
-			}
-			count := 0
-			for rows.Next() {
-				count++
-				if count%1000 == 0 {
-					fmt.Printf("  migrated %d rows\n", count)
-				}
-				var cr cidrow
-				rows.Scan(&cr.ID, &cr.Cid)
-				hash := util.DbHashFromDbCID(&cr.Cid)
-				if err := tx.Exec("update objects set hash=? where id=?", hash, cr.ID).Error; err != nil {
-					return err
-				}
-			}
-			fmt.Printf("  migrated %d rows...done\n", count)
-			if err := tx.Migrator().DropColumn(&Object{}, "cid"); err != nil {
-				return err
-			}
-
-			return nil
-		}); err != nil {
-			return nil, err
-		}
-	}
-
+	db.AutoMigrate(&Object{})
 	db.AutoMigrate(&ObjRef{})
 	db.AutoMigrate(&Collection{})
 	db.AutoMigrate(&CollectionRef{})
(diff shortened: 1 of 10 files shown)


util/database.go
```diff
@@ -77,3 +78,66 @@ func (dbc *DbCID) UnmarshalJSON(b []byte) error {
...
+	if err != nil {
+		return err
+	}
+
+	d.Hash = h
+	return nil
+}
+
+func (d *DbHash) Cid() cid.Cid {
+	// TODO: is cid.Raw appropriate?
```
iand, line 141: Highlighting this TODO

main.go
```diff
@@ -458,7 +457,51 @@ func setupDatabase(cctx *cli.Context) (*gorm.DB, error) {
 	}
 
 	db.AutoMigrate(&Content{})
-	db.AutoMigrate(&Object{})
+
+	// Check if we need to migrate cid to hash.
+	if db.Migrator().HasColumn(&Object{}, "hash") {
+		db.AutoMigrate(&Object{})
+	} else {
+		// Migration from cid to hash is needed
```
iand, line 465: Manual migration only performed if we detect that the database has no hash column. Otherwise we use the auto migration.

main.go
```diff
@@ -458,7 +457,51 @@ func setupDatabase(cctx *cli.Context) (*gorm.DB, error) {
...
+			return nil, err
+		}
+		defer rows.Close()
+
+		if err := db.Transaction(func(tx *gorm.DB) error {
+			if err := tx.Migrator().AddColumn(&Object{}, "hash"); err != nil {
+				return err
+			}
+			count := 0
+			for rows.Next() {
```
whyrusleeping, line 482: Any idea how long this would take on 100 million rows? My worry is that it would take hours, and estuary being offline for that long isnt great.

iand: Will try and benchmark some different sizes to get an estimate. Informally it's about a second for 10k rows in a docker container on my machine.

iand: We can also do it as a multistage. Add the column and write to both cid+hash during every insert. In background update all the hashes. We would have to rewrite where clauses to use cid OR hash. 

Then some time in the future we remove the cid column. We would have to be careful not to break people who have not deployed and run the first stage beforehand.

whyrusleeping: yeah, so at that speed thats over two hours for the full migration, which means the application is offline for that long, which would be pretty bad.

That said, maybe we could add some parallelism? I think I would be okay with under an hour of downtime if you think we could hit that. Then we could just schedule a time for the migration to happen and do it then.

util/database.go
whyrusleeping: Yeah, I think Raw is fine, it just needs to be something consistent, and at the blockstore layer the codec gets ignored anyways

main.go
whyrusleeping: Also, I think @lanzafame has backups of the database laying around, we could try it out on a replica

iand:
I wanted a slightly more general solution to implementing large data migrations while avoiding long periods of downtime. I think the most elegant way to do this is to make migrations multi-stage: a pre-migration of data that retains compatibility followed by a code update to remove the use of the old data. A final stage would then remove the old version of the data.

I implemented a simple migration handler that runs the data migration in batches in the background when estuary starts. It records completion in a new migrations table at which point we can move to the next stage and update estuary to use the hashes instead of cids.  

/cc @whyrusleeping

migration.go
```diff
@@ -0,0 +1,109 @@
...
+				case <-ctx.Done():
+					return ctx.Err()
+				default:
+				}
+
+				for _, o := range objs {
+					o.Hash = util.DbHashFromDbCID(&o.Cid)
+				}
+
+				tx.Save(&objs)
```
whyrusleeping, line 50: the `Save` method in gorm gives me trouble somewhat often. I've switched all my calls of it to just updating the specific field I want to update.

whyrusleeping: Also, is calling this on the DB instance from the closure the right thing? I'm not really sure what the implications are here

migration.go
```diff
@@ -0,0 +1,109 @@
...
+
+		// Kick off the migration
+		if minfo.required {
+			// Migration must run to completion synchronously
+			if err := runMigration(ctx, db, mig, minfo.fn); err != nil {
+				return err
+			}
+		} else {
+			// Migration runs asynch in the background
+			go runMigration(ctx, db, mig, minfo.fn)
```
whyrusleeping, line 85: probably want to at least log the error this returns if it returns one?

whyrusleeping: oh nvm, i see that it does get logged below

migration.go
```diff
@@ -0,0 +1,109 @@
...
+	log.Infof("%s migration started", mig.Name)
+	err := fn(ctx, db)
+	if err != nil {
+		log.Errorf("%s migration failed: %v", mig.Name, err)
+		return err
+	}
+	log.Infof("%s migration complete", mig.Name)
+
+	mig.CompletedAt = time.Now().UTC()
+	if err := db.Save(mig).Error; err != nil {
```
whyrusleeping, line 103: again, i would avoid using gorms Save, it touches *all* the fields, and has done weird things before

whyrusleeping:
I'm going to be traveling this week, so I want to wait to deploy this until im going to be 'available' to debug if anything weird happens.

corinne-antonia:
@iand @whyrusleeping This PR is over 6-months old, I'm worried its a little stale. Do you guys still want this?

PR #36, opened 2021-09-08
use api.estuary.tech as preferred upload endpoint when local adding disabled

iand:
This inverts the condition in `getPreferredUploadEndpoints` to add estuary.tech  when local adding is disabled. Without this change a local web client uses api.estuary.tech for uploads instead of the configured api.


iand:
Already fixed in latest master

PR #37, opened 2021-09-08
add indexes for frequently joined columns

iand:
Surveyed the codebase for columns that commonly appear in joins or where clauses. I limited changes to large tables where we should see the most impact. Adding these indexes changes query plans from using Seq Scan on join tables to Index Scan (tested with a subset of data).

Specifying the `CONCURRENTLY` option should tell postgresql to build the index without taking locks that block reads or inserts.


whyrusleeping (approved):
Oh nice, these will definitely be helpful, thank you!

PR #38, opened 2021-09-09
CM: Protect newly added data from being deleted

arajasek:
Test might come

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -679,7 +679,29 @@ func (cm *ContentManager) addDatabaseTrackingToContent(ctx context.Context, cont
 	var objects []*Object
 	cset := cid.NewSet()
 
+	defer func() {
+		cm.inflightCidsLk.Lock()
+		_ = cset.ForEach(func(c cid.Cid) error {
+			v, ok := cm.inflightCids[c]
+			if !ok || v <= 0 {
+				log.Errorf("cid should be inflight but isn't: %s", c)
+			}
+
+			cm.inflightCids[c]--
+			if cm.inflightCids[c] == 0 {
+				delete(cm.inflightCids, c)
+			}
+			return nil
+		})
+		cm.inflightCidsLk.Unlock()
+	}()
+
 	err := merkledag.Walk(ctx, func(ctx context.Context, c cid.Cid) ([]*ipld.Link, error) {
+		// cset.Visit gets called first, so if we reach here we should immediately track the CID
+		cm.inflightCidsLk.Lock()
+		cm.inflightCids[c]++
+		cm.inflightCidsLk.Unlock()
+
 		node, err := dserv.Get(ctx, c)
 		if err != nil {
 			return nil, err
@@ -705,11 +727,12 @@ func (cm *ContentManager) addDatabaseTrackingToContent(ctx context.Context, cont
 
 		return node.Links(), nil
 	}, root, cset.Visit, merkledag.Concurrent())
+
 	if err != nil {
 		return err
 	}
 
-	if err := cm.addObjectsToDatabase(ctx, cont, objects, "local"); err != nil {
+	if err = cm.addObjectsToDatabase(ctx, cont, objects, "local"); err != nil {
 		return err
 	}
 
--- a/offloading.go
+++ b/offloading.go
@@ -187,31 +187,36 @@ func (cm *ContentManager) OffloadContents(ctx context.Context, conts []uint) (in
 		return 0, err
 	}
 
-	// TODO: I believe that we need to hold a lock for the entire period that
-	// we are deleting objects from the blockstore, otherwise a new file could
-	// come in that has overlapping blocks, and have its blocks deleted by this
-	// process.
 	var deleteCount int
 	var cids []cid.Cid
+	cm.inflightCidsLk.Lock()
+	defer cm.inflightCidsLk.Unlock()
 	for rows.Next() {
 		var dbc util.DbCID
 		if err := rows.Scan(&dbc); err != nil {
 			return deleteCount, err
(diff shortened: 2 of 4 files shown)


offloading.go
```diff
@@ -199,7 +198,10 @@ func (cm *ContentManager) OffloadContents(ctx context.Context, conts []uint) (in
 			return deleteCount, err
 		}
 
-		cids = append(cids, dbc.CID)
+		if cm.inflightCids.Has(dbc.CID) {
```
whyrusleeping, line 201: i think you need a `!` here

arajasek: yes

handlers.go
```diff
@@ -679,7 +679,26 @@ func (cm *ContentManager) addDatabaseTrackingToContent(ctx context.Context, cont
 	var objects []*Object
 	cset := cid.NewSet()
 
-	err := merkledag.Walk(ctx, func(ctx context.Context, c cid.Cid) ([]*ipld.Link, error) {
+	defer func() {
+		cm.inflightCidsLk.Lock()
+		err = cset.ForEach(func(c cid.Cid) error {
+			v, ok := cm.inflightCids[c]
+			if !ok || v <= 0 {
+				return xerrors.Errorf("cid should be inflight but isn't: %s", c)
```
arajasek, line 687: should we just log here and absorb the error? idk if we want to cause the caller to fail if something went funky here

handlers.go
```diff
@@ -651,7 +651,7 @@ func (s *Server) importFile(ctx context.Context, dserv ipld.DAGService, fi io.Re
 
 var noDataTimeout = time.Minute * 10
 
-func (cm *ContentManager) addDatabaseTrackingToContent(ctx context.Context, cont uint, dserv ipld.NodeGetter, bs blockstore.Blockstore, root cid.Cid, cb func(int64)) error {
+func (cm *ContentManager) addDatabaseTrackingToContent(ctx context.Context, cont uint, dserv ipld.NodeGetter, bs blockstore.Blockstore, root cid.Cid, cb func(int64)) (err error) {
```
whyrusleeping, line 654: named returns are the devil, shame upon you and your family, 6 generations back!

handlers.go
whyrusleeping: yeah, log and absorb the error.

We also need to delete the cid from the map if we are the caller that drops its count to zero.

handlers.go
arajasek: first of all i love the devil second of all how else are you supposed to return errors from defers

whyrusleeping: at the very least dont name the out parameter error `err`, name it `outErr` or something if you absolutely must commit such sacrilege.

offloading.go
```diff
@@ -187,10 +187,9 @@ func (cm *ContentManager) OffloadContents(ctx context.Context, conts []uint) (in
 		return 0, err
 	}
 
-	// TODO: I believe that we need to hold a lock for the entire period that
-	// we are deleting objects from the blockstore, otherwise a new file could
-	// come in that has overlapping blocks, and have its blocks deleted by this
-	// process.
+	cm.inflightCidsLk.Lock()
```
whyrusleeping, line 190: hrm... do we need to hold this over the entire duration? Or can we just lock around the inflight checks and the delete calls individually?

handlers.go
```diff
@@ -679,7 +679,29 @@ func (cm *ContentManager) addDatabaseTrackingToContent(ctx context.Context, cont
 	var objects []*Object
 	cset := cid.NewSet()
 
+	defer func() {
+		cm.inflightCidsLk.Lock()
```
arajasek, line 683: we could instead lock and release on each iteration, but this should be a zoomy zoom loop

offloading.go
```diff
@@ -187,25 +187,28 @@ func (cm *ContentManager) OffloadContents(ctx context.Context, conts []uint) (in
...
+		}
+
 		if len(cids) > 100 {
 			if err := cm.Blockstore.DeleteMany(cids); err != nil {
 				return deleteCount, err
 			}
 			deleteCount += len(cids)
 			cids = cids[:0]
+			// unlock and re-lock to unblock other things
+			cm.inflightCidsLk.Unlock()
```
arajasek, line 210: is this Bad?

handlers.go
arajasek: kk it's gone

offloading.go
whyrusleeping: hrm... So I think the lock needs to be held from before when you check if you can delete a thing, until after you delete it. Otherwise its state of 'deleteableness' might change from underneath you.

whyrusleeping: Okay, I think i've convinced myself that this still satisfies that requirement.

arajasek: yeah that is definitely the requirement and is what this is claiming to do. i don't really know a cleaner way to do this.

handlers.go
```diff
@@ -705,11 +727,12 @@ func (cm *ContentManager) addDatabaseTrackingToContent(ctx context.Context, cont
 
 		return node.Links(), nil
 	}, root, cset.Visit, merkledag.Concurrent())
+
 	if err != nil {
 		return err
 	}
 
-	if err := cm.addObjectsToDatabase(ctx, cont, objects, "local"); err != nil {
+	if err = cm.addObjectsToDatabase(ctx, cont, objects, "local"); err != nil {
```
whyrusleeping, line 735: lol, we should always rescope errors when doing a compound if statement. Ensures we dont ever accidentally use the wrong error value

whyrusleeping (approved):
Alright, leeroy jenkins, lets do this

PR #39, opened 2021-09-10
add auto retrieve endpoint

elijaharita:
i think maybe this should probably possibly do what i think it is going to do, but my brain is tired

intention for GET /retrieval-candidates - given a cid, get a list of root cids it belongs to along with the miner addresses they originate from

i've never used sql join before, so that part should probably get extra scrutiny

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -3538,7 +3538,9 @@ func (s *Server) handleGetRetrievalCandidates(c echo.Context) error {
 	// Read the cid from the client request
 	cid, err := cid.Decode(c.Param("cid"))
 	if err != nil {
-		return c.String(http.StatusBadRequest, "Invalid cid")
+		return c.JSON(http.StatusBadRequest, map[string]string{
+			"error": "invalid cid",
+		})
 	}
 
 	// Get the object with the requested cid
--- a/handlers.go
+++ b/handlers.go
@@ -3529,7 +3529,7 @@ func (s *Server) handleGetPublicNodeInfo(c echo.Context) error {
 	})
 }
 
-type retrievalCandidate struct {
+type RetrievalCandidate struct {
 	maddr   address.Address
 	rootCid cid.Cid
 }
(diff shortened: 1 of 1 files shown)


handlers.go
```diff
@@ -3526,3 +3528,64 @@ func (s *Server) handleGetPublicNodeInfo(c echo.Context) error {
 		PrimaryAddress: s.FilClient.ClientAddr,
 	})
 }
+
+type retrievalCandidate struct {
+	maddr   address.Address
```
whyrusleeping, line 3533: these need to be exported (capitalized) otherwise the json formatter wont print them.

Probably also want to include the deal ID

handlers.go
```diff
@@ -3526,3 +3528,64 @@ func (s *Server) handleGetPublicNodeInfo(c echo.Context) error {
...
+
+	// Get the deals for each of those object refs' contents
+	var candidateInfos []struct {
+		miner string
+		cid   util.DbCID
+	}
+	if err := s.DB.
+		Table("content_deals").
+		Where("content IN ? AND NOT failed ", contentIDs).
+		Joins("JOIN contents ON content_deals.content = content.id").
```
whyrusleeping, line 3570: I think you want to say "left join" here?

Also, I think we can do this whole query in one go:
```sql
select miner, cid from content_deals left 
  join contents on content_deals.content = contents.id
  where contents.id in (
    select content from obj_refs
      where object in (select id from objects where cid = ?)
  );
```

whyrusleeping: (that is, we can do away with lines 3544 through 3565)

whyrusleeping: Aside from me playing sql golf though, Your solution here seems correct (potentially minus the 'left' qualifier on the join)

handlers.go
```diff
@@ -3526,3 +3528,64 @@ func (s *Server) handleGetPublicNodeInfo(c echo.Context) error {
...
+			return err
+		}
+
+		candidates = append(candidates, retrievalCandidate{
+			maddr:   maddr,
+			rootCid: candidateInfo.cid.CID,
+		})
+	}
+
+	c.JSON(http.StatusOK, candidates)
```
whyrusleeping, line 3588: its idiomatic to return `c.JSON(...)`, since it can technically error

handlers.go
elijaharita: i've attempted to implement this, although let me know if it is going to do something crazy

handlers.go
```diff
@@ -3526,3 +3528,52 @@ func (s *Server) handleGetPublicNodeInfo(c echo.Context) error {
...
+	}
+
+	var candidateInfos []struct {
+		miner  string
+		cid    util.DbCID
+		dealID uint
+	}
+	if err := s.DB.
+		Table("content_deals").
+		Where("content IN ? AND NOT failed",
```
whyrusleeping, line 3554: youll have to wrap the `?` in parentheses for subqueries, gorm doesnt do it on its own for some reason

PR #41, opened 2021-09-14
feat: support building for calibnet and update Makefile

frrist:
- allows testnet addresses to be used
- creates calibration network-specific database
- defines calibration specific miners
- builds all binaries from `make`
- `make clean` removes all bins

### DIFF
--- a/main.go
+++ b/main.go
@@ -9,12 +9,12 @@ import (
 	"sync"
 	"time"
 
+	"github.com/application-research/estuary/build"
 	"github.com/application-research/estuary/node"
 	"github.com/application-research/estuary/pinner"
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -35,62 +35,6 @@ import (
...
 var log = logging.Logger("estuary")
 
-var defaultMiners []address.Address
-
-var defaultDatabaseValue = "sqlite=estuary.db"
-
-func init() {
-	// miners from minerX spreadsheet
-	minerStrs := []string{
-		"f02620",
```
jimmylee, line 45: Nice cleanup here :-)

whyrusleeping (approved):
Nice, this is really good stuff

PR #42, opened 2021-09-17
doc: add short getting started with development doc

frrist:
This is far from a complete getting started guide. I wrote this over my tribute week while I was figuring out how to use estuary and shuttles together, mostly via trial and error. It attempts to cover most of the things I needed to understand in order to develop with Estuary and to get the services communicating with each other.


whyrusleeping (approved):
This is great, thanks @frrist !

PR #43, opened 2021-09-30
Implemented exponential backoff when shuttle tries to reconnect to node

geoff-vball:
Exact params for backoff are up for discussion and easily changed.

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -31,6 +31,7 @@ import (
 	"github.com/application-research/estuary/stagingbs"
 	"github.com/application-research/estuary/util"
 	"github.com/application-research/filclient"
+	"github.com/cenkalti/backoff/v4"
 	datatransfer "github.com/filecoin-project/go-data-transfer"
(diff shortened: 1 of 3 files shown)


cmd/estuary-shuttle/main.go
```diff
@@ -424,12 +433,14 @@ func (d *Shuttle) RunRpcConnection() error {
...
 			log.Errorf("failed to dial estuary rpc endpoint: %s", err)
-			time.Sleep(time.Second * 10)
+			time.Sleep(backoffTimer.NextBackOff())
 			continue
 		}
 
 		if err := d.runRpc(conn); err != nil {
 			log.Errorf("rpc routine exited with an error: %s", err)
+			backoffTimer.Reset()
+			// TODO Do we still need to wait here?
```
whyrusleeping, line 443: We should at least sleep for the default backoff duration, ten seconds is definitely excessive at this point

whyrusleeping (approved):
LGTM, but i think we should address your TODO and replace the ten second post-disconnect sleep with just another call to the backofftimer

PR #44, opened 2021-09-30
add support for running with a dev-signed cert

marshall:
by default, estuary runs it's webserver on an insecure socket, presumably requiring a TLS termination in front of it before a shuttle can actually connect.

this is probably fine in production, but as it stands, running estuary + shuttle locally in a dev environment won't work unless you setup a terminating proxy, which is quite painful. the error messages when trying to follow the docs as is look something like this from `estuary-shuttle`:
```
2021-09-30T14:47:11.052-0500    ERROR   shuttle estuary-shuttle/main.go:426     failed to dial estuary rpc endpoint: websocket.Dial wss://localhost:3004/shuttle/conn: tls: first record does not look like a TLS handshake
```

this PR makes it straight forward to provide a self-signed cert to estuary directly, allowing the shuttle to connect as outlined in the setup docs without an issue

### DIFF
--- a/GettingStartedDevelopment.md
+++ b/GettingStartedDevelopment.md
@@ -19,6 +19,15 @@ Next, set the `FULLNODE_API_INFO` environment variable to a synced lotus node. T
 $ export FULLNODE_API_INFO=wss://api.chain.love
 ```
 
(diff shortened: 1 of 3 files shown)


whyrusleeping (approved):
Nice, thats pretty straightforward.

PR #62, opened 2021-10-25
Auto retrieve service

elijaharita:
Serve filecoin network content to bitswap clients using info from Estuary


cmd/autoretrieve/main.go
```diff
@@ -0,0 +1,772 @@
...
+
+			if err := r.bsnet.SendMessage(ctx, sender, resMsg); err != nil {
+				logger.Errorf("Could not send bitswap message: %v", err)
+			}
+
+			continue
+		}
+
+		// If it no block found, then check for retrieval candidates
+		unfilteredCandidates, err := GetRetrievalCandidates("https://api.estuary.tech/retrieval-candidates", entry.Cid)
```
whyrusleeping, line 434: this probably wants to be configurable

cmd/autoretrieve/main.go
```diff
@@ -0,0 +1,772 @@
...
+		},
+		&cli.StringFlag{
+			Name:  "endpoint",
+			Value: "https://api.estuary.tech/retrieval-candidates",
+		},
+	}
+
+	app.Commands = []*cli.Command{
+		{
+			Name: "test-blacklist",
```
whyrusleeping, line 79: you could just call this 'check-blacklist' and leave it as a permanent part of the codebase, no need for it to be a 'test' command. I generally like to leave all sorts of utility commands around

cmd/autoretrieve/main.go
```diff
@@ -0,0 +1,772 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		blockstoreDatastore, err := flatfs.CreateOrOpen(filepath.Join(config.dataDir, blockstoreSubdir), parseShardFunc, false)
+		if err != nil {
+			return nil, err
+		}
+
+		node.blockstore = &senderBlockstore{
```
whyrusleeping, line 233: its probably worth throwing some metrics wrapping in here, both above and below the auto-retrieve bits so we can measure them more easily

whyrusleeping: see "measure blockstore" elsewhere

cmd/autoretrieve/main.go
```diff
@@ -0,0 +1,772 @@
...
+		}
+	}
+
+	// Host
+	{
+		var peerkey crypto.PrivKey
+		keyPath := filepath.Join(config.dataDir, "peerkey")
+		keyFile, err := os.ReadFile(keyPath)
+		if err != nil {
+			logger.Infof("Generating new peer key...")
```
whyrusleeping, line 246: log should be below the error check

elijaharita: wdym? we only generate a new key if there is an error, which is the special case, so i wanted to log it. otherwise, we load an existing one. do i not understand this code right?

cmd/autoretrieve/main.go
```diff
@@ -0,0 +1,654 @@
...
+				if retrievalInProgress {
+					break
+				}
+			}
+
+			// If we didn't find any retrieval in progress already...
+			if !retrievalInProgress {
+
+				// ...start it on this goroutine
+				if err := r.retrieveFromBestCandidate(ctx, candidates); err != nil {
```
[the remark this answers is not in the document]
elijaharita: in go-bitswap, the tag string has a uuid appended to it. what's the purpose of that, is it something that would benefit this project as well?

cmd/autoretrieve/main.go
whyrusleeping: well, in this case we say "generating new peer key" and then if the file read failed for some other reason, that wouldnt be true and we would error out here

elijaharita: AHH whoops i totally did not understand what you were trying to say before, lol. i get it, will fix

cmd/autoretrieve/main.go
```diff
@@ -0,0 +1,775 @@
...
+
+			if err := r.bsnet.SendMessage(ctx, sender, resMsg); err != nil {
+				logger.Errorf("Could not send bitswap message: %v", err)
+			}
+
+			continue
+		}
+
+		// If it no block found, then check for retrieval candidates
+		unfilteredCandidates, err := GetRetrievalCandidates("https://api.estuary.tech/retrieval-candidates", entry.Cid)
```
marshall, line 434: this should probably use the `endpoint` provided from the command line

cmd/autoretrieve/main.go
```diff
@@ -0,0 +1,775 @@
...
+				"Retrieval query %v/%v succeeded from miner %s for %s",
+				checked+1,
+				len(candidates),
+				candidate.Miner,
+				candidate.RootCid,
+			)
+
+			queriesLk.Lock()
+			queries = append(queries, CandidateQuery{Candidate: candidate, Response: query})
+			checked++
```
marshall, line 605: instead of mutating / accessing the `checked` global inside of this goroutine, would it be safer to pass in the loop index as part of the goroutine call? i.e..
```go
for idx, candidate := range candidates {
  go func(idx uint, candidate RetrievalCandidate) {
    // .. retrieval logic
  }(idx, candidate)
}

File: cmd/autoretrieve/main.go [hunk 8]
```diff
@@ -0,0 +1,775 @@
...
+		doneLk.Lock()
+		done = true
+		doneLk.Unlock()
+
+		retrieveCancel()
+		logger.Errorf("Retrieval timed out after not receiving data for %s (stopped at %s)", r.config.retrievalTimeout, humanize.IBytes(lastBytesReceived))
+	})
+	stats, err := r.fc.RetrieveContentWithProgressCallback(retrieveCtx, query.Candidate.Miner, proposal, func(bytesReceived uint64) {
+		doneLk.Lock()
+		if !done {
```
marshall, line 749: would it be less contentious to unlock immediately after retrieving the value of `done` here, instead of after the logic block, i.e.:

```go
doneLk.Lock()
isDone := done
doneLk.Unlock()
if !isDone {
  //...
}
```

marshall (changes requested):
I left just a few minor change requests, overall looks good though :)

cmd/autoretrieve/main.go
elijaharita: we also need to lock around `lastBytesReceived` which is used inside the if, and i believe we also need exclusive access to do the timer reset steps in order so its state doesn't get messed up. an instance of this mutex should only get locked and unlocked around 10 times throughout its lifetime of a few seconds, maybe 100 in rare cases. so i don't think it should be much of a problem.

elijaharita: actually, if you have another idea on how to detect retrieval hangs, i'd be open to trying something else. using Timer was the quickest way i was able to figure it out, but it does seem pretty hacky

cmd/autoretrieve/bitswap/provider.go
```diff
@@ -0,0 +1,210 @@
...
+			}
+		}
+	}()
+}
+
+func (provider *Provider) PeerDisconnected(peer peer.ID) {
+	// On disconnect, destroy the peer's queue
+	provider.blockQueuesLk.Lock()
+
+	// TODO: need to clean up waiting blocks in block manager
```
whyrusleeping, line 166: yeah, just closing the channels here will likely cause things to panic elsewhere

cmd/autoretrieve/bitswap/provider.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}
+}
+
+func (provider *Provider) ReceiveError(err error) {
+	logger.Errorf("Error receiving bitswap message: %v", err)
+}
+
+func (provider *Provider) PeerConnected(peer peer.ID) {
+	blockQueue := BlockQueue{
```
whyrusleeping, line 129: you should probably use the peertaskqueue for this.

The code lives here: https://github.com/ipfs/go-peertaskqueue

Basically, you make a queue, and then you push tasks onto it. So below here instead of doing the sends directly (which could block and gum things up elsewhere) you could push a task onto the queue with a topic of "send block", and the data could be the block (or maybe the data would be the cid, so you dont have to keep the blocks in memory for a while).

Then, you have other worker gorotuines calling "PopTasks" and executing the tasks.

cmd/autoretrieve/bitswap/provider.go
whyrusleeping: using the peertaskqueue is probably the right way forward here, no need for us to reimplement all that tricky logic

cmd/autoretrieve/filecoin/retriever.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+func NewRetriever(config RetrieverConfig, host host.Host, api api.Gateway, datastore datastore.Batching, blockManager *blocks.Manager) (*Retriever, error) {
+
+	keystore, err := keystore.OpenOrInitKeystore(filepath.Join(config.DataDir, walletSubdir))
+	if err != nil {
+		return nil, err
+	}
+
+	wallet, err := wallet.NewWallet(keystore)
```
whyrusleeping, line 75: in general, instead of instantiating things inside constructors, pass them into the constructor. In this case, the wallet and filclient instance should just be taken as parameters to the NewRetriever function

elijaharita: hmm, interesting thing to consider. i made it this way because i wanted to keep things that didn't need to be shared inside of their own modules to avoid spaghetti. it doesn't matter if i expose them at the top level since this isn't a library. 

i do see how this might not be future proof, though, in the case that i might want to add another module which does need filclient and wallet. are there any other important reasons?

cmd/autoretrieve/bitswap/provider.go
```diff
@@ -0,0 +1,247 @@
...
+
+	provider.network.SetDelegate(provider)
+
+	return provider, nil
+}
+
+func (provider *Provider) startSend() {
+	provider.sendWorkerCountLk.Lock()
+	// TODO: use a constant or config val
+	if provider.sendWorkerCount < 1 {
```
whyrusleeping, line 88: this is only ever spinning up a single worker?

cmd/autoretrieve/bitswap/provider.go
```diff
@@ -0,0 +1,247 @@
...
+					}
+				}
+
+				if err := provider.network.SendMessage(context.Background(), peer, msg); err != nil {
+					logger.Errorf("Could not send bitswap message to %s: %v", peer, err)
+				}
+			}
+		}()
+	}
+	provider.sendWorkerCountLk.Unlock()
```
whyrusleeping, line 117: Should just defer the Unlocking, right after taking the lock

cmd/autoretrieve/bitswap/provider.go
elijaharita: just had to add config val for it, that number changes max workers

PR #66, opened 2021-10-28
implement upload only api keys

whyrusleeping:
This allows safer delegation of keys to other parties to upload things on your behalf. Still doesnt prevent them from uploading tons of random stuff, but that management is left to those who choose to use this feature.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -119,7 +119,7 @@ func (s *Server) ServeAPI(srv string, logging bool, lsteptok string, cachedir st
 
 	e.GET("/test-error", s.handleTestError)
 
-	e.GET("/viewer", withUser(s.handleGetViewer), s.AuthRequired(util.PermLevelUser))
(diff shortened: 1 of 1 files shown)


handlers.go
```diff
@@ -119,7 +119,7 @@ func (s *Server) ServeAPI(srv string, logging bool, lsteptok string, cachedir st
 
 	e.GET("/test-error", s.handleTestError)
 
-	e.GET("/viewer", withUser(s.handleGetViewer), s.AuthRequired(util.PermLevelUser))
```
jimmylee, line 122: I don't know if I agree with this but I think it allows for a full client flow so thats okay. Its just that the viewer might have information that the upload key shouldn't know about, but its fine for now since I don't think this is something that will live forever.

whyrusleeping: Yeah, the main thing is that when youre uploading you need access to the set of shuttles to upload to, which requires hitting this endpoint.

PR #68, opened 2021-10-29
switch to FFI for commP hashing

marshall:
I've been bench-testing this locally with a manually pinned CID, and am seeing 40-50% speed improvements over `GeneratePieceCommitment`

two questions before this lands:
* should I add a more reproducible bench test?
* curious about other potential ramifications of using FFI, most of my testing has been with relatively controlled data, not clear on the tradeoffs in accuracy, etc

/cc @whyrusleeping for :eyes:

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -233,7 +233,7 @@ func main() {
 				return nil, err
 			}
 
-			commpcid, size, err := filclient.GeneratePieceCommitment(ctx, c, nd.Blockstore)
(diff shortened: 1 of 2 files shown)


whyrusleeping (approved):
LGTM, Ideally we could get away from using FFI for things, but since we still need it for signatures, this is fine

whyrusleeping:
the main reason for not using the FFI stuff is that it makes the build process sketchier, and FFI code is much more prone to causing weird memory corruption issues

ribasushi:
Belated 👍 from me too. The non-ffi code is very naive, and needs a bit of a redesign to start competing again. Will flag @marshall to retest whenever I get around to that...

PR #70, opened 2021-10-30
initial CI for estuary

marshall:
this is rudimentary, but does a basic build and test of all the Go code in the repo for each pull request and push to master. I tested this on my fork's branch first, then changed the branch name to master before sending the PR -- it does seem to work.

as a bonus, I ended up finding and fixing a SQL syntax error in our tests as part of this ;)

the successful run from my branch is here:
https://github.com/marshall/estuary/runs/4053162791?check_suite_focus=true

/cc @whyrusleeping @jimmylee

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,35 @@
+name: Go
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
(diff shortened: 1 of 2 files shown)


elijaharita (approved):
this seems nice! why will probably have to have the final say on this.

on a side note, the syntax error is... interesting. dbmgr isn't actually in use by estuary yet because we're waiting for generics to come out to squish the module into a more maintainable form, but we'll definitely have to remember to fix that problem within `Exists()` then. thanks for catching!

whyrusleeping:
Thanks for this @marshall ! Sorry for the delay on merging, travel has been distracting..

PR #74, opened 2021-11-04
implement an upload proxy

whyrusleeping:
This way we can have people just upload things to `upload.estuary.tech` and it will pick the right shuttle to hit. For people uploading *lots* of data, the will still want to go direct to a shuttle, but this works for more 'normal' users.

### DIFF
--- a/cmd/shuttle-proxy/main.go
+++ b/cmd/shuttle-proxy/main.go
@@ -131,7 +131,10 @@ func (p *Proxy) handleContentAdd(c echo.Context) error {
 
 	c.Response().WriteHeader(resp.StatusCode)
 
-	io.Copy(c.Response().Writer, resp.Body)
+	_, err = io.Copy(c.Response().Writer, resp.Body)
+	if err != nil {
+		log.Errorf("proxying content-add body errored: %s", err)
+	}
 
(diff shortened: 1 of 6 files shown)


elijaharita (approved):
my 3 brain cells don't notice any particular issues looking through. this'll be really nice to have!

cmd/shuttle-proxy/main.go
```diff
@@ -0,0 +1,177 @@
...
+	req.Header.Set("Shuttle-Proxy", "true")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	c.Response().WriteHeader(resp.StatusCode)
+
+	io.Copy(c.Response().Writer, resp.Body)
```
ribasushi, line 134: no error checks...?

cmd/shuttle-proxy/main.go
```diff
@@ -0,0 +1,177 @@
...
```
ribasushi, line 174: same: no error checks...

ribasushi (changes requested):
Few missing error checks, the rest looks legit

PR #75, opened 2021-11-11
allow setting filename and piece commitment when uploading car files

whyrusleeping:


handlers.go
```diff
@@ -488,6 +488,40 @@ func (s *Server) handleAddCar(c echo.Context, u *User) error {
...
+		commpSize = ss
+
+		cc, err := cid.Decode(cpc)
+		if err != nil {
+			return err
+		}
+
+		if cc.Prefix().Codec != cid.FilCommitmentUnsealed {
+			return fmt.Errorf("can only specify an unsealed commitment as a commP")
+		}
```
ribasushi, line 521: This entire thing can be replaced by https://pkg.go.dev/github.com/filecoin-project/go-fil-commcid#CIDToDataCommitmentV1 and just proxy the error (if any) in full. It does codc *and* mh checks.

ribasushi (approved):
LGTM

Left one suggestion, good to ship as-is

PR #78, opened 2021-11-24
add a sub-DAG selector for use in lotus retrieval

### LINKED ISSUE #76: Retrieval commands don't work in the verify webUI
If i try to retrieve a CID using the retrieval command returned by the CID verify webUI fails with:

`$ lotus client retrieve --miner f01392893 bafybeierkfdggz26tvbflaxrzgrlzb6k3g2ptv7hbkpskzxhoqtali2gtu data-2755366
ERROR: The received offer errored: retrieval query offer was unavailable:`

The lotus client retrieve command includes the CID of the file, while the correct way of retrieving a CID is to determine the rootCID of the CID I am trying to retrieve. 

The steps that work as follows: 
Determine the label of the dealID:
`$ lotus state get-deal 2755366
{
  "Proposal": {
    "PieceCID": {
      "/": "baga6ea4seaqawzo77rc75wrt7elirecptorakkv3txjhrmrkkxabvj6za2urogi"
    },
    "PieceSize": 2147483648,
    "VerifiedDeal": true,
    "Client": "f0397376",
    "Provider": "f01392893",
    "Label": "QmR6dciuJGeR5LzSuyn9nUu8pn6qPP9FTdaeeF2mWDzSi3",
    "StartEpoch": 1306869,
    "EndEpoch": 2801589,
    "StoragePricePerEpoch": "0",
    "ProviderCollateral": "330814207684018",
    "ClientCollateral": "0"
  },
  "State": {
    "SectorStartEpoch": 1287571,
    "LastUpdatedEpoch": -1,
    "SlashEpoch": -1
  }
}
`
use the CID in Label field: QmR6dciuJGeR5LzSuyn9nUu8pn6qPP9FTdaeeF2mWDzSi3
`lotus client retrieve --miner f01392893 QmR6dciuJGeR5LzSuyn9nUu8pn6qPP9FTdaeeF2mWDzSi3 QmR6dciuJGeR5LzSuyn9nUu8pn6qPP9FTdaeeF2mWDzSi3`
The problem with that approach I could end up downloading a whole DAG if I only need a single CID. 
To overcome this I have to add a path selector, which means I have to know where my CID is with the DAG beforehand.
`lotus client retrieve --miner f01392893 --datamodel-path-selector 'Links/0/Hash'  QmR6dciuJGeR5LzSuyn9nUu8pn6qPP9FTdaeeF2mWDzSi3 file.png
`

marshall:
this code may be naive, so I'm looking for feedback to make sure this is the best way to provide info needed by estuary-www.

the basic idea is, given the root CID of an aggregated DAG, and the content ID (PK of the `contents` table of the DB) of a specific file:
* fetch and sort all content IDs for the aggregated deal associated with the CID -- this sorted list represents the canonical index of the actual nodes in the aggregated DAG (root CID)
* match the specific file's content ID with it's index in the sorted list from above, and you now have an index to build the selector in the form `Links/$index/Hash`

after this lands, the accompanying estuary-www PR can land as well

closes #76

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -1112,28 +1112,30 @@ type getContentResponse struct {
 func (s *Server) calcSelector(aggregatedIn uint, contentID uint) (string, error) {
 	// sort the known content IDs aggregated in a CAR, and use the index in the sorted list
 	// to build the CAR sub-selector
-	rows, err := s.DB.Table("contents").
-		Select("id").
-		Where("aggregated_in = ?", aggregatedIn).
-		Order("id asc").
-		Rows()
-	defer rows.Close()
 
-	if err != nil {
-		return "", err
+	// SELECT ordinal - 1 FROM (
+	//   SELECT
+	//     id, ROW_NUMBER() OVER ( ORDER BY CAST(id AS TEXT) ) AS ordinal
+	//   FROM contents
+	//   WHERE aggregated_in = ?
+    // ) subq
+	// WHERE id = ?
+
+	subQuery := s.DB.Table("contents").
+		Select("id, ROW_NUMBER() OVER ( ORDER BY CAST(id as TEXT) ) AS ordinal").
+		Where("aggregated_in = ?", aggregatedIn)
+
+	var ordinal uint
+	result := s.DB.Table("(?) subq", subQuery).
+		Select("ordinal - 1").
+		Where("id = ?", contentID).
+		Find(&ordinal)
+
+	if result.Error != nil {
+		return "", result.Error 
 	}
 
-	index := 0
-	for rows.Next() {
-		var id uint
-		rows.Scan(&id)
-		if id == contentID {
-			return fmt.Sprintf("/Links/%d/Hash", index), nil
-		}
-		index += 1
-	}
-
-	return "", nil
+	return fmt.Sprintf("/Links/%d/Hash", ordinal), nil
 }
 
 func (s *Server) handleGetContentByCid(c echo.Context) error {
--- a/handlers.go
+++ b/handlers.go
@@ -1113,23 +1113,15 @@ func (s *Server) calcSelector(aggregatedIn uint, contentID uint) (string, error)
 	// sort the known content IDs aggregated in a CAR, and use the index in the sorted list
 	// to build the CAR sub-selector
 
-	// SELECT ordinal - 1 FROM (
-	//   SELECT
-	//     id, ROW_NUMBER() OVER ( ORDER BY CAST(id AS TEXT) ) AS ordinal
-	//   FROM contents
-	//   WHERE aggregated_in = ?
-    // ) subq
-	// WHERE id = ?
-
-	subQuery := s.DB.Table("contents").
-		Select("id, ROW_NUMBER() OVER ( ORDER BY CAST(id as TEXT) ) AS ordinal").
-		Where("aggregated_in = ?", aggregatedIn)
-
 	var ordinal uint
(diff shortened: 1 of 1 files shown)


handlers.go
```diff
@@ -1105,9 +1105,37 @@ func (s *Server) dealStatusByID(ctx context.Context, dealid uint) (*dealStatus,
...
+		rows.Scan(&id)
+		if id == contentID {
+			return fmt.Sprintf("/Links/%d/Hash", index), nil
+		}
+		index += 1
+	}
+
+	return "", nil
+}
+
```
ribasushi, line 1138: @marshall there is a gnarly nuance here: if you order by *numeric* ID you will get incorrect ordering on "number of digits in id" changes, since the directory assembler performs lexicographic sort over the filenames. So you want to `id::TEXT` the sort. 

But more to the point: the entire thing can be very efficiently be done in-rdbms, so I would just do something like:
```
SELECT ordinal - 1
FROM (
  SELECT
    id,
  ROW_NUMBER() OVER ( ORDER BY id::TEXT ) AS ordinal
  FROM contents
  WHERE aggregated_in = 9961183
) subq
WHERE id = 9961110
```

marshall: ahh yes, you mentioned that the first time we talked about it and I had forgotten the nuance. thank you for reminding me and the SQL code example

whyrusleeping: I think we should probably remove the `sort` call in the exising `createAggregate` method, we're sorting by ID there which gives the reader the impression that things will be ordered numerically, not text based.

handlers.go
```diff
@@ -1105,9 +1105,39 @@ func (s *Server) dealStatusByID(ctx context.Context, dealid uint) (*dealStatus,
...
+	subQuery := s.DB.Table("contents").
+		Select("id, ROW_NUMBER() OVER ( ORDER BY CAST(id as TEXT) ) AS ordinal").
+		Where("aggregated_in = ?", aggregatedIn)
+
+	var ordinal uint
+	result := s.DB.Table("(?) subq", subQuery).
+		Select("ordinal - 1").
+		Where("id = ?", contentID).
+		Find(&ordinal)
```
ribasushi, line 1132: I think doing all of this through gorm is a mistake. Why not just ( as per https://gorm.io/docs/sql_builder.html#Raw-SQL )
*bad formatting!!!*
```suggestion
  var ordinal uint
  result := s.DB.Raw(`
	 SELECT ordinal - 1 FROM (
	   SELECT
	     id, ROW_NUMBER() OVER ( ORDER BY CAST(id AS TEXT) ) AS ordinal
	   FROM contents
	   WHERE aggregated_in = ?
     ) subq
	 WHERE id = ?
  `,aggregatedIn, contentID ).Scan(&ordinal)
```

ribasushi: @marshall ping on ☝️ this?

marshall: sorry for the late reply on this, agreed w/ this suggestion. will send the updated PR post-haste

ribasushi (approved):
:shipit: 

Don't forget to add https://github.com/application-research/estuary-www/pull/18#issuecomment-978942965

marshall:
@ribasushi since application-research/estuary-www#18 was already merged, should I just send a follow up PR?

ribasushi:
@marshall yeah! I think it's important to highlight the "less overhead" option...

marshall:
new PR is up here: https://github.com/application-research/estuary-www/pull/22. merging!

PR #94, opened 2022-01-27
initial basic gateway reimplementation

whyrusleeping:


ribasushi (approved):
This is good, but what is best in life?

I assume the endpoint wiring comes in another PR? or you are adding more here?

whyrusleeping:
Yeah, I think ill add more in this PR unless you want to help review in pieces and I can just push each bit as a separate PR?

ribasushi:
keep adding

handlers.go
```diff
@@ -4302,3 +4305,49 @@ func (s *Server) handleRunGc(c echo.Context) error {
...
+		s.gwayHandler.ServeHTTP(c.Response().Writer, c.Request())
+		return nil
+	}
+
+	return c.Redirect(302, redir)
+}
+
+func (s *Server) checkGatewayRedirect(proto string, cc cid.Cid, segs []string) (string, error) {
+	if proto != "ipfs" {
+		return fmt.Sprintf("https://ipfs.io/%s/%s/%s", proto, cc, strings.Join(segs, "/")), nil
```
ribasushi, line 4330:
```suggestion
		return fmt.Sprintf("https://dweb.link/%s/%s/%s", proto, cc, strings.Join(segs, "/")), nil
```

util/gateway/gateway.go
```diff
@@ -0,0 +1,167 @@
...
+func NewGatewayHandler(bs blockstore.Blockstore) *GatewayHandler {
+	return &GatewayHandler{
+		bs:    bs,
+		dserv: merkledag.NewDAGService(blockservice.New(bs, nil)),
+	}
+}
+
+func (gw *GatewayHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if err := gw.handleRequest(r.Context(), w, r); err != nil {
+		w.Header().Set("Content-Type", "application/json")
```
ribasushi, line 40: I am pretty sure GW doesn't do that / and in general a random `curl`-er won't expect crap in their STDIN
Set the error as a header, and return an empty body?

cmd/estuary-shuttle/main.go
```diff
@@ -805,6 +809,16 @@ func (s *Shuttle) ServeAPI(listen string, logging bool) error {
 	e.GET("/health", s.handleHealth)
 	e.GET("/viewer", withUser(s.handleGetViewer), s.AuthRequired(util.PermLevelUser))
 
+	e.GET("/gw/:path", func(e echo.Context) error {
```
ribasushi, line 812: Maybe `ufsgw` to make it clearer it's "basic"? dunno...

handlers.go
```diff
@@ -4302,3 +4305,49 @@ func (s *Server) handleRunGc(c echo.Context) error {
...
+	if cont.Location == "local" {
+		return "", nil
+	}
+
+	var shuttle Shuttle
+	if err := s.DB.First(&shuttle, "handle = ?", cont.Location).Error; err != nil {
+		return "", err
+	}
+
+	// TODO: maybe check if the shuttle is online?
```
ribasushi, line 4350: Yeah, I'd do that, with a timed-memoize, like 30 secs or something, and then bat folks to regular dweb.link.

util/gateway/gateway.go
```diff
@@ -0,0 +1,167 @@
...
+		return cid.Undef, err
+	}
+
+	lnk, rem, err := nd.ResolveLink(segs)
+	if err != nil {
+		return cid.Undef, err
+	}
+
+	if len(rem) > 0 {
+		return cid.Undef, fmt.Errorf("pathing into ipld nodes not supported")
```
ribasushi, line 143: But I wannnnaaaaaaaaaaaaa :trollface: 
https://ipfs.io/ipfs/bafy2bzacecnamqgqmifpluoeldx7zzglxcljo6oja4vrmtj7432rphldpdmm2/8/1/8/1/0/1/0

util/gateway/gateway.go
whyrusleeping: the main gateway uses http.Error, I can switch to that no problem

util/gateway/gateway.go
whyrusleeping: no, bad riba

handlers.go
```diff
@@ -4302,3 +4305,53 @@ func (s *Server) handleRunGc(c echo.Context) error {
...
+
+const bestGateway = "dweb.link"
+
+func (s *Server) checkGatewayRedirect(proto string, cc cid.Cid, segs []string) (string, error) {
+	if proto != "ipfs" {
+		return fmt.Sprintf("https://%s/%s/%s/%s", bestGateway, proto, cc, strings.Join(segs, "/")), nil
+	}
+
+	var cont Content
+	if err := s.DB.First(&cont, "cid = ? and active and not offloaded", &util.DbCID{cc}).Error; err != nil {
```
ribasushi, line 4336: Is there value in checking for `active`? We are sending hem to the gateway, which will look in the swarm, and likely get back to you with a bitswap session. Might as well serve it if you have it...?
Or does `NOT active` potentially imply you no longer have all the blocks?

handlers.go
```diff
@@ -4302,3 +4305,53 @@ func (s *Server) handleRunGc(c echo.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	if redir == "" {
+		s.gwayHandler.ServeHTTP(c.Response().Writer, c.Request())
+		return nil
+	}
+
+	return c.Redirect(302, redir)
```
ribasushi, line 4325: Actually 302 is incorrect here, we should always 307 ( meaning: as a UA never store the destination, even for temporary sessions. Always come to estuary.tech first )
```suggestion
	return c.Redirect(307, redir)
```

ribasushi (approved):
Looks good. One ( I believe ) must-fix, another just a suggestion.

handlers.go
whyrusleeping: yeah, not active implies no blocks

whyrusleeping: either no blocks *yet* or no blocks for some other reason (like user deleted the thing)

ribasushi: ✔️

ribasushi:
@whyrusleeping this should be good to go no? ( aside from the `307` I still think we should change )

PR #95, opened 2022-01-28
add metadata to various APIs to support rclone

marshall:
my rclone estuary backend is in a new fork here:
https://github.com/application-research/rclone/tree/estuary

this change has two minor additions that make representing estuary's content/collections API as a filesystem a bit more bearable:

* Include `updatedAt` in the json serialization of `Content` which I'm using as the `modTime` for filesystem purposes
* Include the CID in results from `/collections/fs/list` for objects


whyrusleeping (approved):

whyrusleeping:
Nice, can't wait to try rclone stuff out

PR #96, opened 2022-01-31
boost integration

dirkmc:
This PR modifies Estuary to be able to make deals with [Boost](https://boost.filecoin.io/)

Boost listens on v1.2.0 of the deal proposal protocol. So to make deals with Boost, Estuary needs to
- Connect to the Storage Provider
- Check if the SP supports deal proposal protocol v1.2.0
- Create an auth token and tell the shuttle with the data for the deal to accept a request with that auth token
- Send a deal proposal to the SP running Boost, including the auth token and download address
- Boost then downloads the data from the shuttle

The new version of the deal proposal protocol also needs the size of the CAR file (ie, the size of data that will be transferred across the wire). So there are some changes to calculate the CAR file size of deal data before making a deal proposal.

TODO:
- [ ] The estuary primary node gets the announce address from the shuttle. For boost we need the announce address to be public. If the announce address is public does that mean that calls from the primary node to the shuttle with go through the internet (instead of local network)?

Depends on https://github.com/application-research/filclient/pull/60


replication.go
```diff
@@ -2400,12 +2499,33 @@ func (cm *ContentManager) lookupPieceCommRecord(data cid.Cid) (*PieceCommRecord,
 	return &pcr, nil
 }
 
+// calculateCarSize works out the CAR size using the cids and block sizes
+// for the content stored in the DB
+func (cm *ContentManager) calculateCarSize(data cid.Cid) (uint64, error) {
```
whyrusleeping, line 2504: can we add a trace to this? its a potential slow spot we might wanna track

replication.go
```diff
@@ -2067,57 +2062,79 @@ func (cm *ContentManager) makeDealsForContent(ctx context.Context, content Conte
...
+		if err := cm.DB.Create(cd).Error; err != nil {
+			return xerrors.Errorf("failed to create database entry for deal: %w", err)
+		}
+
+		// Send the deal proposal to the storage provider
+		var cleanupDealPrep func() error
+		var propPhase bool
+		isPushTransfer := proto == filclient.DealProtocolv110
+		switch {
+		case proto == filclient.DealProtocolv110:
```
dirkmc, line 2088: TODO: `switch proto {` ...

replication.go
```diff
@@ -2067,57 +2062,79 @@ func (cm *ContentManager) makeDealsForContent(ctx context.Context, content Conte
...
+		dealUUID := uuid.New()
+		cd := &contentDeal{
+			Content:  content.ID,
+			PropCid:  util.DbCID{propnd.Cid()},
+			DealUUID: dealUUID.String(),
+			Miner:    ms[i].String(),
+			Verified: verified,
+		}
+
+		if err := cm.DB.Create(cd).Error; err != nil {
```
dirkmc, line 2079: TODO: instead of creating the database row and then deleting it if there's an error, can we just use the deal UUID as an index?

dirkmc: Tricky to do because:
- Once an SP running Boost receives and accepts a deal proposal it immediately starts downloading from estuary
- When the download starts, and as it progresses, the shuttle makes RPC calls to the primary estuary node
- The handlers for these RPC calls [need the deal to already have been created in the DB](https://github.com/application-research/estuary/blob/ab982078721d568102186d2aae7bf3c1126b2163/shuttle.go#L302-L313) so they can update its state with transfer progress

gmelodie:
@dirkmc can you give me more context on what this does? I have no clue what's going on with this PR as well as what it does.

dirkmc:
@gmelodie sorry about that, I updated the PR description, let me know if you need any more context

cmd/estuary-shuttle/main.go
```diff
@@ -476,9 +504,9 @@ func main() {
 				var received uint64
 
 				for _, xfer := range txs {
-					byState[xfer.Status()]++
-					sent += xfer.Sent()
-					received += xfer.Received()
+					byState[xfer.Status]++
```
en0ma, line 507: why this change?

cmd/estuary-shuttle/main.go
```diff
@@ -1876,7 +1931,7 @@ func writeAllGoroutineStacks(w io.Writer) error {
 
 func (s *Shuttle) handleRestartAllTransfers(e echo.Context) error {
 	ctx := e.Request().Context()
-	transfers, err := s.Filc.TransfersInProgress(ctx)
+	transfers, err := s.Filc.V110TransfersInProgress(ctx)
```
en0ma, line 1934: what is the difference between both versions?

cmd/estuary-shuttle/main.go
```diff
@@ -378,6 +379,33 @@ func main() {
 			}
 		})
 
+		_, err = s.Filc.Libp2pTransferMgr.Subscribe(func(dbid uint, st filclient.ChannelState) {
```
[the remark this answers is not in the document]
dirkmc: These are two different event subscriptions:
- `s.Filc.SubscribeToDataTransferEvents` subscribes to events from legacy go-fil-markets
- `s.Filc.Libp2pTransferMgr.Subscribe` subscribes to events from Boost

cmd/estuary-shuttle/main.go
dirkmc: The format of the data returned by TransfersInProgress [has changed slightly](https://github.com/application-research/filclient/pull/60/files#diff-1dfb05953decb16f2625b57212a3b0c11b22110804a0b24c4ec2cba68162c3f2R940) - it now returns a ChannelState object which has a field called Status instead of a method.

cmd/estuary-shuttle/main.go
dirkmc: `V110TransfersInProgress` returns transfers in progress for legacy markets deals.
`TransfersInProgress` returns _all_ transfers in progress (both legacy markets deals and Boost deals).
Note that in Boost deals the Storage Provider downloads from the client (estuary), so the restart happens on the Storage Provider side. For legacy deals the restart happens on the client (estuary) side.

cmd/estuary-shuttle/main.go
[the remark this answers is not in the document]
en0ma: I see -  a comment about this difference could be really helpful.

cmd/estuary-shuttle/main.go
en0ma: cool, I see why the filclient go mod update

en0ma:
@dirkmc LGTM - great stuff. I think when you resolve the conflicts, I can approve

cmd/estuary-shuttle/main.go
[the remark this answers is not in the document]
dirkmc: I added a comment to explain the difference to the code

en0ma:
@dirkmc can we have a config to turn `on boost (pull)`/ `off boost (push - go-fil-markets)`

gmelodie:
Functionality is looking solid. For anyone that wants to test this out I compiled the steps to make it work: https://gist.github.com/gmelodie/429f2c49eefbcb9016e4f987371320ba

gmelodie (approved):
LGTM

dirkmc:
> can we have a config to turn on boost (pull)/ off boost (push - go-fil-markets)

@en0ma can you please explain the use case for this request

Note that with this PR Estuary will make Boost deals with providers running Boost, and legacy go-fil-markets deals with providers who are not running Boost.

en0ma:
> > can we have a config to turn on boost (pull)/ off boost (push - go-fil-markets)
> 
> @en0ma can you please explain the use case for this request
> 
> Note that with this PR Estuary will make Boost deals with providers running Boost, and legacy go-fil-markets deals with providers who are not running Boost.

@dirkmc there is no use case per see. I am thinking it will be "nice to have" flexibility for anyone running the estuary node to say they want only go-fil-makerts. But this is not necessary, just thinking out loud.

whyrusleeping:
@dirkmc it looks like we still need to update to the latest merged version of filclient now that its landed

PR #99, opened 2022-02-09
support filesystem paths in shuttle /content/add API

marshall:
this brings estuary-shuttle's `/content/add` API up to parity with the primary node's endpoint of the same name, adding support for arbitrary filesystem paths with the `collectionPath` parameter. this required changes in both the shuttle and primary node, and is needed to make uploading with rclone work for both the primary and shuttle upload endpoints

summary of changes:
   - refactored content request/response body types into util/content.go
   - removed []Collection arguments in favor of a single collection w/ a single path (see ContentInCollection)
   - forward along `collectionPath` from the shuttle to the primary node's `/content/create`
   - add handling for `collectionPath` in `/content/create`
   - added support for http shuttle URLs in UploadEndpoints in dev mode

the matching rclone client-side work to support upload endpoints has also landed here for reference:
https://github.com/application-research/rclone/commit/a25e393f7f3ef77e78a94409c7cf87984b9fb780

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,3 +140,7 @@ dist-clean:
 
 calibnet: GOFLAGS+=-tags=calibnet
 calibnet: build
+
+.PHONY: test
+test:
+	go test $(GOFLAGS) -v ./...
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -652,9 +652,14 @@ func (d *Shuttle) getHelloMessage() (*drpc.Hello, error) {
 		return nil, err
 	}
 
-	log.Infow("sending hello", "hostname", d.hostname, "address", addr, "pid", d.Node.Host.ID())
+	hostname := d.hostname
+	if d.dev {
+		hostname = "http://" + d.hostname
+	}
+
+	log.Infow("sending hello", "hostname", hostname, "address", addr, "pid", d.Node.Host.ID())
 	return &drpc.Hello{
-		Host:    d.hostname,
+		Host:    hostname,
 		PeerID:  d.Node.Host.ID().Pretty(),
 		Address: addr,
 		Private: d.Private,
@@ -936,7 +941,10 @@ func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 
 	defer fi.Close()
 
-	collection := c.FormValue("collection")
+	cic := util.ContentInCollection{
+		Collection:     c.FormValue("collection"),
+		CollectionPath: c.FormValue("collectionPath"),
+	}
 
 	bsid, bs, err := s.StagingMgr.AllocNew()
 	if err != nil {
@@ -959,7 +967,7 @@ func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 		return err
 	}
 
-	contid, err := s.createContent(ctx, u, nd.Cid(), fname, collection)
+	contid, err := s.createContent(ctx, u, nd.Cid(), fname, cic)
 	if err != nil {
 		return err
 	}
@@ -989,7 +997,7 @@ func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 		log.Warn(err)
 	}
 
-	return c.JSON(200, &util.AddFileResponse{
+	return c.JSON(200, &util.ContentAddResponse{
 		Cid:       nd.Cid().String(),
 		EstuaryId: contid,
 		Providers: s.addrsForShuttle(),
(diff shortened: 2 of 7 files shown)


marshall:
friendly poke to @whyrusleeping and @gmelodie in case you also want to take a look

whyrusleeping (approved):
LGTM

Makefile
```diff
@@ -140,3 +140,7 @@ dist-clean:
 
 calibnet: GOFLAGS+=-tags=calibnet
 calibnet: build
+
+.PHONY: test
+test:
+	go test $(GOFLAGS) -v ./...
```
gmelodie, line 146: Can you clarify what this does?

cmd/estuary-shuttle/main.go
```diff
@@ -652,9 +652,14 @@ func (d *Shuttle) getHelloMessage() (*drpc.Hello, error) {
...
-	log.Infow("sending hello", "hostname", d.hostname, "address", addr, "pid", d.Node.Host.ID())
+	hostname := d.hostname
+	if d.dev {
+		hostname = "http://" + d.hostname
+	}
+
+	log.Infow("sending hello", "hostname", hostname, "address", addr, "pid", d.Node.Host.ID())
 	return &drpc.Hello{
-		Host:    d.hostname,
+		Host:    hostname,
```
gmelodie, line 662: Nit: maybe leave as it was

cmd/estuary-shuttle/main.go
```diff
@@ -1944,9 +1939,10 @@ func (s *Shuttle) handleGetWantlist(c echo.Context) error {
 }
 
 type importDealBody struct {
-	Name       string   `json:"name"`
-	DealIDs    []uint64 `json:"dealIDs"`
-	Collection string   `json:"collection"`
+	util.ContentInCollection
```
gmelodie, line 1942: I'm curious how this works on Golang. Did you not name this field? What would its name be?

handlers.go
```diff
@@ -3689,16 +3671,27 @@ func (s *Server) handleCreateContent(c echo.Context, u *User) error {
...
 		}).Error; err != nil {
 			return err
 		}
 	}
 
-	return c.JSON(200, createContentResponse{
+	return c.JSON(200, util.ContentCreateResponse{
 		ID: content.ID,
 	})
 }
```
gmelodie, line 3697: What happens if `req.Collection != ""` (aka a collection uuid is present) but `req.CollectionPath == ""` (aka no or empty collection path provided)? I _think_ we'll need a default path (`/filename`) to be set.

gmelodie (commented):
My comments are mostly questions about how things work haha. Great work!

parkan:
looks great! is there an explanation/design doc of how this will work with rclone? will it embed a local "shuttle" or upload to a remote one?

Makefile
marshall: this is a convenience make target, so you can type less characters :) it doesn't bear much importance to this PR, but will be helpful when we add more tests soon.

make targets by default are files on the filesystem -- the `.PHONY` pseudo-target tells make that `test` doesn't actually refer to a file on the filesystem, i.e. it is a virtual target

i.e
`$ make test`
instead of
`$ go test -v ./...`

cmd/estuary-shuttle/main.go
marshall: yeah it isn't a named field at all, but an embedded type. it's a very Go-specific approach to common data fields and logic, since there is no such thing as a class hierarchy.

you can refer to fields on the embedded type by the type name.. more info here:
https://go.dev/doc/effective_go#embedding

handlers.go
marshall: since `path` is a pointer, it's initial value would be `nil`. since `req.CollectionPath == ""`, the pointer never gets updated, and in the call to `DB.create`, the value of the `Path` db column will also be nil, leaving it in it's default state (which should be what is expected here)

marshall:
> looks great! is there an explanation/design doc of how this will work with rclone? will it embed a local "shuttle" or upload to a remote one?

@parkan docs are coming once this is stabilized! this PR paves the way to allow rclone to upload to estuary via shuttle on the public instance, so no need for local shuttle or local estuary node.

parkan:
@marshall ahh I was hoping the backend would embed filc (paralleling the design we're discussing with @ribasushi) so the client would do the data transfer without going through an intermediary, similar to the native uplink/tardigrade mode here https://github.com/rclone/rclone/blob/f22b703a516f0a8ac99fed7a12810cd4ccfac300/backend/storj/fs.go

slightly different use case -- I guess the approach here is more akin go "gateway" mode

handlers.go
gmelodie: Makes sense! I think we definitely need to revisit this whole path logic some time in the future though.

marshall:
@parkan oooh that's interesting. thanks for the link!

my plan is to embed filc in the rclone backend for retrieval in the next phase of the PoC, but I hadn't considered using it for direct uplink. it isn't clear to me how we would go straight to chain w/ the current estuary architecture, maybe by putting the duplication + deal making logic in an on-chain actor?

parkan:
@marshall I think we worked out a pattern where an estuary node would coordinate certain things (like miner selection) and then filc would do the rest, will check in on progress w @ribasushi tmw

PR #100, opened 2022-02-10
Fix directory CID listing as file

gmelodie:
As explained [here](https://gist.github.com/marshall/e7b7201400f1339dd51e904d579b46ad), when listing the contents of a collection using `http://localhost:3004/collections/fs/list?col=my-collection-uuid&dir=/`, we found that all CIDs of directories that were added through `/content/add-ipfs` were showing as files instead of directories. This PR fixes that behavior, however it still does not support listing the contents of those directories as of now.

### Reproducing
I wrote [a small PoC for this](https://gist.github.com/gmelodie/3027b9297b53ebbe2ffd7f9d703f2d86). Here's how to use it:
1. Setup `./estuary`, `estuary-shuttle` and your `ipfs daemon` (obs: not sure if the shuttle is necessary but good measure)
2. Create a collection and save its UUID (you can use the `/collections/create` endpoint for that
```bash
curl -X POST http://localhost:3004/collections/create -d '{ "name": "A collection name here", "description": "A collection test" }' -H "Content-Type: application/json" -H "Authorization: Bearer MY-API-KEY-HERE"
```
3. Edit the PoC script to add your new collection's UUID to it.
4. Run the PoC
```bash
./pin-dir-est.sh
```

### Fixed
Before:
```bash
> curl -X GET -H "Authorization: Bearer $APIKEY" "http://localhost:3004/collections/fs/list?col=cadf85f5-af32-41ec-8373-6de3fd54c0c4&dir=/" 2>/dev/null | 
[
  {
    "name": "testdir",
    "dir": false,
    "size": 676,
    "contId": 14,
    "cid": "QmcKmsYCWgvx1ypqgpo88NqSt4Xw93RXATshzspXd4mTZi"
  }

```
Now (fixed):
```bash
> curl -X GET -H "Authorization: Bearer $APIKEY" "http://localhost:3004/collections/fs/list?col=cadf85f5-af32-41ec-8373-6de3fd54c0c4&dir=/" 2>/dev/null | jq     
[
  {
    "name": "testdir",
    "dir": true,
    "size": 676,
    "contId": 12,
    "cid": "QmcKmsYCWgvx1ypqgpo88NqSt4Xw93RXATshzspXd4mTZi"
  }
]
```

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -40,7 +40,8 @@ import (
 	offline "github.com/ipfs/go-ipfs-exchange-offline"
 	ipld "github.com/ipfs/go-ipld-format"
 	logging "github.com/ipfs/go-log/v2"
-	"github.com/ipfs/go-merkledag"
+	merkledag "github.com/ipfs/go-merkledag"
+	unixfs "github.com/ipfs/go-unixfs"
 	uio "github.com/ipfs/go-unixfs/io"
 	car "github.com/ipld/go-car"
 	"github.com/labstack/echo/v4"
@@ -4201,8 +4202,37 @@ func (s *Server) handleColfsListDir(c echo.Context, u *User) error {
 			return err
 		}
 
+		// Build a new local DAGService to query for the CID if our ref and see if it's a directory
+		bserv := blockservice.New(s.Node.Blockstore, offline.Exchange(s.Node.Blockstore))
+		dserv := merkledag.NewDAGService(bserv)
+		ctx := c.Request().Context()
+		nd, err := dserv.Get(ctx, r.Cid.CID)
+		if err != nil {
+			return err
+		}
+
 		// if the relative path requires pathing up, its definitely not in this dir
 		if strings.HasPrefix(relp, "..") {
+			continue
+		}
+
+		// TODO: maybe find a way to reuse s.Node or s.gwayHandler.dserv
+		// Need to transform node into a unixfs node to check if it's dir
+		fsNode, err := unixfs.ExtractFSNode(nd)
+		if err != nil {
+			return err
+		}
+		if fsNode.IsDir() { // if CID is a dir
+			if !dirs[relp] {
+				dirs[relp] = true
+				out = append(out, collectionListResponse{
+					Name:   relp,
+					Dir:    true,
+					Size:   r.Size,
+					ContID: r.ContID,
+					Cid:    &r.Cid,
+				})
+			}
 			continue
 		}
 
@@ -4217,7 +4247,7 @@ func (s *Server) handleColfsListDir(c echo.Context, u *User) error {
 			continue
 		}
 
-		parts := strings.Split(relp, "/")
+		parts := strings.Split(relp, "/") // if it is a subcollection
 		if !dirs[parts[0]] {
 			dirs[parts[0]] = true
 			out = append(out, collectionListResponse{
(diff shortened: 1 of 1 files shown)


handlers.go
```diff
@@ -4208,11 +4209,40 @@ func (s *Server) handleColfsListDir(c echo.Context, u *User) error {
 			return err
 		}
 
+		// Build a new local DAGService to query for the CID if our ref and see if it's a directory
+		bserv := blockservice.New(s.Node.Blockstore, offline.Exchange(s.Node.Blockstore))
+		dserv := merkledag.NewDAGService(bserv)
+		ctx := c.Request().Context()
+		nd, err := dserv.Get(ctx, r.Cid.CID)
```
marshall, line 4216: hrm, is it possible that trying to look up the CID from the local blockstore fails if pinning in still in progress in `ContentManager`?

if so, I think we may just want to either exclude the `CollectionRef` in the results, or revert to previous behavior instead. as it stands it looks like any CID lookup failure in the local blockstore will make the collection fs list fail altogether

handlers.go
```diff
@@ -4208,11 +4209,40 @@ func (s *Server) handleColfsListDir(c echo.Context, u *User) error {
...
 		// if the relative path requires pathing up, its definitely not in this dir
 		if strings.HasPrefix(relp, "..") {
 			continue
 		}
 
+		// TODO: maybe find a way to reuse s.Node or s.gwayHandler.dserv
+		// Need to transform node into a unixfs node to check if it's dir
+		fsNode, err := unixfs.ExtractFSNode(nd)
+		if err != nil {
+			return err
```
marshall, line 4230: similar to my above comment -- I assume it might be ok that extracting an FS node from a CID is allowed to fail for raw nodes, and that shouldn't (necessarily) fail to list the whole directory. in that case, it seems like listing it as a file would be just fine

marshall (changes requested):
great detective work @gmelodie!

I left a few questions/comments about the failure logic.. once those are addressed I think we have a good path forward :)

handlers.go
whyrusleeping: Yeah, this is an interesting edge case. If content is offloaded then listing will fail. Maybe we should encode some additional information in the collectionRef?

We could also include an extra field in the collectionListResponse to indicate the content is not current accessible on estuary? call it 'Active' or not

handlers.go
gmelodie: Hummm I don't know if I'm too picky with myself, but listing it as a file just doesn't seem good enough. I like the point about not failing all the query though. Maybe we could change the `File: true/false` to a field like `Type: file/dir/undefined`?

handlers.go
marshall: @gmelodie that sounds totally reasonable to me!

whyrusleeping: the pin status endpoint gets you that info, see `handleGetPin`

whyrusleeping: and +1 to your suggestion there, seems reasonable to not list it if its in progress

handlers.go
gmelodie: So I changed it to this: if it can be cast to a unixfs node then we can determine whether its `CidType` is a `File` or a `Dir`, otherwise it's set to `Raw`

marshall (approved):
looks great, thanks for your patience @gmelodie

whyrusleeping (approved):

jimmylee (approved):
LGTM

PR #101, opened 2022-02-12
Fix collection/filename/directory inconsistencies

gmelodie:
I found some bugs while poking around `add-ipfs` and, since they were pretty similar and related, decided to make the fixes in a single PR.

**Edit:** 1, 2 and 4 are now present on #103

~1. Estuary crashes when providing `"collectionPath": ""` in `/add-ipfs`(on `sanitizePath`, in `handlers.go`)~

~2. When `collectionPath` is `nil`, the file becomes unseeable from `/content/fs/list`. It's on the database, but since querying will always take a path (even when you don't provide any it'll default to `/`), it will never show our file.~

3. file in root `/` showing as `.` (using relative path to get filename, now using filename from field `name` of table `contents`)

~4. adding two CIDs on the same directory will make listing show the same CID for both array elements (the one from the newer CID). Fixed with `&util.DbCID{r.Cid.CID}`.~

5. Files with the same path appear overwritten when listing. `collection_refs` should be `path` + `/` + `filename`, not something else.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -497,7 +497,7 @@ func (s *Server) handleAddIpfs(c echo.Context, u *User) error {
 		// default: colp ends in / (does not include filename e.g. /hello/)
 		pathWithFilename := colp + filename
 
-		// if path is does not end in /, it includes the filename
+		// if path does not end in /, it includes the filename
 		if !strings.HasSuffix(colp, "/") {
 			pathWithFilename = colp
 			filename = filepath.Base(colp)
@@ -4746,7 +4746,7 @@ func sanitizePath(p string) (string, error) {
 	}
 
 	if p[0] != '/' {
-		return "", fmt.Errorf("all paths must be absolute")
+		return "", fmt.Errorf("paths must start with /")
 	}
 
 	// TODO: prevent use of special weird characters
--- a/handlers.go
+++ b/handlers.go
@@ -4586,11 +4586,12 @@ func openApiMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
 }
 
 type collectionListQueryRes struct {
-	ContID uint
-	Cid    util.DbCID
-	Size   int64
-	Path   *string
-	Type   util.ContentType
+	ContID   uint
+	Cid      util.DbCID
+	Size     int64
+	Path     *string
+	Type     util.ContentType
+	Filename *string
 }
 
 type CidType string
@@ -4644,7 +4645,7 @@ func (s *Server) handleColfsListDir(c echo.Context, u *User) error {
 	if err := s.DB.Model(CollectionRef{}).
 		Joins("left join contents on contents.id = collection_refs.content").
 		Where("collection = ?", col.ID).
-		Select("contents.id as cont_id, contents.cid as cid, path, size, contents.type").
+		Select("contents.id as cont_id, contents.cid as cid, contents.name as filename, path, size, contents.type").
 		Scan(&refs).Error; err != nil {
 		return err
 	}
@@ -4706,7 +4707,7 @@ func (s *Server) handleColfsListDir(c echo.Context, u *User) error {
 				contentType = Dir
 			}
 			out = append(out, collectionListResponse{
-				Name:   filepath.Base(relp),
+				Name:   *r.Filename,
 				Type:   contentType,
 				Size:   r.Size,
 				ContID: r.ContID,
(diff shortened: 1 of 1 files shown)


handlers.go
```diff
@@ -426,25 +426,34 @@ func (s *Server) handleAddIpfs(c echo.Context, u *User) error {
...
+		colp := &defaultPath
+		if params.CollectionPath != nil && *params.CollectionPath != "" {
 			p, err := sanitizePath(*params.CollectionPath)
 			if err != nil {
 				return err
 			}
 			colp = &p
 		}
+		// TODO: path is a dir or is the entire path with file? (if we put / the file is named . when listing)
+		pathWithFilename := *colp + filename
```
marshall, line 452: FWIW, the path as it's stored in the database is the absolute path, including filename :)

handlers.go
```diff
@@ -426,25 +426,34 @@ func (s *Server) handleAddIpfs(c echo.Context, u *User) error {
 		return err
 	}
 
+	filename := params.Name
+	if filename == "" {
+		filename = params.Root
+	}
+
```
marshall, line 433: hrm.. as it stands, the `/content/add` API expects `collectionPath` to be the absolute path of the content you're uploading, including it's filename. I just did a bunch of parity fixes to make sure that is the case across both main node and shuttle in #99..

we should probably get on the same page about the API convention before merging, so these behave as expected

marshall (changes requested):
nice work! lets touch base about the behavior of how `collectionPath` works before merging so the behavior is consistent

handlers.go
gmelodie: Are you sure? I remember one of the issues was exactly when you submitted an empty (`""`, not `nil`) filename when uploading a file.

handlers.go
gmelodie: > we should probably get on the same page about the API convention before merging, so these behave as expected

Absolutely! This bit of code is old though, I just replaced it a bit higher up because I needed `filename` to be set on a function call IIRC

marshall: yeah, sorry -- I should've highlighted below where you append `filename` to the collection path

handlers.go
```diff
@@ -4215,7 +4218,8 @@ func (s *Server) handleColfsListDir(c echo.Context, u *User) error {
 
 		if !strings.Contains(relp, "/") {
 			out = append(out, collectionListResponse{
-				Name:   filepath.Base(relp),
+				// Name:   filepath.Base(relp),
+				Name:   *r.Filename,
```
whyrusleeping, line 4222: do we have confidence this will never be nil?

gmelodie: Good one! In https://github.com/application-research/estuary/blob/67fce679308c584eef3aa5833ff426ae8b429f40/handlers.go#L476-L479 there's a check for the empty string, I assumed that if you don't specify a `name` field on the request it gets set as an empty string. Is that assumption right?

marshall: FWIW, this is more a question about the database than the request handler. if we have any database rows with a `nil` Filename. maybe to protect against that we can just add a check on line 4196, i.e...

```go
		if r.Path == nil || r.Filename == nil {
				continue
		}
```

gmelodie: Oh I see! Makes sense

whyrusleeping:
The thing this PR does that i'm not sure about is make it impossible to add files to a collection without a path. I had initially intended to allow contents to be loosely associated with a collection, like a big bag of things. But i'm not really sure thats entirely useful at this point. Happy to move in that direction, but want us to be cognizant that we are changing behavior

gmelodie:
There are two points I'd like to address: things not making sense, and things changing.

First of all, this PR makes it more intuitive that every content will get a path, even if it is a default one. Personally, it took me a long time to figure out the difference between files that had a path and files that didn't. The "big bag of things" part is exactly what made collections so confusing to me and to other Estuary users as well. Any directory _already is_ a big bag of things. Collections continue to be "big bags of things". Design is all about what you expect. All users of Linux/Windows/Mac expect a file to have a path, that is the default thought most of our brains have. That's what we _expect_. Things **not** having a path is weird, not the other way around.

Secondly, about the functionality changing. I don't think this changes all that much to be honest. There are two use cases this PR _could_ change: adding files, and listing files. None of these operations changed as far as the user is concerned.
1. For adding files, instead of being able to add them without a path, now they have a default path of `/`
2. For listing files "without a path", which was done through `/collections/content/:collection-id`, that also remains the same. 

If a user doesn't care about paths, then they simply don't specify `collectionPath` and keep using collections as they are.

PR #103, opened 2022-02-14
Fix ipfs-add endpoint bugs

gmelodie:
This PR contains changes from #101  that are "harmless" (aka don't change API behavior, and are backwards compatible). As discussed with @marshall 

1. Estuary crashes when providing `"collectionPath": ""` in `/add-ipfs` (on `sanitizePath`, in `handlers.go`)

2. When `collectionPath` is `nil`, the file becomes unseeable from `/content/fs/list`. It's on the database, but since querying will always take a path (even when you don't provide any it'll default to `/`), it will never show our file. Now it defaults to `/my-filename`.

3. Adding two CIDs on the same directory will make listing show the same CID for both array elements (the one from the newer CID). Fixed with `&util.DbCID{r.Cid.CID}`.

Obs: don't quote me on the "harmless" thing about my code... ever

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -429,7 +429,7 @@ func (s *Server) handleAddIpfs(c echo.Context, u *User) error {
 		defaultPath := "/" + params.Name
 		colp := &defaultPath
 		if params.CollectionPath != "" {
-			p, err := sanitizePath(*params.CollectionPath)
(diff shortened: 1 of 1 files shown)


marshall (approved):
looks good!

PR #106, opened 2022-02-18
Update go.yml

gmelodie:
One of the tests (and probably more in the future) needs the `ESTUARY_TOKEN` to be set. This generates and exports it.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -32,5 +32,10 @@ jobs:
     - name: Build
       run: make clean all
 
+    - name: Setup
+      run: echo "ESTUARY_TOKEN=$(./estuary setup | grep Token | cut -d ' ' -f 3)" >> $GITHUB_ENV
+
     - name: Test
       run: go test -v ./...
+      env:
+        ESTUARY_TOKEN: $ESTUARY_TOKEN


alvin-reyes (commented):
I'm thinking we can use the `env` to set the `ESTUARY_TOKEN`

```
jobs:
       - name: Test
          run: .....
          env:
            ESTUARY_TOKEN: $ESTUARY_TOKEN
```

.github/workflows/go.yml
```diff
@@ -32,5 +32,8 @@ jobs:
     - name: Build
       run: make clean all
 
+    - name: Setup
+      run: echo "ESTUARY_TOKEN=$(./estuary setup | grep Token | cut -d ' ' -f 3)" >> $GITHUB_ENV
+
     - name: Test
-      run: go test -v ./...
+      run: ESTUARY_TOKEN=$ESTUARY_TOKEN go test -v ./...
```
marshall, line 39: hrm, `ESTUARY_TOKEN` should already be in the environment, i.e. adding `ESTUARY_TOKEN=$ESTUARY_TOKEN` before `go test` seems redundant.

if I'm reading the [docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-6) correctly, you can probably just omit the this altogether

marshall (approved):
looks good! check my comment about the `ESTUARY_TOKEN` env var, and see if it makes sense, but otherwise this looks good to merge

.github/workflows/go.yml
gmelodie: Correct! Done and tested [here](https://github.com/gmelodie/estuary/actions/runs/1876540621)

gmelodie:
> I'm thinking we can use the env to set the ESTUARY_TOKEN

Absolutely! Although just leaving it out would work, I prefer this as it's more verbose and obvious.

PR #108, opened 2022-02-21
Add CID directory database tracking (disallow listings)

gmelodie:
This PR addresses the issues raised [here](https://application-research.canny.io/estuary-bugs/p/support-pinned-ipfs-dirs-in-collectionsfslist). 
1. We now hold on the database (under the `contents` table) whether a pinned CID is a dir or a file, and show that when listing the contents of a collection (on `handleColfsListDir`). 
2. We also disallow listing directories added by CID (what we're calling "CID dirs") since the files under the root CID might not be on the same machine as the Estuary main node (aka on the shuttles) and listing them might be too expensive. 

**Obs:** we also thought about expanding the directory as it is pinned, but keeping metadata about all the files under a CID recursively on the database might not be feasible in terms of computational power.

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -1184,11 +1184,16 @@ func (s *Shuttle) addrsForShuttle() []string {
 }
 
 func (s *Shuttle) createContent(ctx context.Context, u *User, root cid.Cid, fname string, cic util.ContentInCollection) (uint, error) {
+	bserv := blockservice.New(s.Node.Blockstore, s.Node.Bitswap)
+	dserv := merkledag.NewDAGService(bserv)
+	contentType := util.FindCIDType(root, dserv)
+
 	data, err := json.Marshal(util.ContentCreateBody{
 		ContentInCollection: cic,
 		Root:                root,
 		Name:                fname,
 		Location:            s.shuttleHandle,
+		Type:                contentType,
 	})
 	if err != nil {
 		return 0, err
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -1186,7 +1186,7 @@ func (s *Shuttle) addrsForShuttle() []string {
 func (s *Shuttle) createContent(ctx context.Context, u *User, root cid.Cid, fname string, cic util.ContentInCollection) (uint, error) {
 	bserv := blockservice.New(s.Node.Blockstore, s.Node.Bitswap)
 	dserv := merkledag.NewDAGService(bserv)
-	contentType := util.FindCIDType(root, dserv)
+	contentType := util.FindCIDType(ctx, root, dserv)
 
 	data, err := json.Marshal(util.ContentCreateBody{
 		ContentInCollection: cic,
--- a/dbmgr/dbmgr.go
+++ b/dbmgr/dbmgr.go
@@ -341,6 +341,7 @@ type Content struct {
 	UserID      UserID
 	Description string
 	Size        int64
+	Type        util.ContentType
 	Active      bool
 	Offloaded   bool
 	Replication int
--- a/handlers.go
+++ b/handlers.go
@@ -42,7 +42,6 @@ import (
 	ipld "github.com/ipfs/go-ipld-format"
 	logging "github.com/ipfs/go-log/v2"
 	merkledag "github.com/ipfs/go-merkledag"
-	unixfs "github.com/ipfs/go-unixfs"
 	uio "github.com/ipfs/go-unixfs/io"
 	car "github.com/ipld/go-car"
 	"github.com/labstack/echo/v4"
@@ -838,7 +837,7 @@ func (cm *ContentManager) addDatabaseTrackingToContent(ctx context.Context, cont
 		return err
 	}
 
-	if err = cm.addObjectsToDatabase(ctx, cont, objects, "local"); err != nil {
+	if err = cm.addObjectsToDatabase(ctx, cont, dserv, root, objects, "local"); err != nil {
 		return err
 	}
 
@@ -3667,10 +3666,15 @@ func (s *Server) handleCreateContent(c echo.Context, u *User) error {
 		}
 	}
 
+	bserv := blockservice.New(s.Node.Blockstore, s.Node.Bitswap)
+	dserv := merkledag.NewDAGService(bserv)
+	contentType := util.FindCIDType(context.Background(), req.Root, dserv)
+
 	content := &Content{
 		Cid:         util.DbCID{req.Root},
 		Name:        req.Name,
 		Active:      false,
+		Type:        contentType,
 		Pinning:     false,
 		UserID:      u.ID,
 		Replication: defaultReplication,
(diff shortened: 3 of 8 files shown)


dbmgr/dbmgr.go
```diff
@@ -341,6 +341,7 @@ type Content struct {
 	UserID      UserID
 	Description string
 	Size        int64
+	Dir         bool
```
whyrusleeping, line 344: We probably want to have this be more than a bool, Users can (and do) pin things outside of files and directories, like blockchain segments, random ipld graphs, cbor maps of things, and so on.
Maybe for now we can make it be an integer field called `Kind` ('Type' overloaded and overused). It can be any of (for now) unknown=0, file=1, directory=2 and we can add more types later if needed

handlers.go
```diff
@@ -4147,6 +4147,7 @@ type collectionListQueryRes struct {
 	Cid    util.DbCID
 	Size   int64
 	Path   *string
+	Dir    bool
```
whyrusleeping, line 4150: You'll also want to wire this into `handleCreateContent` and use it on the shuttle, `cmd/estuary-shuttle/main.go` line ~1184

replication.go
```diff
@@ -2813,8 +2816,20 @@ func (cm *ContentManager) addObjectsToDatabase(ctx context.Context, content uint
 		attribute.Int("numObjects", len(objects)),
 	)
 
+	// check if pinned CID is a file or a dir (defaults to file)
+	isDir := false
+	nilCID := cid.Cid{}
+	if root != nilCID && dserv != nil {
+		nd, err := dserv.Get(ctx, root)
```
whyrusleeping, line 2823: this first error needs checking

dbmgr/dbmgr.go
marshall: +1 to this, @gmelodie maybe similar to your proposal in this PR:
https://github.com/application-research/estuary/pull/100/files#diff-56307bdb7ded67ebe516f562b90384407b21b2d34e8b5766b42602e1fb9705f0R4161

handlers.go
```diff
@@ -4203,6 +4204,11 @@ func (s *Server) handleColfsListDir(c echo.Context, u *User) error {
 			return err
 		}
 
+		// trying to list a CID dir, not allowed
+		if strings.HasPrefix(relp, ".") && r.Dir == true {
+			return fmt.Errorf("listing CID directories is not allowed")
```
marshall, line 4209: just to verify .. this will fail the entire listing if a CID directory is a child of an existing collection? is there a middle ground between the CID dir being the same as `relp` and being a child of `relp` ?

marshall (changes requested):
Agreed on @whyrusleeping comments, and had a question of my own that I added to the mix.
 
great work on all of this!

handlers.go
gmelodie: The relative path `relp` is checked against `dir` from the request. `relp` will only be `.` if `dir` and `r.Path` are exactly the same (aka the request is `collections/fs/list?col=uuid&dir=/path/to/cid/dir`)

dbmgr/dbmgr.go
gmelodie: Agreed! I personally think `Type` is a good name since it's inside the `Content` struct all of the time, otherwise I used `contentType` when handling it outside of these structs. Here's what I did:
1. Added a `Type` field to `Content` and set three types (`Unknown`, `File` and `Directory`)
2. Changed the logic to use the new `Type` field and still keep the response of `handleColsFsDir` intact, returning a bool field `dir`

handlers.go
gmelodie: Donezo!

marshall (approved):
looks good! I left a suggestion for making the logic in `FindCIDType` a bit more flat, but that is totally optional :)

util/content.go
```diff
@@ -26,11 +37,36 @@ type ContentAddResponse struct {
...
+				contentType = File
+				if fsNode.IsDir() {
+					contentType = Directory
+				}
+			}
+		}
+	}
+
+	return contentType
+}
```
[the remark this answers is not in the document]
gmelodie: How do these `return`s work? I'm guessing I should be updating them with what I want to return? Or is there some magic happening I'm missing?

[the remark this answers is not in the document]
marshall: yeah, it's a go-ism that relies on the fact that the return value has been named/declared as `contentType` in the function signature. the short hand automatically returns the variables w/o explicit arguments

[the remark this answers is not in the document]
gmelodie: oh wow! I didn't know about that. Thought was more of a functional thing like when the returned value is the last generated variable. Cool!

util/unixfs.go
```diff
@@ -41,3 +44,21 @@ func ImportFile(dserv ipld.DAGService, fi io.Reader) (ipld.Node, error) {
...
+	case *merkledag.ProtoNode:
+		n, err := unixfs.FSNodeFromBytes(nd.Data())
+		if err != nil {
+			return nil, err
+		}
+		if n.Type() == unixfs.TSymlink {
+			return nil, fmt.Errorf("symlinks not supported")
+		}
+		return n, nil // success!
+	case *merkledag.RawNode:
```
whyrusleeping, line 59: We should treat RawNodes as files, I think this will error out and FindCIDType will return unknown.

whyrusleeping (approved):
One comment then the rest looks good!

handlers.go
```diff
@@ -3667,10 +3666,15 @@ func (s *Server) handleCreateContent(c echo.Context, u *User) error {
 		}
 	}
 
+	bserv := blockservice.New(s.Node.Blockstore, s.Node.Bitswap)
+	dserv := merkledag.NewDAGService(bserv)
+	contentType := util.FindCIDType(context.Background(), req.Root, dserv)
```
alvin-reyes, line 3671: Just wondering if we can wrap this into a global util function since the `Blockstore`?

`func util.FindCIDContentType(bs blockstore.Blockstore, rem exchange.Interface, root cid.Cid)  (contentType ContentType)`

cmd/estuary-shuttle/main.go
```diff
@@ -1184,11 +1184,16 @@ func (s *Shuttle) addrsForShuttle() []string {
 }
```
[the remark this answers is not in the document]
gmelodie: I def think that's something we need! But I think maybe in a different PR? I'm trying to add docs only to the functions I'm creating right now.

handlers.go
gmelodie: So the problem here is that sometimes we already have `dserv` but not necessarily `bs` or `bserv` for that matter. What I would love though (but not sure it'd be possible) is to have a magic `getDserv` function that simply finds the right blockstore and returns the correct `dserv`

cmd/estuary-shuttle/main.go
```diff
@@ -1184,11 +1184,16 @@ func (s *Shuttle) addrsForShuttle() []string {
...
 
 func (s *Shuttle) createContent(ctx context.Context, u *User, root cid.Cid, fname string, cic util.ContentInCollection) (uint, error) {
+	bserv := blockservice.New(s.Node.Blockstore, s.Node.Bitswap)
```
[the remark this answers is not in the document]
gmelodie: I think we do (this was a suggestion from @whyrusleeping IIRC) since this is ran on the shuttles, while `hadleCreateContent` runs on the main estuary node.

PR #111, opened 2022-02-26
chore: troubleshooting guide and move add/make sh to scripts directory

alvin-reyes:
# Chore / Changes

Just a simple troubleshooting section on the `readme.md`. We can add more to these section later on and separate it to a new page if we have more troubleshooting guides.

This also moves the `add-file.sh` and `make-deal.sh` to scripts directory. This is just to tidy up and can possibly be a precursor for creating `regression` and `chaos` testing later on.

I have added a few shell scripts to allow developers to experiment with the existing endpoints via curl and quickly test them as soon as they have estuary up and running.


lanzafame (approved):

whyrusleeping (approved):
Nice, having it set up as a regression test is nice, although I do use the 'add-file' script directly to add files to estuary. The replacement you wrote for it isnt usable for that purpose as is, maybe we can have a version that works this way somewhere else?

alvin-reyes:
> 

@whyrusleeping I added the original file as `add-file-to-collection.sh` on the latest commit.

PR #112, opened 2022-03-04
add opencensus handler to shuttle

lanzafame:
Also, add interface compile-time check for rcmgr.MetricsReporter.

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -62,7 +62,6 @@ import (
 	"github.com/labstack/echo/v4/middleware"
 	rcmgr "github.com/libp2p/go-libp2p-resource-manager"
 	routed "github.com/libp2p/go-libp2p/p2p/host/routed"
-	promhttp "github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/whyrusleeping/memo"
 )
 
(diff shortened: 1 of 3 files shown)


lanzafame:
The extra .gitignore entries are based of the what estuary/shuttle add to the directory when I started them with minimal flags.

alvin-reyes (approved):
LGTM

gmelodie:
> is the opencensus endpoint actually different than the prometheus one?

@whyrusleeping it doesn't seem to be different in this case. [Exporter()](https://github.com/lanzafame/estuary/blob/401b92946f194b4a00d7dc2d7b6972a5721cd7d7/metrics/exporter.go#L11-L29) basically creates a default [Exporter](https://github.com/census-ecosystem/opencensus-go-exporter-prometheus/blob/aaaf21250a48ed3f2da42bab1aab23ae7d48bcd6/prometheus.go#L36-L41) object with the `estuary` namespace. But in the end it acts as an API endpoint it seems to only return the `http.Handler` part of that object (not sure how Go is clever to make that work, maybe the casting in the function return?). It seems that `Exporter()` and `promhttp.Handler()` return almost identical objects the way we have `Exporter()` implemented now, that is. I think the way to go is to use `Exporter()` so that we can add options and configs if need be in the future.

cmd/estuary-shuttle/main.go
```diff
@@ -376,6 +376,7 @@ func main() {
 		})
 
 		go func() {
+			http.Handle("/debug/metrics/opencensus", estumetrics.Exporter())
 			http.Handle("/debug/metrics/prometheus", promhttp.Handler())
```
gmelodie, line 380: We may want to remove this endpoint in favor of the new one

gmelodie (commented):
Small remark, but looking good!

lanzafame:
@whyrusleeping @gmelodie personal opinion, we remove the opencensus and prometheus endpoints and just have `/debug/metrics`. Thoughts?

lanzafame:
@gmelodie Shall I make the same endpoint change to the main estuary handlers as well?

PR #117, opened 2022-03-07
Add autoretrieve init (create token) endpoint

gmelodie:
This will enable us to register autoretrieve servers just as we register shuttles. In fact, pretty much all the code is a copy of what we do on shuttles for the time being.


whyrusleeping (approved):
One optional nitpick, then LGTM

PR #121, opened 2022-03-08
add public endpoint for miner rankings

whyrusleeping:
cc @brendalee , this will expose the miner ranking data publicly


brendalee:
@davidd8 FYI!

PR #122, opened 2022-03-09
Autoretrieve heartbeat endpoint

gmelodie:
Added the autoretrieve heartbeat endpoint. Simple POST endpoint that updates a `LastConnection` field.

### DIFF
--- a/util/retrieval.go
+++ b/util/retrieval.go
@@ -1,6 +1,10 @@
 package util
 
-import "gorm.io/gorm"
+import (
+	"time"
+
+	"gorm.io/gorm"
+)
 
 type RetrievalFailureRecord struct {
 	gorm.Model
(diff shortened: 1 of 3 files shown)


util/retrieval.go
```diff
@@ -15,3 +19,25 @@ type RetrievalProgress struct {
 	Wait   chan struct{}
 	EndErr error
 }
+
+type HeartbeatAutoretrieveResponse struct {
+	Handle         string    `json:"handle"`
+	LastConnection time.Time `json:"lastConnection"`
+}
+
+type AutoretrieveListResponse struct {
```
alvin-reyes, line 28: the `AutoretrieveListResponse` and `InitAutoretrieveResponse` looks to have the same structure. Can we just merge them into `AutoRetrieveResponse` or is there a particular reason for it? e.i is `AutoretrieveListResponse` meant to be a [] list?

util/retrieval.go
```diff
@@ -15,3 +19,25 @@ type RetrievalProgress struct {
...
+	// AddrInfo       *peer.AddrInfo  `json:"addrInfo"`
+	// Address        address.Address `json:"address"`
+	// Hostname       string          `json:"hostname"`
+
+}
+
+type InitAutoretrieveResponse struct {
+	Handle         string `json:"handle"`
+	Token          string `json:"token"`
+	LastConnection time.Time
```
alvin-reyes, line 42: Do we need to add `json:"lastConnection"` on the `LastConnection` property?

whyrusleeping (approved):
aside from @alvin-reyes's comment, this LGTM

util/retrieval.go
gmelodie: I was following the same structure (read: copy pasting) of the shuttle register and listing api. In the future we might want these two responses to be different.

alvin-reyes (approved):
LGTM

PR #125, opened 2022-03-16
Autoretrieve index updater job

gmelodie:
This PR adds additional functionality to the `autoretrieve` endpoints. It takes address and peerID data on the `init` API call so it can tell the indexes about the autoretrieve server.

## Larger context
Basically there's this server called [autoretrieve](https://github.com/application-research/autoretrieve) that acts as a compatibility gateway from the last version of IPFS protocol (bitswap) to the newer one  (graphsync). As estuary doesn't support bitswap, we want people to be able to query autoretrieve with bitswap, and then autoretrieve will query estuary using graphsync. To do this, we need to update the indexer nodes (who hold information about where which CID is) to say "all CIDs that are in estuary can be found in the autoretrieve node".
Then the use case becomes: peer asks indexer where CID is > idexer tells autoretrieve has it > peer asks autoretrieve > autoretrieve gets it from estuary using graphsync > autoretrieve serves it to peer using bitswap.
This PR allows autoretrieve servers to tell estuary they're up using the heartbeat endpoint

## Testing
First we run the estuary node with an update interval of 1min (default is 720 or 1h)
```
INDEX_UPDATE_INTERVAL=1 ./estuary --datadir=mystorage --logging
```

Then we start `ipfs daemon` and list its ID information, we'll use its peerID and addresses for testing, but you can skip this step if you already have a valid peerID at hand
```
ipfs daemon
ipfs id
```

Now we can register a new autoretrieve server with our peerID
```
curl -X POST -H "Authorization: Bearer YOUR_API_KEY" "localhost:3004/admin/autoretrieve/init" -d "peerId=YOUR_PEER_ID&addresses=ONE_OR_MORE_ADDRESSES_SEPARATED_BY_COMMA" | jq
```

Finally we can send heartbeats using the `SECRET` we got from the `init` endpoint
```
curl -X POST -H "Authorization: Bearer YOUR_SECRET_KEY" "localhost:3004/autoretrieve/heartbeat" | jq
``` 
**Obs**: ignore the branch name here

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -3586,18 +3586,29 @@ func (s *Server) handleShuttleConnection(c echo.Context) error {
 }
 
 func (s *Server) handleAutoretrieveInit(c echo.Context) error {
+	// validate peerid and peer multi addresses
+	addresses := strings.Split(c.FormValue("addresses"), ",")
+	addrInfo, err := util.ValidatePeerInfo(c.FormValue("peerId"), addresses)
+	if err != nil {
+		return err
+	}
+
 	autoretrieve := &Autoretrieve{
 		Handle:         "AUTORETRIEVE" + uuid.New().String() + "HANDLE",
 		Token:          "SECRET" + uuid.New().String() + "SECRET",
 		LastConnection: time.Now(),
+		PeerID:         c.FormValue("peerId"),
+		Addresses:      c.FormValue("addresses"), // cant store []string in gorm
 	}
 	if err := s.DB.Create(autoretrieve).Error; err != nil {
 		return err
 	}
 
-	return c.JSON(200, &util.InitAutoretrieveResponse{
-		Handle: autoretrieve.Handle,
-		Token:  autoretrieve.Token,
+	return c.JSON(200, &util.AutoretrieveInitResponse{
+		Handle:         autoretrieve.Handle,
+		Token:          autoretrieve.Token,
+		LastConnection: autoretrieve.LastConnection,
+		AddrInfo:       addrInfo,
 	})
 }
 
@@ -3608,10 +3619,20 @@ func (s *Server) handleAutoretrieveList(c echo.Context) error {
 	}
 
 	var out []util.AutoretrieveListResponse
-	for _, a := range autoretrieves {
+
+	for _, ar := range autoretrieves {
+		// any of the multiaddresses of the peer should work to get addrInfo
+		// we get the first one
+		addresses := strings.Split(ar.Addresses, ",")
+		addrInfo, err := peer.AddrInfoFromString(addresses[0])
+		if err != nil {
+			return err
+		}
+
 		out = append(out, util.AutoretrieveListResponse{
-			Handle: a.Handle,
-			Token:  a.Token,
+			Handle:         ar.Handle,
+			LastConnection: ar.LastConnection,
+			AddrInfo:       addrInfo,
 		})
 	}
 
@@ -3634,9 +3655,18 @@ func (s *Server) handleAutoretrieveHeartbeat(c echo.Context) error {
 		return err
 	}
 
+	// any of the multiaddresses of the peer should work to get addrInfo
+	// we get the first one
+	addresses := strings.Split(autoretrieve.Addresses, ",")
+	addrInfo, err := peer.AddrInfoFromString(addresses[0])
+	if err != nil {
+		return err
+	}
+
 	out := util.HeartbeatAutoretrieveResponse{
 		Handle:         autoretrieve.Handle,
 		LastConnection: autoretrieve.LastConnection,
+		AddrInfo:       addrInfo,
 	}
 
 	return c.JSON(200, out)
--- a/main.go
+++ b/main.go
@@ -124,27 +124,30 @@ func (s *Server) updateAutoretrieveIndex(tickInterval time.Duration, quit chan s
 	var lastTickTime time.Time
 	ticker := time.NewTicker(tickInterval)
 
+	defer ticker.Stop()
 	for {
+		lastTickTime = time.Now().UTC().Add(-tickInterval)
+
+		// Find all autoretrieve servers that are online (that sent heartbeat)
+		err := s.DB.Find(&autoretrieves, "last_connection > ?", lastTickTime).Error
+		if err != nil {
+			log.Errorf("unable to query autoretrieve servers from database: %s", err)
+			return err
+		}
+		if len(autoretrieves) > 0 {
+			for _, ar := range autoretrieves {
+				fmt.Println("online: ", ar) // TODO: remove
+			}
+		} else {
+			log.Info("no autoretrieve servers online")
+		}
+
+		// wait for next tick, or quit
 		select {
 		case <-ticker.C:
-			lastTickTime = time.Now().UTC().Add(-tickInterval)
-
-			// Find all autoretrieve servers that are online (that sent heartbeat)
-			err := s.DB.Find(&autoretrieves, "last_connection > ?", lastTickTime).Error
-			if err != nil {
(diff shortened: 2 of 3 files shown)


jimmylee (approved):
LGTM, implementation looks clean, not sure who made the product ask though.

main.go
```diff
@@ -115,6 +116,39 @@ type ObjRef struct {
 	Offloaded uint
 }
 
+// updateAutoretrieveIndex ticks every tickInterval and checks for new information to add to autoretrieve
+// If so, it updates the filecoin index with the new CIDs, saying they are present on autoretrieve
+// With that, clients using bitswap can query autoretrieve servers using bitswap and get data from estuary
+func (s *Server) updateAutoretrieveIndex(tickInterval time.Duration, quit chan struct{}) error {
```
en0ma, line 122: can we change the name to something more related to the aim of the function? something along the line of `updateIndexer` - I believe this is the end goal.

main.go
```diff
@@ -115,6 +116,39 @@ type ObjRef struct {
 	Offloaded uint
 }
 
+// updateAutoretrieveIndex ticks every tickInterval and checks for new information to add to autoretrieve
+// If so, it updates the filecoin index with the new CIDs, saying they are present on autoretrieve
```
en0ma, line 120: `If so, it updates the filecoin indexer`?

en0ma:
> First we run the estuary node with an update interval of 1min (default is 720 or 1h)

@gmelodie looking at the code, the default update interval is 720min, not sure about the 1h bit from this statement.

jimmylee:
Thanks for reviewing @en0ma

main.go
```diff
@@ -115,6 +116,39 @@ type ObjRef struct {
...
+// With that, clients using bitswap can query autoretrieve servers using bitswap and get data from estuary
+func (s *Server) updateAutoretrieveIndex(tickInterval time.Duration, quit chan struct{}) error {
+	var autoretrieves []Autoretrieve
+	var lastTickTime time.Time
+	ticker := time.NewTicker(tickInterval)
+
+	for {
+		lastTickTime = time.Now().UTC().Add(-tickInterval)
```
en0ma, line 128: If I understand the intent, by the time a tick is read of the ticker channel, the `autoretrieves` wont exactly be peers whose `last_connection` is slightly above `- tickInterval`, it will be almost double (peers we last saw 720*2 minutes ago). Better to do this logic inside the `ticker.C` case

main.go
```diff
@@ -115,6 +116,39 @@ type ObjRef struct {
...
+
+		// Find all autoretrieve servers that are online (that sent heartbeat)
+		err := s.DB.Find(&autoretrieves, "last_connection > ?", lastTickTime).Error
+		if err != nil {
+			return err
```
en0ma, line 133: I am not sure we want to quit if no `autoretrieves` were found. Only quit for a non `ErrRecordNotFound` error

main.go
gmelodie: I'm assuming you want some clarification as to what that is? https://github.com/filecoin-project/storetheindex

main.go
gmelodie: I like having `autoretrieve` on the name so we know that's related to autoretrieve

gmelodie:
@en0ma 
> @gmelodie looking at the code, the default update interval is 720min, not sure about the 1h bit from this statement.

That's me not knowing how to count haha, it should read 12h.

> isn't the index update interval too wide?

TI think it's okay to use that since this can get quite heavy for us and for the indexers too. Plus it's not a super pressing thing we need to update (it's only used to provide backwards compatibility).


> how did you come to that default value?
    
This was suggested by @whyrusleeping


> how often is the heartbeat endpoint called?

That's for autoretrieve people to decide, but it's a good question. @elijaharita any ideas here?

> what happens when requests are made to the indexer for CIDs that are already pinned in Estuary but have not been indexed yet (not indexed in indexer to point to Autoretrieve for them, given the wide update interval)?

The indexer doesn't know about those files so it returns that it doesn't know about them

en0ma:
> @en0ma
> 
> > @gmelodie looking at the code, the default update interval is 720min, not sure about the 1h bit from this statement.
> 
> That's me not knowing how to count haha, it should read 12h.
> 
> > isn't the index update interval too wide?
> 
> TI think it's okay to use that since this can get quite heavy for us and for the indexers too. Plus it's not a super pressing thing we need to update (it's only used to provide backwards compatibility).
> 
> > how did you come to that default value?
> 
> This was suggested by @whyrusleeping
> 
> > how often is the heartbeat endpoint called?
> 
> That's for autoretrieve people to decide, but it's a good question. @elijaharita any ideas here?
> 
> > what happens when requests are made to the indexer for CIDs that are already pinned in Estuary but have not been indexed yet (not indexed in indexer to point to Autoretrieve for them, given the wide update interval)?
> 
> The indexer doesn't know about those files so it returns that it doesn't know about them

Thank you @gmelodie for these answers, it helps put things in perspective (especially this - The indexer doesn't know about those files so it returns that it doesn't know about them)

main.go
gmelodie: I don't think `Find` throws that error, from [here](https://gorm.io/docs/query.html):
> If you want to avoid the ErrRecordNotFound error, you could use Find like db.Limit(1).Find(&user), the Find method accepts both struct and slice data

handlers.go
```diff
@@ -3608,10 +3619,20 @@ func (s *Server) handleAutoretrieveList(c echo.Context) error {
...
+		// we get the first one
+		addresses := strings.Split(ar.Addresses, ",")
+		addrInfo, err := peer.AddrInfoFromString(addresses[0])
+		if err != nil {
+			return err
+		}
+
 		out = append(out, util.AutoretrieveListResponse{
-			Handle: a.Handle,
-			Token:  a.Token,
```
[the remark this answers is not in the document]
gmelodie: Huuuuh... Do we have to though? I think, it's pretty much standard API behavior. But if we did, where would this information be communicated?

main.go
whyrusleeping: Yeah, Find will return an empty list, `First` will error if none are found

whyrusleeping (approved):
This seems right to me, can always add more things later if we learn we need them

main.go
en0ma: isn't `storetheindex` a filecoin indexer? https://github.com/filecoin-project/storetheindex#running-the-indexer-service

main.go
en0ma: I think that doc your referenced is for v2 of gorm(`Find` no longer returns this error in v2, but does in v1), and we are using v1, no? or better still can you verify that scenario?

elijaharita:
> > how often is the heartbeat endpoint called?
> 
> That's for autoretrieve people to decide, but it's a good question. @elijaharita any ideas here?

Every 5 minutes or so seems fine to me?

handlers.go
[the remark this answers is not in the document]
en0ma: 😅 I am just thinking out loud. Basically, I am saying, for now, we keep returning it until we can communicate this. But again, maybe I am over thinking this - If removing it now, wont break any client flow, then lets do it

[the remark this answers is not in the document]
elijaharita: if i understand correctly, tokens are used for authentication, and should be privately distributed to autoretrieve operators. is this an admin-only endpoint?

en0ma (approved):
LGTM

handlers.go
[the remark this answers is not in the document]
gmelodie: > is this an admin-only endpoint

Even if it is, I don't think we should be keeping plaintext tokens to begin with. I'll just let this as it is.

PR #128, opened 2022-03-19
unify CONTRIBUTING.md and DEVELOPMENT.md

gmelodie:
I think this will be less confusing for newcomers


alvin-reyes (approved):
LGTM - was also thinking of separating the TROUBLESHOOTING.md or possibly use the Wiki for it.

PR #138, opened 2022-03-29
Change ContentCreateBody.Root field data type to string

en0ma:
Currently, on the `/content/create` endpoint, the `ContentCreateBody.Root` field has a type`(cid.Cid)` that echo cannot bind to (we can't send that type over the wire as json). It fails with the error;

Screenshot 2022-03-29 at 18 40 22

The `ContentCreateBody.Root` field type has to be changed to a `string`, this PR does just that (see below test image)

Screenshot 2022-03-29 at 18 44 17


gmelodie (approved):
LGTM

PR #141, opened 2022-03-30
Create config files for estuary and estuary-shuttle

softwareplumber:
Hoping for an initial review; also maybe need to discuss with @gmelodie how to implement some tests.

Objective is for new config functionality to introduce no changes to actual behavior. Defaults should remain the same, command line options remain the same. To achieve this we set up the configuration in three phases; the first loads command line options into a default configuration (hence maintaining defaults as defined by app.Flags). Then any options specified in the config file are loaded. Then any options actually set on the command line are reloaded into the config object, overriding whatever is set in the config file.

This implementation adds no new configuration options except the configure command and the --config flag which specifies the path to the config file. All existing command line options should have an equivalent in the config file.


gmelodie:
@softwareplumber why not just load the config options before and then overwrite anything with the CLI options instead of loading the CLI options twice?

cmd/estuary-shuttle/main.go
```diff
@@ -196,49 +339,31 @@ func main() {
...
-		}
+	app.Commands = []*cli.Command{
+		{
+			Name:  "configure",
+			Usage: "Saves a configuration file to the location specified by the config parameter",
+			Action: func(cctx *cli.Context) error {
+				cfg := newDefaultConfig(cctx)
+				configuration := cctx.String("config")
+				cfg.Load(configuration) // Assume error means no configuration file exists
+				overrideSetOptions(cctx, cfg)
```
gmelodie, line 350: Again, everything except `overrideSetOptions` should be outside of the `Action` logic

gmelodie (changes requested):
Looking good! Just a couple comments but this goes a super long way to help us make the code more usable and it also adds some cool new functionality!!

cmd/estuary-shuttle/main.go
```diff
@@ -75,6 +78,140 @@ func init() {
...
+	}
+
+	return cfg
+}
+
+func overrideSetOptions(cctx *cli.Context, cfg *config.Shuttle) error {
+	var err error
+
+	for _, flag := range cctx.FlagNames() {
+		if cctx.IsSet(flag) {
```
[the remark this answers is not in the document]
softwareplumber: Done

main.go
```diff
@@ -152,6 +153,131 @@ func (s *Server) updateAutoretrieveIndex(tickInterval time.Duration, quit chan s
...
+			Datastore:        filepath.Join(ddir, "estuary-leveldb"),
+			WalletDir:        filepath.Join(ddir, "estuary-wallet"),
+			WriteLogTruncate: cctx.Bool("write-log-truncate"),
+			NoLimiter:        true,
+			WriteLog:         config.MakeAbsoluteDefault(ddir, cctx.String("write-log"), ""),
+		},
+	}
+
+	return cfg
+}
```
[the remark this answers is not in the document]
softwareplumber: I've moved this to config. Caveats per slack.

main.go
```diff
@@ -268,7 +402,10 @@ func main() {
 			Name:  "setup",
 			Usage: "Creates an initial auth token under new user \"admin\"",
 			Action: func(cctx *cli.Context) error {
-				db, err := setupDatabase(cctx)
+				cfg := newDefaultConfig(cctx)
+				cfg.Load(cctx.String("config"))
+				overrideSetOptions(cctx, cfg)
```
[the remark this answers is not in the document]
softwareplumber: I've moved out the 'new' but the Load needs to stay there - because we need cctx,String("config") in case the user has overridden the config file location.

main.go
```diff
@@ -443,26 +566,26 @@ func main() {
...
-		cm.isDealMakingDisabled = cctx.Bool("disable-deal-making")
-		cm.contentAddingDisabled = cctx.Bool("disable-content-adding")
-		cm.localContentAddingDisabled = cctx.Bool("disable-local-content-adding")
+		cm.isDealMakingDisabled = cfg.Deals.Disable
+		cm.contentAddingDisabled = cfg.Content.GlobalDisable
+		cm.localContentAddingDisabled = cfg.Content.Disable
 
 		cm.tracer = otel.Tracer("replicator")
 
 		if cctx.Bool("enable-auto-retrive") {
```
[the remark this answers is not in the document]
softwareplumber: Yup

main.go
```diff
@@ -307,43 +444,29 @@ func main() {
...
 				return nil
 			},
+		}, {
+			Name:  "configure",
+			Usage: "Saves a configuration file to the location specified by the config parameter",
+			Action: func(cctx *cli.Context) error {
+				cfg := newDefaultConfig(cctx)
+				configuration := cctx.String("config")
+				cfg.Load(configuration) // Assume error means no configuration file exists
+				overrideSetOptions(cctx, cfg)
```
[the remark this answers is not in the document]
softwareplumber: yes, but need the config option

cmd/estuary-shuttle/main.go
```diff
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	_ "net/http/pprof"
 	"os"
+	"os/user"
```
en0ma, line 12: I think you should use `github.com/mitchellh/go-homedir`, it works without the use of cgo, so the library can be used in cross-compilation environments

config/config.go
```diff
@@ -34,38 +39,104 @@ type BitswapConfig struct {
...
 
-		return encode(f, cfg)
-	*/
-	return nil
+	return encode(cfg, f)
+}
+
+var ErrEmptyPath = errors.New("node not initialized, please run configure")
+
+func MakeAbsolute(root string, path string) (error, string) {
```
en0ma, line 101: please change the return type order `(string, error)`

cmd/estuary-shuttle/main.go
```diff
@@ -91,154 +158,156 @@ func main() {
...
-			bsdir = filepath.Join(ddir, "blocks")
-		} else if bsdir[0] != '/' && bsdir[0] != ':' {
-			bsdir = filepath.Join(ddir, bsdir)
-		}
-
-		wlog := cctx.String("write-log")
-		if wlog != "" && wlog[0] != '/' {
-			wlog = filepath.Join(ddir, wlog)
-		}
+	app.Commands = []*cli.Command{
```
en0ma, line 287: I think the approach below is more intuitive, and we can easily expand the `config` command to have more sub-commands in the future. Also, I think the code should own the config filename and users should only specify the path.

```
app.Commands = []*cli.Command{
	{
		Name: "config",
		Subcommands: []*cli.Command{
			{
				Name:  "init",
				Usage: "creates a configuration file in the location specified by the path parameter",
				Flags: []cli.Flag{
					&cli.StringFlag{
						Name:     "path",
						Usage:    "specify path for where the config file with be created",
						Required: true,
					},
				},
				Action: func(cctx *cli.Context) error {
					path := cctx.String("path")
					if err := cfg.LoadFromPath(path, true); err != nil {
						return err
					}
					return overrideConfigWithDefaults(app.Flags, cctx, cfg) // call cfg.Save(configuration) in overrideWithDefaults - given we will take this line out in the future
				},
			},
		},
	},
}
```

This way, we can have `./estuary-shuttle  config init --path=` to create the config file `estuary-shuttle.json` in `path`

cmd/estuary-shuttle/main.go
```diff
@@ -91,154 +158,156 @@ func main() {
 	logging.SetLogLevel("rcmgr", "debug")
 
 	app := cli.NewApp()
+	usr, _ := user.Current()
+
+	cfg := config.NewShuttle()
+
 	app.Flags = []cli.Flag{
```
en0ma, line 165: Can we abstract all of these flags into the shuttle config? so we can reduce the notice on `main.go` and also, it makes testing easy as we can test these components individually.

```
app.Flags = cfg.GetAppFlags() //  since cfg = config.NewShuttle(), it will return only shuttle app flags. We can do the same for Estuary

File: cmd/estuary-shuttle/main.go [hunk 11]
```diff
@@ -75,6 +76,72 @@ func init() {
 	}
 }
 
+func overrideSetOptions(flags []cli.Flag, cctx *cli.Context, cfg *config.Shuttle) error {
```
en0ma, line 79: I know this method is temp, but I think you should move it under the shuttle config (`cfg.overrideWithDefaults(flags []cli.Flag, cctx *cli.Context)`), to reduce noise on `main.go` and to allow for easy testing.

cmd/estuary-shuttle/main.go
```diff
@@ -75,6 +76,72 @@ func init() {
...
+		if cctx.IsSet(name) {
+			log.Debugf("flag %s is set to %s", name, cctx.String(name))
+		} else {
+			continue
+		}
+		switch name {
+		case "datadir":
+			cfg.SetDataDir(cctx.String("datadir"))
+		case "blockstore":
+			err, cfg.Node.Blockstore = config.MakeAbsolute(cfg.DataDir, cctx.String("blockstore"))
```
en0ma, line 93: handle the error - return?

cmd/estuary-shuttle/main.go
```diff
@@ -91,154 +158,156 @@ func main() {
...
-		listens := []string{
-			"/ip4/0.0.0.0/tcp/6745",
-			"/ip4/0.0.0.0/udp/6746/quic",
-		}
+	app.Action = func(cctx *cli.Context) error {
 
-		if cctx.Bool("libp2p-websockets") {
-			listens = append(listens, "/ip4/0.0.0.0/tcp/6747/ws")
-		}
+		cfg.Load(cctx.String("config")) // Ignore error for now; eventually error out if no configuration file
```
en0ma, line 302:
```
cfg, _ = cfg.LoadFromPath(app.Flags, cctx) // call overrideWithDefaults internally -- lets abstract as much as we can out of main.go

File: config/content.go [hunk 14]
```diff
@@ -0,0 +1,6 @@
+package config
+
+type ContentConfig struct {
+	Disable       bool `json:",omitempty"`
+	GlobalDisable bool `json:",omitempty"` // not valid for shuttle
```
en0ma, line 5: please let us use the opportunity to name these configs so they are intuitive. In this case, using the current names `localContentAddingDisabled` and `contentAddingDisabled` is more discriptive

config/deals.go
```diff
@@ -0,0 +1,7 @@
+package config
+
+type DealsConfig struct {
+	FailOnTransferFailure bool `json:",omitempty"`
```
en0ma, line 4: `QuitOnTransferFailure` ?

config/deals.go
```diff
@@ -0,0 +1,7 @@
+package config
+
+type DealsConfig struct {
```
en0ma, line 3: `Deal`, also, let's drop the `Config` suffix on all the config struct

config/config.go
```diff
@@ -1,16 +1,21 @@
...
 	"os"
+	"path/filepath"
+
+	"github.com/facebookgo/atomicfile"
 )
-*/
+
+var ErrNotInitialized = errors.New("node not initialized, please run configure")
 
 type Config struct {
```
en0ma, line 16: Rename this to `Node` and move the struct and its methods into its own file. Then create an empty `config` struct in this file - it will have all the helper methods, load, save, etc. It can then be embedded in `Shuttle` and `Estuary` config structs

config/estuary.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/application-research/estuary/build"
+)
+
+type Estuary struct {
+	Database       string
```
en0ma, line 11: `DatabaseConnString`

config/estuary.go
```diff
@@ -0,0 +1,93 @@
...
+	StagingData    string
```
en0ma, line 12: `StagingDataDir`

config/estuary.go
```diff
@@ -0,0 +1,93 @@
...
+	DataDir        string
+	ApiListen      string
+	AutoRetrieve   bool
+	LightstepToken string
+	Hostname       string
+	Node           Config
+	Jaeger         JaegerConfig
+	Deals          DealsConfig
```
en0ma, line 20: `Deal`

config/config.go
```diff
@@ -1,16 +1,21 @@
...
-	ListenAddrs   []string
-	AnnounceAddrs []string
+	ListenAddrs   []string `json:"Swarm"`
+	AnnounceAddrs []string `json:"Announce"`
 
 	Blockstore string
```
en0ma, line 20: `BlockstoreDir`

config/estuary.go
```diff
@@ -0,0 +1,93 @@
...
+
+		Logging: LoggingConfig{
+			ApiEndpointLogging: false,
+		},
+
+		Node: Config{
+			ListenAddrs: []string{
+				"/ip4/0.0.0.0/tcp/6744",
+			},
+			Blockstore:       filepath.Join(pwd, "estuary-blocks"),
```
en0ma, line 82: let's change the config from `Blockstore` to `BlockstoreDir`

config/jaeger.go
```diff
@@ -0,0 +1,7 @@
+package config
+
+type JaegerConfig struct {
+	JaegerTracing      bool
```
en0ma, line 4: `EnableTracing`

config/jaeger.go
```diff
@@ -0,0 +1,7 @@
...
+	JaegerProviderUrl  string
```
en0ma, line 5: `ProviderUrl`

cmd/estuary-shuttle/main.go
softwareplumber: I've made the change, but I doubt anyone is going to actually run this on MacOs. Is developer convenience worth adding yet another dependency which could be vulnerable to supply-chain attack? My personal view is that the whole DeFi ecosystem needs to be vastly more careful about whose code gets included.

config/config.go
softwareplumber: Oops. Sorry, I'm a bit of a go newb. I can see how this would irritate!!

cmd/estuary-shuttle/main.go
softwareplumber: I feel this verges on extra functionality. I was trying to get existing flags working in a consistent way in this PR. 

Agree that config is in the long term likely to have a bunch of subcommands. Before launching into that though I think we should maybe discuss how the various pieces of estuary might interact (for example, to streamline the process of registering a shuttle).

cmd/estuary-shuttle/main.go
softwareplumber: I think there is an argument for keeping cli stuff out of the config package. It could become a dependency monster if every way to mutate configuration lives in the config package. For testing purposes isn't it enough just to expose the overrideSetOptions function in main?

cmd/estuary-shuttle/main.go
softwareplumber: I do see your concerns -- my view is more that command line processing is a primary function of main, and actually it's almost everything else that should be moved out. On first reading of the code, my instinct was that app.Action is way to long, and there should be another file containing a simple runEstuary(config.Estuary) that does the heavy lifting.  From a testing perspective I think this is better as it lets test code stand up an estuary instance without getting involved in any command line issues. Let's bounce it around Slack for consensus? I'll do it your way if that's what everyone else thinks.

cmd/estuary-shuttle/main.go
```diff
@@ -75,6 +76,72 @@ func init() {
...
+		case "blockstore":
+			err, cfg.Node.Blockstore = config.MakeAbsolute(cfg.DataDir, cctx.String("blockstore"))
+		case "no-blockstore-cache":
+			cfg.Node.NoBlockstoreCache = cctx.Bool("no-blockstore-cache")
+		case "write-log-truncate":
+			cfg.Node.WriteLogTruncate = cctx.Bool("write-log-truncate")
+		case "write-log-flush":
+			cfg.Node.HardFlushWriteLog = cctx.Bool("write-log-flush")
+		case "write-log":
+			err, cfg.Node.WriteLog = config.MakeAbsolute(cfg.DataDir, cctx.String("write-log"))
```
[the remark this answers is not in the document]
softwareplumber: Fixing as we speak...

cmd/estuary-shuttle/main.go
softwareplumber: see earlier responses...

config/content.go
softwareplumber: No problem.

config/deals.go
softwareplumber: I have no earthly idea what this actually does, so no opinion. Quit would imply to me that the whole system shuts down if a transfer fails... is that the case here?

config/deals.go
```diff
@@ -0,0 +1,7 @@
...
+	FailOnTransferFailure bool `json:",omitempty"`
+	Disable               bool `json:",omitempty"`
+	NoStorageCron         bool `json:",omitempty"`
```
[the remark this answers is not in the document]
softwareplumber: Renamed it, but I've left this where it is for now.  ContentConfig is common to both the main estuary and shuttle configuration, whereas DealsConfig applies only to estuary.  I'm not sure therefore that moving it to ContentConfig is the right thing to do.

config/config.go
softwareplumber: Agree, I was just being lazy here.

config/deals.go
softwareplumber: For now I've just renamed to DealConfig

config/config.go
softwareplumber: For now, I have renamed the Config struct to NodeConfig (to keep things consistent), and moved it into a new file node.go

config/estuary.go
softwareplumber: Done

config/estuary.go
softwareplumber: Done

config/estuary.go
softwareplumber: Done

config/config.go
softwareplumber: Done

config/estuary.go
```diff
@@ -0,0 +1,93 @@
...
+			Libp2pKeyFile:    filepath.Join(pwd, "estuary-peer.key"),
+			Datastore:        filepath.Join(pwd, "estuary-leveldb"),
```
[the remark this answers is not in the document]
softwareplumber: Done

config/estuary.go
```diff
@@ -0,0 +1,93 @@
...
+			WalletDir:        filepath.Join(pwd, "estuary-wallet"),
+			WriteLog:         "",
```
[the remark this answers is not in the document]
softwareplumber: Done

config/jaeger.go
softwareplumber: Done

config/jaeger.go
```diff
@@ -0,0 +1,7 @@
...
+	JaegerSamplerRatio float64
```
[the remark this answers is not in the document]
softwareplumber: Done

config/jaeger.go
softwareplumber: Done

config/deals.go
en0ma: Well, I use the package name and read with the struct `Config.Deal`. Also, for the case of instance name, I support the suffix.

```
type Estuary struct  { 
  DealConfig Config.Deal 
}

File: config/deals.go (hunk 26 above)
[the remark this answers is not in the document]
Comment (en0ma) [reply config/deals.go]: Then we move it to Estuary? it has nothing to do with deal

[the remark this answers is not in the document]
Comment (softwareplumber) [reply config/deals.go]: OK, that sounds better

File: config/deals.go (hunk 16 above)
Comment (softwareplumber) [reply config/deals.go]: Umm... OK. Since Deal and Estuary are in the same package, my compiler complains about config.Deal. I'm probably missing something. Almost nobody will ever use the struct name (because most accesses of the struct members will be via the enclosing struct). I think it is wrong to choose the shorter name for the struct name, therefore. But, I really really need to get this merged so I'll go ahead and rename all the members to 'XConfig' and all the structs to 'X'.

Comment (en0ma) [reply config/deals.go]: No, that is an example of how it will be used outside of a config package

Comment (en0ma) [reply config/deals.go]:
```
type Estuary struct  { 
  DealConfig Config.Deal
}

the above assumes  Estuary is in another package, that is using the deal config - in this case the suffix works. 

```
type Estuary struct  { 
  Deal Deal 
}

the above assumes  Estuary and Deal are both in the config package, in this case, we do not need the suffix for both - when external packages uses them, the external package will have a instance name with the suffix like;

```
type External struct  { 
  EstuaryConfig Config.Estuary
  or 
  Config Config.Estuary
}

en0ma: if this is too much trouble, we can ignore it and get this merged. We can revisit it later.

cmd/estuary-shuttle/main.go
en0ma: @softwareplumber I still think the code should own the config file name, and users should specify only dir path.

config/deals.go
en0ma: Lastly, do you have a dump of a sample `json` file of these configs? I am curios how it looks like when marshalled to a file. Are you going to add json tags to the structs?

en0ma: ok

en0ma:
@softwareplumber can you resolve the conflict, so we can get this merged

PR #144, opened 2022-04-03
Bump to lotus 1.15.1-rc5

ribasushi:
Depends on https://github.com/application-research/filclient/pull/65 ( already pulls it into the go.mod )

Also resolves https://github.com/ipfs/go-path/issues/56 for @gmelodie 

NOTE: UNTESTED!!! ( because we don't have tests yet ;)


gmelodie (approved):
Looks good! Thanks for doing this!

ribasushi:
@gmelodie you are too quick
The `go.mod` you merged refers to an **unmerged** commit from `filclient`. See the `Depends on...` 1st line above.

go.mod
```diff
@@ -86,6 +86,298 @@ require (
 	gorm.io/gorm v1.21.15
 )
 
+require (
```
en0ma, line 89: @ribasushi is this second `require` not weird?

ribasushi: @en0ma no, it is a go1.17+ thing. Note how everything is `indirect`

PR #148, opened 2022-04-04
Change the content size limit to 31GiB512MiB, so it can fit within a sector size 32Gib.

en0ma:
Currently, we are seeing issues of deals getting stuck in `StorageDealWaitingForData`. This happens when content uploaded, has a size greater than a sector and so will fit.

This PR changes the maximum content size limit to `31GiB512MiB`, so it can fit within a sector size of `32Gib`. It also, makes sure to do some proper preliminary checks (check the maximum and minimum content size, check that the user account has deal-making turned on, check that content has not been offloaded, etc) before creating a deal

This will make sure any newly uploaded content's deal, will likely get sealed and not fail due to content size being so large to fit into a sector.

PS: This does not add a content size limit to the content upload endpoints (`content/add`, `content/add-car`, `content/add-ipfs` and `content/create`), it only makes sure a piece to be sent to Filecoin (if bigger than 31GiB512MiB, is split into chunks) can fit into a sector. 

Maybe in the future, we can add an upload size limit on the upload endpoints to reject content uploads over a limit and suggest that users should split their files before uploading.

PS#2: This will not correct contents that are already stuck in `StorageDealWaitingForData`.

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -54,7 +54,7 @@ var defaultReplication = 6
 // maxFilecoinContentSizeLimit = 31GiB512MiB
 const maxFilecoinContentSizeLimit = 33_822_867_456
 
-// minFilecoinContentSizeLimit = 0.24GiB (256 << 10)
+// minFilecoinContentSizeLimit = 256KiB (256 << 10)
 const minFilecoinContentSizeLimit = 262_144
 
 // Making default deal duration be three weeks less than the maximum to ensure
--- a/replication.go
+++ b/replication.go
@@ -51,7 +51,8 @@ import (
 
 var defaultReplication = 6
 
-const defaultContentSizeLimit = 34_000_000_000
+// defaultContentSizeLimit = 31GiB512MiB
+const defaultContentSizeLimit = 33_822_867_456
 
 // Making default deal duration be three weeks less than the maximum to ensure
 // miners who start their deals early dont run into issues
--- a/replication.go
+++ b/replication.go
@@ -51,8 +51,11 @@ import (
 
 var defaultReplication = 6
 
-// defaultContentSizeLimit = 31GiB512MiB
-const defaultContentSizeLimit = 33_822_867_456
+// maxFilecoinContentSizeLimit = 31GiB512MiB
+const maxFilecoinContentSizeLimit = 33_822_867_456
+
+// minFilecoinContentSizeLimit = 0.24GiB (256 << 10)
+const minFilecoinContentSizeLimit = 262_144
 
 // Making default deal duration be three weeks less than the maximum to ensure
 // miners who start their deals early dont run into issues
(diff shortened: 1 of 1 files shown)


replication.go
```diff
@@ -51,7 +51,11 @@ import (
 
 var defaultReplication = 6
 
-const defaultContentSizeLimit = 34_000_000_000
+// maxFilecoinContentSizeLimit = 31GiB512MiB
+const maxFilecoinContentSizeLimit = 33_822_867_456
+
+// minFilecoinContentSizeLimit = 0.24GiB (256 << 10)
```
Jorropo, line 57: That is 0.25MiB not GiB.
Also I think it's just clearer if it's `256KiB`.

en0ma: shot, thanks @Jorropo

whyrusleeping (approved):
LGTM!

replication.go
```diff
@@ -51,7 +51,11 @@ import (
 
 var defaultReplication = 6
 
-const defaultContentSizeLimit = 34_000_000_000
+// maxFilecoinContentSizeLimit = 31GiB512MiB
+const maxFilecoinContentSizeLimit = 33_822_867_456
```
ribasushi, line 55: >  // maxFilecoinContentSizeLimit = 31GiB512MiB 

@en0ma this is incorrect. The maximum byte-sized-payload you can get into a 32GiB filecoin sector is: `32GiB / 128 * 127` which comes out to `34,091,302,912` bytes. The value of `34,000,000,000` was chosen to roughly **over**-account for IPLD overhead within the serialized dag. Dropping it further is not warranted.

As a "living example" - this car file weighing at `34,091,006,682` bytes is replicated multiple times within filecoin: https://cargo.web3.storage/deal-cars/bafybeifaow2p3qrzndvr4dzvms7fun3e3bijuwkwrmzou4zupatyennufy_baga6ea4seaqi3u2amdhq6dba6kdaccbevfcbqskvqkutdfphfjsa7ntitll7ajy.car

Jorropo: @ribasushi I have LOTS some smaller `31.86GiB` cars that aren't uploading on estuary, or even some `31.73GiB` that are not working.

I belive the biggest issue is that the serialisation overhead depends with the number of blocks.

So a .car that stores a few big blocks would allow a limit really close to `32GiB` while one of lots of small block need far more space for the serialisation overhead.

Jorropo: I think the correct solution would be trying to do at least the first step of the CARCompPiece step, to get the final size of the piece, and then decide depending of how big it is (if the piece is above 32GiB or not).

ribasushi: @Jorropo I am strictly pointing out how the filecoin chain works. The 31.75 is the maximum value of the thing you run commp over. This is all after the IPLD encoding and spitting out as car. 31.75 - 34000000000 gives you 91MiB of overhead for IPLD ( which is way too much already ).

en0ma: Thanks for the details @Jorropo  @ribasushi. I believe I have a better understanding now. I agree that we should do the check of `34000000000` only after the `CARCompPiece` step, and not before.

ribasushi: @en0ma sorry, need to clarify here, details are important.

> we should do the check of 34000000000 only after the CARCompPiece step

No. The check of what you feed into `CARCompPiece` is for `34,091,302,912` bytes, as what you run `commp` over is literally what you put into the sector payload - there is no other variable wrapping overhead.

en0ma:
closing this, to be reworked based on learnings from @ribasushi and @Jorropo

PR #149, opened 2022-04-05
Add BasicLimiter and ConnectionMananger configuration options to estuary and estuary-shuttle

### LINKED ISSUE #131: Estuary opens an unlimited number of connections
Since d00e817f93b9657d3be640e078ddf8a545cd8135 there is no way to limit the number of connections opened by Estuary. (The connection manager high water mark does not seem to apply to newly opened connections).

It's relatively unusual these days for internet traffic not to pass through some kind of stateful packet filter these days. Opening a connection requires any stateful packet filter between estuary and the connection target to track state. State are tables are limited in size. ISPs do throttle the number of states even if they say they don't. (my use case!)  The consequence of exceeding the state limit imposed by infrastructure will be the sudden termination of otherwise perfectly good connections.

softwareplumber:
Closes #131 - The number of connections opened can be effectively limited using the estuary configuration file.

### DIFF
--- a/config/limits.go
+++ b/config/limits.go
@@ -1,0 +1,49 @@
+package config
+
+import rcmgr "github.com/libp2p/go-libp2p-resource-manager"
+
+type SystemLimit struct {
+	MinMemory      int64
+	MaxMemory      int64
+	MemoryFraction float64
+
+	StreamsInbound  int
+	StreamsOutbound int
+	Streams         int
+
+	ConnsInbound  int
+	ConnsOutbound int
+	Conns         int
+
+	FD int
+}
+
+func (sl *SystemLimit) apply(lim *rcmgr.BasicLimiter) {
+	lim.SystemLimits = lim.SystemLimits.WithFDLimit(sl.FD).WithConnLimit(sl.ConnsInbound, sl.ConnsOutbound, sl.Conns).WithStreamLimit(sl.StreamsInbound, sl.StreamsOutbound, sl.Streams).WithMemoryLimit(sl.MemoryFraction, sl.MinMemory, sl.MaxMemory)
+}
+
+type TransientLimit struct {
+	StreamsInbound  int
+	StreamsOutbound int
+	Streams         int
+
+	ConnsInbound  int
+	ConnsOutbound int
+	Conns         int
+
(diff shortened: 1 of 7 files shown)


en0ma (approved):
LGTM

jimmylee:
Pretty awesome, LGTM, and will help us test in different environments.

config/shuttle.go
```diff
@@ -108,8 +109,39 @@ func NewShuttle() *Shuttle {
...
+					ConnsOutbound: 256,
+					Conns:         512,
+
+					FD: 1024,
+				},
+			},
+		},
+		ConnectionManagerConfig: ConnectionManager{
+			LowWater:  2000,
+			HighWater: 3000,
```
gmelodie, line 143: I'm a bit unsure as to where all these values came from. It seems you got them to the previous code, but it'd be good to have @whyrusleeping help us put some useful comments here.

config/limits.go
```diff
@@ -0,0 +1,49 @@
...
+
+type Limits struct {
+	SystemLimitConfig    SystemLimit
+	TransientLimitConfig TransientLimit
+}
+
+func (limits *Limits) apply(lim *rcmgr.BasicLimiter) {
+	limits.SystemLimitConfig.apply(lim)
+	limits.TransientLimitConfig.apply(lim)
+}
```
gmelodie, line 49: I think we could just have the structs here and only one `Apply(lim *rcmgr.BasicLimiter, limits *Limits)` function. It took me quite a bit of time to understand the code flow of this snippet.

gmelodie (commented):
I like the functionality! I'm just unsure about the way it's organized. I honestly cannot tell if I'm nitpicking so I'll ask for Lodge's opinion to confirm.

jlogelin:
In general, I think this looks good, I would suggest adding a couple of unit tests to verify default and overloaded configurations.

softwareplumber:
@gmelodie sorry, forced context switch. Back now. Yes, I guess some unit tests make sense. I'll take a look at that tomorrow.

gmelodie (approved):
LGTM!! Thanks for adding the tests!!

PR #150, opened 2022-04-07
minor bug fixes

en0ma:
This PR makes sure;

1 - We add a newly created bucket to the bucket list, only if the content was added successfully, else when end up with a bucket whose size and content are incorrect.
2 - Properly check shuttle host


en0ma:
I will merged this to my other PR

PR #154, opened 2022-04-10
fix the shell to use the latest swaggo and yq released version

Icarus9913:
I noticed the scripts defines the swaggo and yq version with hard code. So I changed the shell to use the latest released version.


jimmylee (approved):
LGTM, thank you

alvin-reyes:
LGTM. Thank you!

PR #159, opened 2022-04-18
Autoretrieve/Indexer Integration

gmelodie:
This PR finishes adding the support for [autoretrieve](https://github.com/application-research/autoretrieve) on estuary. It allows Estuary to send advertisements to [indexers](https://github.com/filecoin-project/storetheindex) on behalf of autoretrieve servers, so that bitswap users looking for CIDs that are on Estuary will think autoretrieve servers are the ones with these CIDs. The clients then go to the autoretrieve servers and ask for the CIDs. Autoretrieve knows the CIDs belong to Estuary, queries and serves them back to the clients through bitswap.

The flow for this on Estuary's end is: 
1. Estuary receive an HTTP POST request on `/admin/autoretrieve/init` to register a new autoretrieve server (this can be done by running the following command on autoretrieve)
```bash
./autoretrieve register-estuary http://api.estuary.tech/ your-admin-api-key
```
2. Estuary receives regular heartbeats from the registered autoretrieve server on `/autoretrieve/heartbeat`
3. Estuary regularly announces to indexers saying every autoretrieve server that has submitted a heartbeat recently has its CIDs

Related: https://github.com/application-research/autoretrieve/issues/35

### Testing/Running
Checkout and build the branch
```
git checkout tell-indexer
make estuary
```
Run estuary (the `--indexer-url` is optional, it'll point to `https://cid.contact` by default, use it if you want to point to `https://dev.cid.contact` instead). `--indexer-tick-interval` is important because it sets how many minutes estuary waits until it advertises again (and the default is 12h so be sure to set it to a small value when testing). 
```
./estuary --logging --indexer-tick-interval=1 --indexer-url=https://dev.cid.contact
```
Finally, register and run an autoretrieve instance (this is important as it'll heartbeat to the estuary instance. If no one is heartbeating then estuary will not issue advertisements). The code to register is not yet merged to autoretrieve master, so we need to checkout the `feat/estuary-registration`.
```
# clone and build autoretrieve
cd ..
git clone git@github.com:application-research/autoretrieve.git
cd autoretrieve 
git checkout feat/estuary-registration
go build

# register autoretrieve on estuary
./autoretrieve register-estuary http://localhost:3004 <ESTUARY-API-KEY>

# run!
./autoretrieve
```

Every new CID uploaded to Estuary should now show up on `https://cid.contact` or `https://dev.cid.contact` (the whole process generally takes around 5mins after the CID is uploaded on estuary).

### Verification
Adding content:
image

Listing pins:
image

Content stats:
image


estuary.log
```diff
@@ -0,0 +1,48 @@
+2022/04/14 16:10:08 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
```
en0ma, line 1: @gmelodie this shouldn't be committed

storetheindex.log
```diff
@@ -0,0 +1 @@
+zsh: no such file or directory: ./storetheindex
```
en0ma, line 1: remove from commit

util/retrieval.go
```diff
@@ -57,11 +60,15 @@ func validateAddresses(addresses []string) []string {
 	return invalidAddresses
 }
 
-func ValidatePeerInfo(peerID string, addresses []string) (*peer.AddrInfo, error) {
-	// check if peerid format is correct
-	_, err := peer.IDFromString(peerID)
+func ValidatePeerInfo(privKeyStr string, addresses []string) (*peer.AddrInfo, error) {
```
en0ma, line 63: move this into the Autoretrieve package

util/retrieval.go
```diff
@@ -25,15 +26,17 @@ type RetrievalProgress struct {
...
-	LastConnection time.Time      `json:"lastConnection"`
-	AddrInfo       *peer.AddrInfo `json:"addrInfo"`
+	Handle            string         `json:"handle"`
+	LastConnection    time.Time      `json:"lastConnection"`
+	LastAdvertisement time.Time      `json:"lastAdvertisement"`
+	AddrInfo          *peer.AddrInfo `json:"addrInfo"`
 }
 
 type AutoretrieveListResponse struct {
-	Handle         string         `json:"handle"`
```
en0ma, line 34: move to  Autoretrieve package

main.go
```diff
@@ -117,30 +122,134 @@ type ObjRef struct {
...
+
+	// Create index-provider engine (s.Node.IndexProvider) to send announcements to
+	// this needs to keep running continuously because storetheindex
+	// will come to fetch for advertisements "when it feels like it"
+	s.Node.ArEngine, err = autoretrieve.NewAutoretrieveEngine()
+	if err != nil {
+		return err
+	}
 
+	s.Node.ArEngine.RegisterMultihashLister(func(ctx context.Context, contextID []byte) (provider.MultihashIterator, error) {
```
en0ma, line 143: move this registration into the Autoretrieve package - register it as part of NewAutoretrieveEngine internally, no need to expose it here

en0ma: so in `main.go` we can have something;

```
arMgr, err  := autoretrieve.NewAutoretrieveEngine() // this should register the listener and also call start - this way main can fail if the manger/engine cannot be created/started
if err != nil {
	return err
}
defer arMgr.Shutdown()
go arMgr.Run(context.Background()) // move everything in https://github.com/application-research/estuary/pull/159/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R182 here

Comment (en0ma) [reply main.go:143]: this main, main will be lite and autoretrieve APIs will be easy to see how it can be used

File: main.go [hunk 6]
```diff
@@ -117,30 +122,134 @@ type ObjRef struct {
...
+			case <-quit:
+				break
 			}
-		} else {
-			log.Info("no autoretrieve servers online")
+		}
+
+		log.Infof("announcing new CIDs to %d autoretrieve servers", len(autoretrieves))
+		// send announcement with new CIDs for each autoretrieve server
+		for _, ar := range autoretrieves {
```
en0ma, line 204: Also good to move this loop into a function, and call the function in a goroutine at each tick

main.go
```diff
@@ -117,30 +122,134 @@ type ObjRef struct {
...
+	s.Node.ArEngine.Start(context.Background())
+	defer s.Node.ArEngine.Shutdown()
+
+	// start ticker
+	ticker := time.NewTicker(tickInterval)
 	defer ticker.Stop()
-	for {
-		lastTickTime = time.Now().UTC().Add(-tickInterval)
 
+	for {
```
en0ma, line 182: I think this code can be more simplified;

```

for {
   select {
			case <-ticker.C:
				go notifyIndexer(time.Now().Add(-tickInterval)) // internally, use the time, pull the latest records and process
			case <-quit:
				break
	}
}

File: main.go [hunk 8]
```diff
@@ -117,30 +122,134 @@ type ObjRef struct {
...
 			}
-		} else {
-			log.Info("no autoretrieve servers online")
+		}
+
+		log.Infof("announcing new CIDs to %d autoretrieve servers", len(autoretrieves))
+		// send announcement with new CIDs for each autoretrieve server
+		for _, ar := range autoretrieves {
+
+			newContextID = []byte("AR_" + ar.Handle)
```
en0ma, line 206: any need for the AR_ prefixe? the hanlde already has `AUTORETRIEVE` prefix

main.go
```diff
@@ -117,30 +122,134 @@ type ObjRef struct {
...
+
+			log.Infof("sending announcement to %s", ar.Handle)
+			adCid, err := s.Node.ArEngine.NotifyPut(context.Background(), newContextID, providerID, retrievalAddresses, metadata.New(metadata.Bitswap{}))
+			if err != nil {
+				log.Errorf("could not announce new CIDs: %s", err)
+				continue
+			}
+
+			ar.LastAdvertisement = time.Now()
+			if err := s.DB.Save(&ar).Error; err != nil {
```
en0ma, line 248: avoid using `save` for updates, instead, use update column/columns

handlers.go
```diff
@@ -306,9 +307,9 @@ func (s *Server) ServeAPI(srv string, logging bool, lsteptok string, cachedir st
...
 	shuttle.GET("/list", s.handleShuttleList)
 
-	autoretrieve := admin.Group("/autoretrieve")
-	autoretrieve.POST("/init", s.handleAutoretrieveInit)
-	autoretrieve.GET("/list", s.handleAutoretrieveList)
+	ar := admin.Group("/autoretrieve")
+	ar.POST("/init", s.handleAutoretrieveInit)
+	ar.GET("/list", s.handleAutoretrieveList)
 
 	e.POST("/autoretrieve/heartbeat", s.handleAutoretrieveHeartbeat)
```
en0ma, line 314: you missed adding this to the `ar` route group

node/node.go
```diff
@@ -110,6 +110,8 @@ type Node struct {
 
 	Bwc *metrics.BandwidthCounter
 
+	ArEngine *autoretrieve.AutoretrieveEngine
```
en0ma, line 113: from my comments above, there will be no need to have this in the node

main.go
```diff
@@ -560,7 +671,7 @@ func main() {
 		}
 
 		stopUpdateIndex := make(chan struct{})
-		go s.updateAutoretrieveIndex(time.Duration(intervalMinutes)*time.Minute, stopUpdateIndex)
+		go s.updateAutoretrieveIndex(time.Duration(intervalMinutes)*time.Second, stopUpdateIndex)
```
en0ma, line 674: this could look like;

```

arMgr, err  := autoretrieve.NewAutoretrieveEngine() /
if err != nil {
	return err
}
defer arMgr.Shutdown()
go arMgr.Run(context.Background(), time.Duration(intervalMinutes)*time.Second, stopUpdateIndex)

File: main.go [hunk 13]
```diff
@@ -117,30 +122,134 @@ type ObjRef struct {
...
+				log.Errorf("no providerID for autoretrieve %s, skipping", ar.Handle)
+				continue
+			}
+			if len(retrievalAddresses) == 0 {
+				log.Errorf("no retrieval addresses for autoretrieve %s, skipping", ar.Handle)
+				continue
+			}
+
+			var newContents []Content
+			err = s.DB.Find(&newContents, "active = true and created_at >= ?", ar.LastAdvertisement).Error
```
en0ma, line 229: if we don't need the contents, why not count them instead?

main.go
```diff
@@ -117,30 +122,134 @@ type ObjRef struct {
...
+	// Create index-provider engine (s.Node.IndexProvider) to send announcements to
+	// this needs to keep running continuously because storetheindex
+	// will come to fetch for advertisements "when it feels like it"
+	s.Node.ArEngine, err = autoretrieve.NewAutoretrieveEngine()
+	if err != nil {
+		return err
+	}
 
+	s.Node.ArEngine.RegisterMultihashLister(func(ctx context.Context, contextID []byte) (provider.MultihashIterator, error) {
+		var newContents []Content
```
en0ma, line 144: nitpick: it is ok to move these variables closer to select queries, no need to have them here since there is not they can be out of scope - easier to read

autoretrieve/options.go
```diff
@@ -0,0 +1,273 @@
...
+// Note that this option only takes effect if the PublisherKind is set to HttpPublisher.
+// See: WithPublisherKind.
+func WithHttpPublisherListenAddr(addr string) Option {
+	return func(o *options) error {
+		o.pubHttpListenAddr = addr
+		return nil
+	}
+}
+
+// WithTopicName sets toe topic name on which pubsub announcements are published.
```
en0ma, line 184: * the

autoretrieve/autoretrieve.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return nil, err
+	}
+	pubT, err := pubG.Join(topic)
+	if err != nil {
+		return nil, err
+	}
+
+	newEngine, err := New(
+		WithTopic(pubT),      // TODO: remove, testing
```
en0ma, line 74: any need to have both, I thought `WithTopic` will override  `WithTopicName`

autoretrieve/engine.go
```diff
@@ -0,0 +1,524 @@
...
+	"github.com/ipld/go-ipld-prime"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+)
+
+const (
+	keyToCidMapPrefix      = "map/keyCid/"
+	cidToKeyMapPrefix      = "map/cidKey/"
+	keyToMetadataMapPrefix = "map/keyMD/"
+	latestAdvKey           = "sync/adv/"
+	linksCachePath         = "/cache/links"
```
en0ma, line 30: are these namespaces not too generic? should we add the package name along `autoretrieve//cache/links` ?

autoretrieve/engine.go
```diff
@@ -0,0 +1,524 @@
...
+		return err
+	}
+
+	// Initialize publisher with latest advertisement CID.
+	adCid, err := e.getLatestAdCid(ctx)
+	if err != nil {
+		return fmt.Errorf("could not get latest advertisement cid: %w", err)
+	}
+	if adCid != cid.Undef {
+		if err = e.publisher.SetRoot(ctx, adCid); err != nil {
```
en0ma, line 123: for `NoPublisher` wont call to `e.publisher.SetRoot` panic since  `e.publisher` will be nil?

autoretrieve/options.go
```diff
@@ -0,0 +1,273 @@
...
+func WithTopicName(t string) Option {
+	return func(o *options) error {
+		o.pubTopicName = t
+		return nil
+	}
+}
+
+// WithTopic sets the pubsub topic on which new advertisements are announced.
+// To use the default pubsub configuration with a specific topic name, use WithTopicName. If both
+// options are specified, WithTopic takes presence.
```
en0ma, line 198: >If both options are specified, WithTopic takes presence.
 

looking at `newPublisher` I see that the dt publisher always needs topic name even if pubtopic is provided. The quote above says something different, no?

autoretrieve/engine.go
```diff
@@ -0,0 +1,524 @@
...
+	switch e.pubKind {
+	case NoPublisher:
+		log.Info("Remote announcements is disabled; all advertisements will only be store locally.")
+		return nil, nil
+	case DataTransferPublisher:
+		dtOpts := []dtsync.Option{dtsync.Topic(e.pubTopic), dtsync.WithExtraData(e.pubExtraGossipData)}
+		if e.pubDT != nil {
+			return dtsync.NewPublisherFromExisting(e.pubDT, e.h, e.pubTopicName, e.lsys, dtOpts...)
+		}
+		ds := dsn.Wrap(e.ds, datastore.NewKey("/legs/dtsync/pub"))
```
en0ma, line 141: why not make `/legs/dtsync/pub` as const like the others

autoretrieve/engine.go
```diff
@@ -0,0 +1,524 @@
...
+	}
+
+	// Sign the advertisement.
+	if err := adv.Sign(e.key); err != nil {
+		return cid.Undef, err
+	}
+	return e.Publish(ctx, adv)
+}
+
+func (e *AutoretrieveEngine) putKeyCidMap(ctx context.Context, contextID []byte, c cid.Cid) error {
```
en0ma, line 453: maybe move all the store CRUD methods to a `datastore.go` file?

autoretrieve/engine.go
```diff
@@ -0,0 +1,524 @@
...
+		return nil, err
+	}
+
+	e := &AutoretrieveEngine{
+		options: opts,
+	}
+
+	e.lsys = e.mkLinkSystem()
+
+	return e, nil
```
en0ma, line 85: given `mhlister` is required, you should validate it is set before creating an engine

autoretrieve/engine.go
```diff
@@ -0,0 +1,524 @@
...
+	}
+	if err := e.entriesChunker.Close(); err != nil {
+		errs = multierror.Append(errs, fmt.Errorf("error closing link entriesChunker: %s", err))
+	}
+	return errs
+}
+
+// GetAdv gets the advertisement associated to the given cid c.
+// The context is not used.
+func (e *AutoretrieveEngine) GetAdv(_ context.Context, adCid cid.Cid) (*schema.Advertisement, error) {
```
en0ma, line 287: if context.Context is not used, maybe remove it

autoretrieve/engine.go
```diff
@@ -0,0 +1,524 @@
...
+
+// RegisterMultihashLister registers a provider.MultihashLister that is used to look up the
+// list of multihashes associated to a context ID. At least one such registration
+// must be registered before calls to AutoretrieveEngine.NotifyPut and AutoretrieveEngine.NotifyRemove.
+//
+// Note that successive calls to this function will replace the previous registration.
+// Only a single registration is supported.
+//
+// See: provider.Interface
+func (e *AutoretrieveEngine) RegisterMultihashLister(mhl provider.MultihashLister) {
```
en0ma, line 239: I don't think we need mutex. when initializing the engine, let the caller pass the `mhlister` handler func and set it once - it gives the added advantage of validating it exists before boot up (since the engine wont work without it) and one less complexity with mutex

autoretrieve/engine.go
```diff
@@ -0,0 +1,524 @@
...
+	err := e.ds.Put(ctx, datastore.NewKey(keyToCidMapPrefix+string(contextID)), c.Bytes())
+	if err != nil {
+		return err
+	}
+	// And the other way around when graphsync ios making a request,
+	// so the lister in the linksystem knows to what contextID we are referring.
+	return e.ds.Put(ctx, datastore.NewKey(cidToKeyMapPrefix+c.String()), contextID)
+}
+
+func (e *AutoretrieveEngine) getKeyCidMap(ctx context.Context, contextID []byte) (cid.Cid, error) {
```
en0ma, line 465: can we change the `key` in these functions to `contextID` - `key` is too generic also, no need for the `map` suffix. - `getContextIDCID` , `putContextIDCID` etc

autoretrieve/engine.go
```diff
@@ -0,0 +1,524 @@
...
+			}
+
+			if md.Equal(prevMetadata) {
+				// Metadata is the same; no change, no need for new advertisement.
+				return cid.Undef, provider.ErrAlreadyAdvertised
+			}
+
+			// Linked list is the same, but metadata is different, so generate
+			// new advertisement with same linked list, but new metadata.
+			cidsLnk = cidlink.Link{Cid: c}
```
en0ma, line 381: how is the metadata used?

autoretrieve/engine.go
```diff
@@ -0,0 +1,524 @@
...
+		if err != nil {
+			log.Errorw("Failed to announce advertisement on pubsub channel ", "err", err)
+			return cid.Undef, err
+		}
+	}
+	return c, nil
+}
+
+// PublishLatest re-publishes the latest existing advertisement to pubsub.
+func (e *AutoretrieveEngine) PublishLatest(ctx context.Context) error {
```
en0ma, line 210: how is this used? I can't see anywhere this is used

autoretrieve/engine.go
```diff
@@ -0,0 +1,524 @@
...
+	return e.publishAdvForIndex(ctx, contextID, providerID, addresses, md, false)
+}
+
+// NotifyRemove publishes an advertisement that signals the list of multihashes associated to the given
+// contextID is no longer available by this provider.
+//
+// Note that prior to calling this function a provider.MultihashLister must be registered.
+//
+// See: AutoretrieveEngine.RegisterMultihashLister, AutoretrieveEngine.Publish.
+func (e *AutoretrieveEngine) NotifyRemove(ctx context.Context, contextID []byte) (cid.Cid, error) {
```
en0ma, line 266: I want to assume we will use this for when contents are offloaded, right?

autoretrieve/linksystem.go
```diff
@@ -0,0 +1,181 @@
...
+			return nil, errNoEntries
+		}
+
+		ctx := lctx.Ctx
+		c := lnk.(cidlink.Link).Cid
+		log.Debugf("Triggered ReadOpener from engine's linksystem with cid (%s)", c)
+
+		// Get the node from main datastore. If it is in the
+		// main datastore it means it is an advertisement.
+		val, err := e.ds.Get(ctx, datastore.NewKey(c.String()))
```
en0ma, line 34: are you not supposed to be using the `dsLatestAdvKey` or calling `e.getLatestAdCid` here?

en0ma:
@gmelodie I added some comments and recommendations. Also, I can't fully see how we will handle contents in the shuttles. Since the autoretrieve servers will come asking for these contents, won't they come to the primary node and not to the shuttle nodes?

main.go
gmelodie: I thought of doing something along those lines but on `NewAutoretrieveEngine` when registering the lister we need to be able to access `s` to do `s.DB.Find`, and I'm not sure how to do this on this model

handlers.go
gmelodie: `ar` is an alias to `/admin/autoretrieve`, which is different than `/autoretrieve/heartbeat`

node/node.go
gmelodie: Same as above, we need to access the `s` (server) to query the database

autoretrieve/engine.go
gmelodie: This is not my code, I stole it from https://github.com/filecoin-project/index-provider/blob/dfc4fa4ad02a4f0494b67f7f6c15581c434ef89b/engine/engine.go#L25-L31.

autoretrieve/engine.go
gmelodie: Same as above, see https://github.com/filecoin-project/index-provider/blob/dfc4fa4ad02a4f0494b67f7f6c15581c434ef89b/engine/engine.go. Might be worth checking.

autoretrieve/options.go
gmelodie: Same as above, https://github.com/filecoin-project/index-provider/

autoretrieve/engine.go
gmelodie: Same as above, https://github.com/filecoin-project/index-provider/

autoretrieve/engine.go
gmelodie: Same as above, https://github.com/filecoin-project/index-provider/

autoretrieve/engine.go
gmelodie: `mhlister` is not required. You can theoretically use the engine without it.

autoretrieve/autoretrieve.go
```diff
@@ -0,0 +1,128 @@
...
+		m.offset++
+		return hash, nil
+	}
+	return nil, io.EOF
+}
+
+// newIndexProvider creates a new index-provider engine to send announcements to storetheindex
+// this needs to keep running continuously because storetheindex
+// will come to fetch advertisements "when it feels like it"
+func NewAutoretrieveEngine() (*AutoretrieveEngine, error) {
```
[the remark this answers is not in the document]
gmelodie: Same problem as above, we need to be able to access `s` on the lister function.

autoretrieve/engine.go
gmelodie: 100% agree, but think it's better suited for https://github.com/filecoin-project/index-provider/

autoretrieve/engine.go
gmelodie: I used it during testing but it turns out it wasn't the correct approach. I'll delete this

autoretrieve/engine.go
gmelodie: Maybe not even then. This is a question I had for @whyrusleeping: even when content is offloaded, we're still able to retrieve it for users right? So theoretically we would never need to remove anything from the advertisements.

gmelodie:
> Also, I can't fully see how we will handle contents in the shuttles

I don't have the internals figured out, but I think the main node is able to transparently get content from the shuttles it has registered, right?

main.go
en0ma: then pass the `s.DB` to `NewAutoretrieveEngine` and set a DB field in the engine

autoretrieve/engine.go
en0ma: @gmelodie I just looked, I can see they have fixed my concern here https://github.com/filecoin-project/index-provider/commit/563785de47bf47c8ae5ffc407d6fd7294e219a03, so I think you have to check, else it will panic

autoretrieve/options.go
en0ma: That repo is a reference, the code says otherwise, we can address our implementation of the reference repo, and even contribute a fix to it, based on our implementation

autoretrieve/engine.go
en0ma: again, it is a reference repo, we can improve things in our implementation

autoretrieve/engine.go
en0ma: it is required, you cant make the `mkLinkSystem` without it, it will fail. The reference repo is too relaxed, we can improve our implementation, to make sure the engine won't start if `mhlister` is not set - so we don't fail later after main.go bootstraps

autoretrieve/engine.go
en0ma: Given that interface was created in the reference repo, as the engine is a different package from the provider. This implementation is not the case, we should be fine without it

autoretrieve/engine.go
en0ma: no need for this function, hence no need for locking and unlocking (blocking)

autoretrieve/engine.go
en0ma: Again, I think our implementation can be different from the reference repo

main.go
gmelodie: huuuuh yeah that makes sense

autoretrieve/autoretrieve.go
```diff
@@ -0,0 +1,322 @@
...
+		m.offset++
+		return hash, nil
+	}
+	return nil, io.EOF
+}
+
+// newIndexProvider creates a new index-provider engine to send announcements to storetheindex
+// this needs to keep running continuously because storetheindex
+// will come to fetch advertisements "when it feels like it"
+func NewAutoretrieveEngine(stopCh chan struct{}, tickInterval time.Duration, db *gorm.DB) (*AutoretrieveEngine, error) {
```
elijaharita, line 71: instead of storing `stopCh` here, the `.Run()` function should probably just take a context like `Run(ctx context.Context)` and listen for `<-ctx.Done()`. a lot nicer if you're also working with other functions that take a context, because they can all be canceled in a consistent way

autoretrieve/autoretrieve.go
```diff
@@ -0,0 +1,322 @@
...
+	host, err := libp2p.New()
```
whyrusleeping, line 72: this shouldnt be spinning up its own libp2p instance, it should just be using the host of the estuary node thats running it.

gmelodie: Nice catch! It's leftover from when I was trying to spoof the autoretrieve server on advertisements. I'll remove this since this is not required.

autoretrieve/autoretrieve.go
```diff
@@ -0,0 +1,294 @@
...
+			case <-arEng.context.Done():
+				break
+			}
+		}
+
+		log.Infof("announcing new CIDs to %d autoretrieve servers", len(autoretrieves))
+		// send announcement with new CIDs for each autoretrieve server
+		for _, ar := range autoretrieves {
+
+			newContextID = []byte(ar.Handle)
```
whyrusleeping, line 157: lets put the contents of this loop into a separate function

estuary.log
whyrusleeping: this file is still here, need to delete it and rebase it out of the history so we dont end up with it in the repo

storetheindex.log
whyrusleeping: still hasnt been deleted

alvin-reyes (commented):
Approving under the assumption that we have thoroughly tested this on staging environment and the results are acceptable.

Please attached any verification test results here (if any). Test should include:
- Results of AR (database entries, records, new endpoint).
- Potential negative impact (Test any core functionality of estuary like /content/*, /collections, /public, )

docs/swagger.yaml
```diff
@@ -110,6 +110,37 @@ info:
   title: Estuary API
   version: 0.0.0
 paths:
+  /admin/autoretrieve/init:
```
alvin-reyes, line 113: Please add the proper annotations instead. This file will be automatically generated by our swagger workflow.

autoretrieve/autoretrieve.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+	}
+	return invalidAddresses
+}
+
+func ValidatePeerInfo(pubKeyStr string, addresses []string) (*peer.AddrInfo, error) {
+	// check if peerid is correct
+	pubKey, err := stringToPubKey(pubKeyStr)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode public key: %s", err)
```
alvin-reyes, line 261: replace with `log.Errorf("unable to decode public key %s: %w", pubKeyStr, err)`

docs/swagger.yaml
gmelodie: I _think_ this was automatically generated when I did `make swagger`. Was I not supposed to do that?

alvin-reyes (approved):
LGTM. 🏅

gmelodie:
Phew!

Issue #162, opened 2022-04-25
add-car via shuttle-proxy fails


olizilla:
POSTing to add-car on shuttle-4 works

```sh
curl -sS -X POST https://shuttle-4.estuary.tech/content/add-car -H "Authorization: Bearer ***" -H "Accept: application/json" -T lovely.car
{"cid":"bafybeif34a7jxp5b6od7fejcfuhjrop2qeexacp76muwukivmnh6nfqlk4","estuaryId":28340318,"providers":["/ip4/3.134.223.177/tcp/6745/p2p/12D3KooWN8vAoGd6eurUSidcpLYguQiGZwt4eVgDvbgaS7kiGTup"]}
```

but the same call to upload.estuary.tech fails

```sh
curl -sS -X POST https://upload.estuary.tech/content/add-car -H "Authorization: Bearer ***" -H "Accept: application/json" -T lovely.car
{"error":"request Content-Type isn't multipart/form-data"}
```

PR #165, opened 2022-04-26
Properly set shuttle-proxy content url paths

en0ma:
Right now, shuttle proxy content URLs paths for both `/content/add-car` and `/content/add` will always send both request to `content/add` on the shuttles and not the actual `/content/add-car` for the case of car file upload. This PR makes sure the file uploads are properly proxied to the right paths.

This should fix https://github.com/application-research/estuary/issues/162

### DIFF
--- a/cmd/shuttle-proxy/main.go
+++ b/cmd/shuttle-proxy/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/labstack/echo/v4"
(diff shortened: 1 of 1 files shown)


cmd/shuttle-proxy/main.go
```diff
@@ -104,19 +105,32 @@ func (p *Proxy) getEndpoints(c echo.Context) ([]string, error) {
 		return nil, c.String(code, err.Error())
 	}
 
-	if len(view.Settings.UploadEndpoints) == 0 {
-		return nil, fmt.Errorf("all upload endpoints are unavailable")
+	var endps []string
+	for _, endp := range view.Settings.UploadEndpoints {
```
en0ma, line 109: the values of `UploadEndpoints` has `/content/add` as a path, so we have to remove them

whyrusleeping (approved):
Ah, great catch, thank you!

PR #167, opened 2022-04-26
properly cleanup temp blockstore after file upload

en0ma:


handlers.go
```diff
@@ -568,6 +568,14 @@ func (s *Server) handleAddCar(c echo.Context, u *User) error {
 		return err
 	}
 
+	defer func() {
+		go func() {
```
gmelodie, line 572: Why are you using `defer` *and* goroutines here? If I understand correctly, `defer` will execute the code at the end of the function or if the function panics. Either way, I don't see why a goroutine here would be necessary.

handlers.go
```diff
@@ -769,6 +771,14 @@ func (s *Server) handleAdd(c echo.Context, u *User) error {
...
```
gmelodie, line 775: Same questions as above

handlers.go
```diff
@@ -769,6 +771,14 @@ func (s *Server) handleAdd(c echo.Context, u *User) error {
...
+			if err := s.StagingMgr.CleanUp(bsid); err != nil {
+				log.Errorf("failed to clean up staging blockstore: %s", err)
```
gmelodie, line 777: Same Nit:
```suggestion
                        err := s.StagingMgr.CleanUp(bsid)
			if err != nil {
```

handlers.go
en0ma: The call to `s.StagingMgr.CleanUp(bsid)` will take a while to return and we need not block the request because of it. That is why it is called in a goroutine

handlers.go
en0ma: see https://github.com/application-research/estuary/pull/167#discussion_r859897687 above

handlers.go
```diff
@@ -568,6 +568,14 @@ func (s *Server) handleAddCar(c echo.Context, u *User) error {
 		return err
 	}
 
+	defer func() {
+		go func() {
+			if err := s.StagingMgr.CleanUp(bsid); err != nil {
```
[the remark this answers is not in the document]
en0ma: Actually, `if err := func(); err != nil {` is idiomatic way if you don't need the variable outside the block scope, so it is fine

handlers.go
en0ma: see https://github.com/application-research/estuary/pull/167#discussion_r859900169 above

Issue #21, opened 2021-08-16
Preventing making a deal with providers if DataCap is low


jimmylee:
`unrecognized response state 11: deal rejected: verified deal DataCap too small for proposed piece size`

We should automatically stop making deals when DataCap is too low.

PR #170, opened 2022-04-26
Preventing making a deal with providers if DataCap is low

en0ma:
This PR will make sure we don't make deals and endup with error like ` deal rejected: verified deal DataCap too small for proposed piece size`. It closes  https://github.com/application-research/estuary/issues/21

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -42,6 +42,7 @@ import (
 	"github.com/labstack/echo/v4"
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/pkg/errors"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/xerrors"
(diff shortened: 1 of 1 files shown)


gmelodie (commented):
Code looks good! Can you add some pointers as to how to test this locally?

en0ma:
> Code looks good! Can you add some pointers as to how to test this locally?

I did a quick trick to test this, Go to `https://verify.glif.io/` and check your local node wallet address datacap allocation. Go to your contents table and look for a content that has no deal, modify the size just large enough than your datacap allocation

jimmylee (approved):
LGTM, would like @whyrusleeping to take a quick look

replication.go
```diff
@@ -1395,6 +1396,17 @@ func (cm *ContentManager) ensureStorage(ctx context.Context, content Content, do
...
 			return nil
 		}
+
+		bl, err := cm.FilClient.Balance(ctx)
+		if err != nil {
+			return errors.Wrap(err, "could not retrieve dataCap from client balance")
+		}
+
+		if bl.VerifiedClientBalance.LessThan(big.NewIntUnsigned(uint64(abi.UnpaddedPieceSize(content.Size).Padded()))) {
+			// how do we notify admin to top up datacap?
```
whyrusleeping, line 1406: A loud error log message *should* make it through. Ideally we get the error log rate on grafana down low enough that we can alert on it. 

aside from that, I think @lanzafame can help us set up some alerts

whyrusleeping (approved):
LGTM! Thanks @en0ma !

Issue #137, opened 2022-03-25
Get CID for all the contents in a Collection


gmelodie:
Want to add a `commit` endpoint that returns a CID for all its contents in that point in time, much like a version control system.

A collection should now have: `uuid, humanReadableLabel, currentCID, [prevCIDs]`

en0ma:
closing this as it has been addressed

PR #171, opened 2022-04-27
Collection CIDs (commit endpoint)

gmelodie:
This PR adds the `/collections/commit/:coluuid` endpoint, to which you can `POST` to get a CID that references all the files in that collection in that point in time. The CID gets pinned to the Estuary node and can be used to reference those files in that point in time. The functioning is very similar to an SVG (aka Git-like programs), thus the `commit` name of the endpoint.

Testing: `curl -X POST -H "Authorization: Bearer $YOUR-APIKEY" "localhost:3004/collections/commit/YOUR-COLLECTION-UUID" | jq`

Implements #137

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -3163,7 +3163,7 @@ func (s *Server) handleAddContentsToCollection(c echo.Context, u *User) error {
 // @Tags         collections
 // @Produce      json
 // @Success      200  {object}  string
-// @Router       /collections/{collection-id}/commit [post]
+// @Router       /collections/{coluuid}/commit [post]
 func (s *Server) handleCommitCollection(c echo.Context, u *User) error {
 	colid := c.Param("coluuid")
 
@@ -3176,14 +3176,28 @@ func (s *Server) handleCommitCollection(c echo.Context, u *User) error {
 	if err := s.DB.Model(CollectionRef{}).
 		Where("collection = ?", col.ID).
 		Joins("left join contents on contents.id = collection_refs.content").
-		Select("contents.*").
+		Select("contents.*, collection_refs.path").
 		Scan(&contents).Error; err != nil {
 		return err
 	}
 
+	bserv := blockservice.New(s.Node.Blockstore, nil)
+	dserv := merkledag.NewDAGService(bserv)
+
+	// create DAG respecting directory structure
 	collectionNode := unixfs.EmptyDirNode()
+	dserv.Add(context.Background(), collectionNode)
 	for _, c := range contents {
-		collectionNode.AddRawLink(c.Name, &ipld.Link{
+		dirs, err := util.DirsFromPath(c.Path, c.Name)
+		if err != nil {
+			return err
+		}
+
+		lastDirNode, err := util.EnsurePathIsLinked(dirs, collectionNode, dserv)
+		if err != nil {
+			return err
+		}
+		lastDirNode.AddRawLink(c.Name, &ipld.Link{
 			Size: uint64(c.Size),
 			Cid:  c.Cid.CID,
 		})
--- a/handlers.go
+++ b/handlers.go
@@ -43,6 +43,7 @@ import (
 	ipld "github.com/ipfs/go-ipld-format"
 	logging "github.com/ipfs/go-log/v2"
 	merkledag "github.com/ipfs/go-merkledag"
+	"github.com/ipfs/go-unixfs"
 	uio "github.com/ipfs/go-unixfs/io"
 	car "github.com/ipld/go-car"
 	"github.com/labstack/echo/v4"
@@ -215,6 +216,8 @@ func (s *Server) ServeAPI(srv string, logging bool, lsteptok string, cachedir st
 	cols.POST("/create", withUser(s.handleCreateCollection))
 	cols.POST("/add-content", withUser(s.handleAddContentsToCollection))
 	cols.GET("/content/:coluuid", withUser(s.handleGetCollectionContents))
+	cols.POST("/:coluuid/commit", withUser(s.handleCommitCollection))
+
 	colfs := cols.Group("/fs")
 	colfs.GET("/list", withUser(s.handleColfsListDir))
 	colfs.POST("/add", withUser(s.handleColfsAdd))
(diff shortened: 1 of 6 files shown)


handlers.go
```diff
@@ -215,6 +216,7 @@ func (s *Server) ServeAPI(srv string, logging bool, lsteptok string, cachedir st
 	cols.POST("/create", withUser(s.handleCreateCollection))
 	cols.POST("/add-content", withUser(s.handleAddContentsToCollection))
 	cols.GET("/content/:coluuid", withUser(s.handleGetCollectionContents))
+	cols.POST("/commit/:coluuid", withUser(s.handleCommitCollection))
```
en0ma, line 219: Can you change this to `/:coluuid/commit` to be more restful? One thing I hope we do in the future is standardizing the endpoints to be restful and consistent. it will be nice going forward, that new endpoints are restful.

handlers.go
```diff
@@ -3153,6 +3155,66 @@ func (s *Server) handleAddContentsToCollection(c echo.Context, u *User) error {
...
+		lnk, err := format.MakeLink(nd)
+		if err != nil {
+			return err
+		}
+		collectionNode.AddRawLink("a link", lnk)
+	}
+
+	// update DB with new collection CID
+	col.CID = collectionNode.Cid().String()
+	if err := s.DB.Save(&col).Error; err != nil {
```
en0ma, line 3204: Avoid using `save` for updating (it is good for creating new rows), as it could update fields you don't intend to update. Instead do something like;

```
if err := s.DB.Model(Collection{}).Where("id = ?", col.ID).UpdateColumn("c_id", collectionNode.Cid().String()).Error; err != nil {
		return err
	}

File: handlers.go [hunk 3]
```diff
@@ -3153,6 +3155,66 @@ func (s *Server) handleAddContentsToCollection(c echo.Context, u *User) error {
...
+		return err
+	}
+
+	bserv := blockservice.New(s.Node.Blockstore, offline.Exchange(s.Node.Blockstore))
+	dserv := merkledag.NewDAGService(bserv)
+
+	collectionNode := merkledag.ProtoNode{}
+	for _, content := range contents {
+		curCID := content.Cid.CID
+		nd, err := dserv.Get(context.Background(), curCID)
```
en0ma, line 3190: I don't think this is the best way to do this;

1 - How would you get blocks for content hosted in shuttles?
2 - They can be a simpler way, using an empty dir node 😸 

```
// create a dir to wrap the collection content at this point in time - everytime the collection content changes, this dir cid/commit changes
	collectionNode := unixfs.EmptyDirNode()
	for _, c := range contents {
		collectionNode.AddRawLink(c.Name, &ipld.Link{
			Size: uint64(c.Size),
			Cid:  c.Cid.CID,
		})
	}

File: handlers.go [hunk 4]
```diff
@@ -3153,6 +3155,66 @@ func (s *Server) handleAddContentsToCollection(c echo.Context, u *User) error {
...
+
+	// update DB with new collection CID
+	col.CID = collectionNode.Cid().String()
+	if err := s.DB.Save(&col).Error; err != nil {
+		return err
+	}
+
+	ctx := c.Request().Context()
+	dserv.Add(ctx, &collectionNode)
+	pinstatus, err := s.CM.pinContent(ctx, u.ID, collectionNode.Cid(), collectionNode.Cid().String(), []*CollectionRef{}, []peer.AddrInfo{}, 0, nil)
```
en0ma, line 3210: Regarding pinning, a side effect of this is that a `content` will be created for the collection node, and a deal will be attempted - is this what we want? that is, the contents in the collection will have their own deal, and the collection node (which will now be treated as a content) will have its own too

en0ma: Also, you need to get the host `AddrInfo` in case the pin request is sent to a remote shuttle, it will need the address to find the blocks with bitswap. In the case where the pin is done locally, I am not sure what happens when the host dial itself

en0ma: >In the case where the pin is done locally, I am not sure what happens when the host dial itself

I just checked the libp2p codebase, if we try to connect to self, it will fail with a `ErrDialToSelf` error. We need to do a check at https://github.com/application-research/estuary/blob/master/pinning.go#L102 to only connect if 

```
s.Node.Host.ID != pi.ID
```

the same check should be done https://github.com/application-research/estuary/blob/master/cmd/estuary-shuttle/main.go#L1373. We should have this checks in place even without this PR

cc @jimmylee @whyrusleeping

en0ma:
so sorry, I mistakenly closed this earlier

handlers.go
gmelodie: I did not know `save` did that!  There are other places where this occurs and I think I'd be good to change them too

handlers.go
gmelodie: We definitely don't want the nodes to make deals out of these. I'm unsure if `pinContent` actually fires up something that makes the deals. Can you point me to the code that actually does this? If that's really the case we can always manually pin the CID without adding it to the `contents` table

en0ma: It first creates a content here https://github.com/application-research/estuary/blob/master/pinning.go#L198 and then, for local content, it is added to the pinning queue here https://github.com/application-research/estuary/blob/master/pinning.go#L248 and the doPining operation queues it for a deal here https://github.com/application-research/estuary/blob/master/pinning.go#L116

handlers.go
```diff
@@ -3153,6 +3156,75 @@ func (s *Server) handleAddContentsToCollection(c echo.Context, u *User) error {
...
+	// update DB with new collection CID
+	col.CID = collectionNode.Cid().String()
+	err := s.DB.Model(Collection{}).Where("id = ?", col.ID).UpdateColumn("c_id", collectionNode.Cid().String()).Error
+	if err != nil {
+		return err
+	}
+
+	// transform listen addresses (/ip/1.2.3.4/tcp/80) in full p2p multiaddresses
+	// e.g. /ip/1.2.3.4/tcp/80/p2p/12D3KooWCVTKbuvrZ9ton6zma5LNhCEeZyuFtxcDzDTmWh2qPtWM
+	fullP2pMultiAddrs := []multiaddr.Multiaddr{}
```
en0ma, line 3202: if you want to validate like this, best to do this before updating the collection, so we can fail early. Also, why not just pass `[]peer.AddrInfo{{ID: s.Node.Host.ID(), Addrs: s.Node.Host.Addrs()}}`, any need for the validation?

handlers.go
```diff
@@ -3153,6 +3156,75 @@ func (s *Server) handleAddContentsToCollection(c echo.Context, u *User) error {
...
+func (s *Server) handleCommitCollection(c echo.Context, u *User) error {
+	colid := c.Param("coluuid")
+
+	var col Collection
+	if err := s.DB.First(&col, "uuid = ? and user_id = ?", colid, u.ID).Error; err != nil {
+		return err
+	}
+
+	contents := []Content{}
+	if err := s.DB.Debug().
```
whyrusleeping, line 3176: probably drop the debug flag?

handlers.go
```diff
@@ -3153,6 +3156,75 @@ func (s *Server) handleAddContentsToCollection(c echo.Context, u *User) error {
...
+		Where("collection = ?", col.ID).
+		Joins("left join contents on contents.id = collection_refs.content").
+		Select("contents.*").
+		Scan(&contents).Error; err != nil {
+		return err
+	}
+
+	collectionNode := unixfs.EmptyDirNode()
+	for _, c := range contents {
+		collectionNode.AddRawLink(c.Name, &ipld.Link{
```
whyrusleeping, line 3187: `c.Name` is going to be different than the name of the thing within the collection, we need to make sure we use whatever is on the `CollectionRef` object in question here.

whyrusleeping (changes requested):
A couple things that should be fixed, but overall good.

handlers.go
gmelodie: Ah! look at me copying code without checking

handlers.go
gmelodie: I'm not validating, I'm just transforming the listen addresses in a format that's suited for `peer.AddrInfosFromP2pAddrs`

handlers.go
gmelodie: This is just a name for the link. Also, `CollectionRef` doesn't really have a name for the content, it just has a path for it inside the collection and the ID pointing to the content in the table.

pinning.go
```diff
@@ -149,9 +151,9 @@ func (cm *ContentManager) refreshPinQueue() error {
 	// anyways
```
en0ma, line 151: nitpick: declare `makeDeal = true` here and pass it below in place of `true`

util/content.go
```diff
@@ -73,3 +78,44 @@ func FindCIDType(ctx context.Context, root cid.Cid, dserv ipld.NodeGetter) (cont
...
+	if dir, file := filepath.Split(collectionPath); file == filename { // path ends in the filename
+		collectionPath = dir // only keep the part with dirs
+	}
+	dirs := strings.Split(collectionPath, "/")
+	dirs = removeEmptyStrings(dirs)
+	return dirs, nil
+}
+
+func EnsurePathIsLinked(dirs []string, rootNode *merkledag.ProtoNode, ds format.DAGService) (*merkledag.ProtoNode, error) {
+	lookupNode := rootNode
```
en0ma, line 104: why do this?

gmelodie: Changing variable names so it's obvious that we start with the root node and lookup downwards

en0ma (approved):
LGTM

Issue #180, opened 2022-05-05
Estuary API should not treat missing token as an error


softwareplumber:
When an API request requiring some level of authorization is made, without a token, estuary treats this as an error and creates multiple log entries. 

This is a PITA because, for some reason, the web UI server generates such requests every second or so. The log traffic is pointless. Also the request causes a database search for a token with the value 'undefined', which is unnecessary.

This should be handled as a specific case, returning an unauthorized error to the API client without causing un-necessary database hits or log traffic.

alvin-reyes:
I have prepared a PR for this @softwareplumber. Please check. I think its better if you just add your changes here.

https://github.com/application-research/estuary/tree/empty-auth-check

alvin-reyes:
Resolved by https://github.com/application-research/estuary/pull/181

PR #181, opened 2022-05-05
chore: check if the auth token is empty before starting the tracer

alvin-reyes:
# Change

An issue was raised regarding unnecessary tracer/checks on estuary for request that doesn't come with a non-empty auth token. This is a valid concern since anyone can basically run unlimited request to the API endpoints which can cause DDOS on our endpoints. 

This PR checks the auth token first for endpoints that requires AUTHORIZATION only. 

# Related PR
https://github.com/application-research/estuary/issues/180


gmelodie (approved):
LGTM

Issue #182, opened 2022-05-05
Check if we can connect to providers when pinning


gmelodie:
When Estuary is pinning content [here](https://github.com/application-research/estuary/blob/eaefd9f7b0ac131b33572460a77aa5f500a2b092/pinning.go#L101-L105) tries to connect to the peers that provide the content to be pinned. We want to add a check on that piece of code to make sure at least one of the peers is reachable, otherwise we can fail early saying "unable to connect to any of the provider peers for the content to be pinned".

P.S.: careful when all the peers are aliases for the current node (it will try to dial itself). In that case even though none of the connect attempts work it doesn't matter because we won't have to connect to ourselves when effectively pinning.

alvin-reyes:
Question: The piece of code here already checks if the host can connect to the list of peers. If it connects, it is reachable. If we loop through the peers here and one of them connects, then we can safely say that the specific peer is available. 

Do we want to check and just return an error if no peers is available? 

```
connectedToAtLeastOne := false
	for _, pi := range op.Peers {
		if err := s.Node.Host.Connect(ctx, pi); err != nil {
			log.Warnf("failed to connect to origin node for pinning operation: %s", err)
		} else {
			connectedToAtLeastOne = true
		}
	}

	if !connectedToAtLeastOne {
		log.Errorf("unable to connect to any of the provider peers for pinning operation")
		return nil
	}
```

gmelodie:
Yup that's pretty much it, but if the list of peers all refer to the current node (meaning they're multiaddresses to the Estuary node), they will not connect because a peer cannot connect to itself, but in this case it shouldn't error out.

PR #183, opened 2022-05-05
chore: check legitimate providing peers before pinning operations

alvin-reyes:
# Change

Simple change to check any legitimate available peers before pinning.

# Related Issue
https://github.com/application-research/estuary/issues/182

### DIFF
--- a/pinning.go
+++ b/pinning.go
@@ -100,7 +100,7 @@ func (s *Server) doPinning(ctx context.Context, op *pinner.PinningOperation, cb 
 
 	connectedToAtLeastOne := false
 	for _, pi := range op.Peers {
-		if err := s.Node.Host.Connect(ctx, pi); err != nil {
+		if err := s.Node.Host.Connect(ctx, pi); err != nil && s.Node.Host.ID() != pi.ID {
 			log.Warnf("failed to connect to origin node for pinning operation: %s", err)
 		} else {
 			//	Check if it's trying to connect to itself since we only want to check if the
--- a/pinning.go
+++ b/pinning.go
@@ -98,10 +98,23 @@ func (s *Server) doPinning(ctx context.Context, op *pinner.PinningOperation, cb 
 	ctx, span := s.tracer.Start(ctx, "doPinning")
 	defer span.End()
 
+	connectedToAtLeastOne := false
 	for _, pi := range op.Peers {
 		if err := s.Node.Host.Connect(ctx, pi); err != nil {
 			log.Warnf("failed to connect to origin node for pinning operation: %s", err)
+		} else {
+			//	Check if it's trying to connect to it's own since we only want to check if the
+			//	the connection is between the host and the external/other peers.
+			if s.Node.Host.ID() != pi.ID {
+				connectedToAtLeastOne = true
+			}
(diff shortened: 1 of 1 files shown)


pinning.go
```diff
@@ -98,12 +98,25 @@ func (s *Server) doPinning(ctx context.Context, op *pinner.PinningOperation, cb
 	ctx, span := s.tracer.Start(ctx, "doPinning")
 	defer span.End()
 
+	connectedToAtLeastOne := false
 	for _, pi := range op.Peers {
 		if err := s.Node.Host.Connect(ctx, pi); err != nil {
 			log.Warnf("failed to connect to origin node for pinning operation: %s", err)
+		} else {
+			//	Check if it's trying to connect to it's own since we only want to check if the
```
gmelodie, line 106:
```suggestion
			//	Check if it's trying to connect to itself since we only want to check if the
```

pinning.go
```diff
@@ -98,12 +98,25 @@ func (s *Server) doPinning(ctx context.Context, op *pinner.PinningOperation, cb
...
+			}
 		}
 	}
 
+	//	If it can't connect to any legitimate provider peers, then we fail the entire operation.
+	if !connectedToAtLeastOne {
+		log.Errorf("unable to connect to any of the provider peers for pinning operation")
+		return nil
+	}
+
```
gmelodie, line 119: If all the addresses refer to itself we want the code to succeed. This makes it fail.

gmelodie (changes requested):
Looking good, just a small comment

pinning.go
alvin-reyes: We can just remove the condition that checks the host.ID() and the pi.ID. This means the process will only proceed if it's connected to any providers including itself.

pinning.go
```diff
@@ -98,12 +98,23 @@ func (s *Server) doPinning(ctx context.Context, op *pinner.PinningOperation, cb
...
 	defer span.End()
 
+	connectedToAtLeastOne := false
 	for _, pi := range op.Peers {
 		if err := s.Node.Host.Connect(ctx, pi); err != nil {
 			log.Warnf("failed to connect to origin node for pinning operation: %s", err)
+		} else {
+			//	Check if it's trying to connect to itself since we only want to check if the
+			//	the connection is between the host and the external/other peers.
+			connectedToAtLeastOne = true
```
gmelodie, line 108: If all addresses point to itself, I believe this will fail since they will all error out. Try this instead:
```suggestion
		if err := s.Node.Host.Connect(ctx, pi); err != nil && s.Node.Host.ID() != pi.ID {
			log.Warnf("failed to connect to origin node for pinning operation: %s", err)
		} else {
			//	Check if it's trying to connect to itself since we only want to check if the
			//	the connection is between the host and the external/other peers.
			connectedToAtLeastOne = true
```

gmelodie (changes requested):
I think this is the check we actually need to do.

pinning.go
en0ma: Please address this, so this PR can be merged

en0ma (approved):
LGTM

PR #184, opened 2022-05-06
Report badly formatted config file

softwareplumber:
For backwards compatibility, my initial config development would ignore any errors loading the config file so that the case where the config file does not exist would not break anything (command line/environment config would be handled just as before).

However, this also ignores any formatting errors, causing the system to just silently fail to apply the configuration if there's a syntax error in the file. This is not friendly behavior.

This PR any the error returned by the config file loader. The case where the file does not exist is still ignored, but other errors are reported.


en0ma (approved):
LGTM

en0ma:
@softwareplumber @gmelodie anything holding this back?

gmelodie (approved):
LGTM

Issue #173, opened 2022-05-02
Standardize collection APIs


gmelodie:
Collections APIs use two main arguments: 
- `collection` or `col` to refer to the collection's UUID
- `collectionPath` or `dir` to refer to the path of a certain content inside a collection

We should name those things consistently, not have two names for the same thing.

aeryz:
- I think for collection UUID, `collection` should be used. For example I was playing around with estuary for the first time and thought for a second that `col` means `column`. `collection` is definitely more clear.
- Also `dir` is short version of `directory` which is confusing because absolute path to files also work. (see: `/collections/fs/list`) so I think instead of `dir`, `path` should be used (taking [this issue](https://github.com/application-research/estuary/issues/186) into account).

I can create a PR for both issues if there are no other ideas about naming.

PR #188, opened 2022-05-09
refactor: standardize naming

### LINKED ISSUE #186: Standardize path, collectionPath, and filename
There seems to be very little convention around paths, collectionPaths and filenames all accross the code, we should standardize. My suggestions: 

- Use `path` as the full path inside a collection **including the filename** (e.g. `/dir1/dir2/file.txt`)
- Use `filename` as a variable name instead of `file`, `name` and `filename` scattered around the codebase

The above would allow us to treat files and directories much more easily with packages like [`path`](https://pkg.go.dev/path)

aeryz:
Namings in the API are standardized. Following endpoints are effected:

1. `POST /content/add-ipfs` -> `name` is changed to `filename`
2. `POST /content/add` -> `name` is changed to `filename`.
3. `GET /content/stats` -> `file` is changed to `filename` in response.
4. `GET /collections/fs/list` -> `col` is changed to `collection` and `dir` is changed to `collectionPath`.
5. `GET /collections/fs/add` -> `col` (which is the shorthand for `collection`) is removed. 

Fixes #186 and #173

Note that path params like `coluuid` are not changed since they are not directly used by users. I can change those as well though.

Signed-off-by: aeryz <abdullaheryz@protonmail.com>


gmelodie (approved):
LGTM!

docs/docs.go
```diff
@@ -209,14 +219,14 @@ const docTemplate_swagger = `{
                     {
                         "type": "string",
                         "description": "Collection",
-                        "name": "col",
```
gmelodie, line 212: `collection` or `coluuid`?

docs/swagger.json
```diff
@@ -65,6 +65,23 @@
...
+        "description": "This endpoint is used to delete an existing collection.",
+        "tags": [
+          "collections"
+        ],
+        "summary": "Deletes a collection",
+        "parameters": [
+          {
+            "type": "string",
+            "description": "Collection ID",
+            "name": "coluuid",
```
gmelodie, line 79: `collection` or `coluuid`?

docs/swagger.json
```diff
@@ -202,14 +212,14 @@
           {
             "type": "string",
             "description": "Collection",
-            "name": "col",
+            "name": "collection",
```
gmelodie, line 215: `collection` or `coluuid`?

docs/swagger.yaml
```diff
@@ -118,6 +118,18 @@ paths:
       tags:
         - admin
   /collections/add-content:
+    delete:
+      description: This endpoint is used to delete an existing collection.
+      parameters:
+        - description: Collection ID
+          in: path
+          name: coluuid
```
gmelodie, line 126: `collection` or `coluuid`?

docs/swagger.yaml
```diff
@@ -228,12 +235,12 @@ paths:
       parameters:
         - description: Collection
           in: query
-          name: col
+          name: collection
```
gmelodie, line 238: `collection` or `coluuid`?

docs/swagger.yaml
```diff
@@ -228,12 +235,12 @@ paths:
       parameters:
         - description: Collection
```
gmelodie, line 236: This should say `Collection ID`

docs/swagger.json
```diff
@@ -202,14 +212,14 @@
           {
             "type": "string",
             "description": "Collection",
```
gmelodie, line 214: This should say `Collection ID`

docs/docs.go
```diff
@@ -209,14 +219,14 @@ const docTemplate_swagger = `{
                     {
                         "type": "string",
                         "description": "Collection",
```
gmelodie, line 221: This should say `Collection ID`

cmd/barge/client.go
```diff
@@ -298,7 +298,7 @@ type collectionListResponse struct {
 
 func (c *EstClient) CollectionsListDir(ctx context.Context, col, path string) ([]collectionListResponse, error) {
 	var out []collectionListResponse
-	_, err := c.doRequest(ctx, "GET", fmt.Sprintf("/collections/fs/list?col=%s&dir=%s", col, url.PathEscape(path)), nil, &out)
+	_, err := c.doRequest(ctx, "GET", fmt.Sprintf("/collections/fs/list?collection=%s&collectionPath=%s", col, url.PathEscape(path)), nil, &out)
```
gmelodie, line 301: `collection` or `coluuid`?

docs/swagger.yaml
```diff
@@ -197,11 +209,6 @@ paths:
       description: This endpoint adds a file to a collection
       parameters:
         - description: Collection ID
-          in: query
-          name: col
-          required: true
-          type: string
-        - description: Collection ID Long
           in: query
           name: collection
```
gmelodie, line 213: `collection` or `coluuid`?

handlers.go
```diff
@@ -4860,18 +4860,13 @@ func sanitizePath(p string) (string, error) {
...
 // @Param        path query string true "Path to file"
 // @Produce      json
 // @Router       /collections/fs/add [post]
 func (s *Server) handleColfsAdd(c echo.Context, u *User) error {
-	colid := c.QueryParam("col")
-	colidlong := c.QueryParam("collection")
-	if colid == "" {
-		colid = colidlong
-	}
+	colid := c.QueryParam("collection")
```
gmelodie, line 4869: `coluuid` here

gmelodie (changes requested):
I did a more careful review. I think we should use `coluuid` instead of `collection` for the UUID of the collection, and adjust the docs to always say "Collection ID" instead of just "Collection". `collection` and "Collection" make me think of the collection object, not the UUID of the collection.

aeryz:
> I did a more careful review. I think we should use `coluuid` instead of `collection` for the UUID of the collection, and adjust the docs to always say "Collection ID" instead of just "Collection". `collection` and "Collection" make me think of the collection object, not the UUID of the collection.

Thanks for the in-depth review. Working on it asap.

aeryz:
`collection` is changed to `coluuid` and docs are fixed.

We changed `collection` to `coluuid` but we use `collectionPath` for path in the collection. We might wanna change it to `colpath` to be more consistent. (We shouldn't make it just `path` because `ContentInCollection` struct is used in several other types. Changing it to `path` would create confusions.)

gmelodie:
@aeryz `colpath` sounds like a good variable name. Go for it!

PR #202, opened 2022-05-17
Feat/support both verified and unverified deal making

en0ma:
Estuary currently only makes verified deals (which requires a data cap). 

This PR allows Estuary to be more flexible, to enable making verified/unverified deals

Context: https://filecoinproject.slack.com/archives/C016APFREQK/p1652148940464009


whyrusleeping (approved):

alvin-reyes (approved):
LGTM

PR #204, opened 2022-05-17
Add more auth string checks

gmelodie:
Along the lines of #200, I thought we could go ahead and implement a few more checks for the auth string, so I packaged that inside a function and added a bit more detail into the checks. This will help us deliver more accurate error messages.


en0ma (approved):
LGTM

Issue #81, opened 2021-12-10
bug: Verify doesn't match against identical CIDs with different versions


Jorropo:
Take this CID for example: `bafybeih4i7cwnx6kbgsqme22y5prc2v27jsetvvbuksrq6d2lf3r3qg37i` have the "This CID is verified on Estuary" message.
However it's CIDv0 conversion `QmfKSvYvVXLSo5vpb1ZgBzJMNkWYKFRefSPPSoCdQSuEMF` doesn't have it.

Note verify actually works with different multibase if they are all CIDv1 (`uAXASIPxHxWbfygmlBhNax18Rarr6ZEnWoaKlGHh6WXcdwNv6`).

PR #206, opened 2022-05-18
verify should support identical CIDs with different versions and base

en0ma:
Currently, the `verify-cid` endpoint only supports `cidv1` see ([bug: Verify doesn't match against identical CIDs with different versions](https://github.com/application-research/estuary/issues/81)). This PR will allow support for both `cidv1` (and it's various base) and `cidv0`.

Below are test for the same CID for;

cidv1
Screenshot 2022-05-18 at 15 51 27

cidv1 base64url
Screenshot 2022-05-18 at 15 50 22

cidv0
Screenshot 2022-05-18 at 15 50 52

closes https://github.com/application-research/estuary/issues/81

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -1331,10 +1331,11 @@ func (s *Server) handleGetContentByCid(c echo.Context) error {
 		return err
 	}
 
-	// TODO: check both cidv0 and v1 for dag-pb cids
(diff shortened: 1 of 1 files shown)


gmelodie (approved):
LGTM

PR #207, opened 2022-05-18
Fix/report-correct-error-when-shuttle-is-not-online

en0ma:
- For boost deals, if shuttle addr is nil, it means shuttle is not online - the error should say so

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


gmelodie (changes requested):
Change looks good, but I wonder if that's really the only case where `addrInfo` will be `nil`. Maybe we should add a call to `cm.shuttleIsOnline` and, if that returns `false` we add "shuttle is not online" to the error string. It essentially does the same thing but I think it's more readable

gmelodie (approved):
Looking good!

PR #213, opened 2022-05-19
Fix isValidAuth function

gmelodie:
#204 introduces a bug where it doesn't correctly trim the `EST`, `ARY` and `SECRET` parts of an auth string before calling `uuid.Parse`, and thus causes it to fail. This fixes that.


en0ma (approved):

alvin-reyes (approved):
LGTM

whyrusleeping:
just to make sure, you tested this with a normal api request with a normal api key right? like the `/viewer` endpoint?

gmelodie:
@whyrusleeping yes (see SS below), but I also added some unit tests to be sure.
image

Issue #187, opened 2022-05-06
DHT Announces are sent from the master instead of the shuttle making files unavailable


Jorropo:
- 1.car: bafybeic6vqc3omyauh25qjr5nysq3zxokscru7vgbuimfyjez5kyfnsrku
- 2.car: bafybeic62wbngy65cqpuorrchstvffc7h2quczevmmraradrxbkzt5vdoq

I can't fetch thoses files on estuary, verify say they are there, but
```
ipfs ls bafybeic6vqc3omyauh25qjr5nysq3zxokscru7vgbuimfyjez5kyfnsrku --size=false --resolve-type=false
```
never resolves (it timeout if I put one)
If I findprovs, it find
```console
$ ipfs dht findprovs bafybeic6vqc3omyauh25qjr5nysq3zxokscru7vgbuimfyjez5kyfnsrku
12D3KooWLV128pddyvoG6NBvoZw7sSrgpMTPtjnpu3mSmENqhtL7
```
However this node doesn't have the file:
```console
$ vole bitswap check bafybeic6vqc3omyauh25qjr5nysq3zxokscru7vgbuimfyjez5kyfnsrku /ip4/35.74.45.12/tcp/6745/p2p/12D3KooWLV128pddyvoG6NBvoZw7sSrgpMTPtjnpu3mSmENqhtL7
{"Found":false,"Responded":true,"Error":null}
```
(**"Found": false**)

hashaxis:
+1 On this issue.

teacoat:
+1 here too

en0ma:
closing this as it has been addressed

PR #214, opened 2022-05-19
the blockstore config flag is not a path and making it absolute is incorrect

whyrusleeping:
This is whats causing the problems seen in #187 
The DHT announcement issue is a separate one that is not the actual root cause in the linked issue


Jorropo:
Seems like darkmagic for it to fix #187 but if you say so then I trust you.

PR #215, opened 2022-05-19
avoid trying to traverse through sector cids in pinned data

whyrusleeping:
This will make it slightly better if we try to pin filecoin state roots


en0ma:
@whyrusleeping I am just curious to learn more on why this needed

en0ma:
@whyrusleeping I understand now 😄

en0ma (approved):
LGTM

alvin-reyes:
> @whyrusleeping I am just curious to learn more on why this needed

@whyrusleeping @en0ma Just for my understanding, can you confirm: 

we don’t need to traverse through the sectors cids because its not really part of the file cids (links) and will have additional overhead?

en0ma:
> we don’t need to traverse through the sectors cids because its not really part of the file cids (links) and will have additional overhead?

What I understand is, from a storage client perspective (estuary as a storage client), clients almost don't deal with filecoin state roots (besides piece CID needed for deals, and the blocks are never stored on the client-side). So these roots should not be in the blockstore, If they even get into the blockstore, this PR will make sure they are not traversed - as they wont be anything to travserve

gmelodie (approved):
LGTM

PR #220, opened 2022-05-20
handle shuttle getViewer error

en0ma:
Currently, shuttle-proxy this https://github.com/application-research/estuary/blob/master/cmd/shuttle-proxy/main.go#L105 does not return any error even if the `getViewer` returns an error. This causes it to panic at https://github.com/application-research/estuary/blob/master/cmd/shuttle-proxy/main.go#L132. 

This PR makes sure the above does not happen. It also adds shuttle-proxy to build (I could easily test with this).

### DIFF
--- a/cmd/shuttle-proxy/main.go
+++ b/cmd/shuttle-proxy/main.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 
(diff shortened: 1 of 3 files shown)


cmd/shuttle-proxy/main.go
```diff
@@ -73,23 +74,26 @@ func (p *Proxy) getViewer(auth string) (*util.ViewerResponse, int, error) {
...
-
 	defer resp.Body.Close()
 
 	if resp.StatusCode != 200 {
-		return nil, resp.StatusCode, fmt.Errorf("auth check failed")
+		bodyBytes, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, 500, err
+		}
+		return nil, resp.StatusCode, fmt.Errorf("auth check failed: %s", bodyBytes)
```
whyrusleeping, line 90: probably want to do `string(bodyBytes)`

en0ma: This works too - I tested it

en0ma:
@alvin-reyes @gmelodie can you guys help look at this so I can merge it

alvin-reyes (approved):
LGTM.

PR #222, opened 2022-05-23
chore: force use swag 1.8.1

alvin-reyes:
There's no 86_64 darwin/linux build for 1.8.2 and this causes the swag generation to fail.

This is a temporary fix to use 1.8.1 swag. I'm going to move the entire swagger functionality to it's own project.


en0ma (approved):
LGTM

PR #226, opened 2022-05-24
properly count file size

en0ma:
Currently, the API response for the UI that shows the files incorrectly doubles the file size. This is because it counts both aggregates and their contents. This PR fixes this issue by properly counting content sizes for all 3 possible scenarios estuary can have;

1 - When content is not aggregated and not a split content (`aggregated_in = 0`) - content size is counted directly.
2 - When content is aggregated (`aggregated_in > 0`) - the aggregate content size is counted
3 - When content is a split - it will take either option 1 or 2 - so the above rules will apply

before
Screenshot 2022-05-24 at 15 17 39

After
Screenshot 2022-05-24 at 15 15 39

closes https://github.com/application-research/estuary-www/issues/38

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -2857,9 +2857,10 @@ type userStatsResponse struct {
 // @Router       /user/stats [get]
 func (s *Server) handleGetUserStats(c echo.Context, u *User) error {
 	var stats userStatsResponse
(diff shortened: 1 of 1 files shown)


gmelodie (approved):
LGTM

PR #227, opened 2022-05-24
Do not double log echo HTTP errors

en0ma:
Currently, echo HTTP error floods Grafana and gets in the way when investigating issues. I do not think we should be logging echo HTTP errors explicitly since these errors are already logged by the logging middleware and so it is redundant and misleading (since it shows up as `ERROR` e.g `No Content`). 

E.g - all of this error is not useful - took me days to know what is going on (and it is not actionable)
Screenshot 2022-05-24 at 17 49 56

This PR is a minor refactor and moves explicit logging to non-echo HTTP error handlers


gmelodie (approved):
LGTM

PR #229, opened 2022-05-24
chore: added test scripts

alvin-reyes:
# Changes
In order for us to have some quality gates before we deploy our changes, we need to ensure that the major functionality of estuary is at the very least functioning properly. 

This PR includes some test cases that covers crucial functionality such as upload, download, api key validations, api calls and an entirely new unit test case for handler.go.

This will also generate a clean report for all the test runs.

+ Added regression and chaos testing
+ Added handler.go unit tests
+ Report generation tool for test cases

## To run
```
cd tests
make all
```


handlers_test.go
```diff
@@ -0,0 +1,152 @@
+package main
```
en0ma, line 1: I think we should remove this for now. I know it is boiler plate, but we should strive for a blackboox test, which will come when the refactoring starts.

scripts/samples/add-file-to-collection.sh
```diff
@@ -15,9 +35,7 @@ echo "$fname"
 EST_HOST="https://shuttle-4.estuary.tech"
 #EST_HOST="https://shuttle-5.estuary.tech"
 
-COLLECTION=""
-
+COLLECTION="185d7ffc-85d8-4b73-86bb-1ca6419bc10d"
```
en0ma, line 38: while this collection always be available? why not also allow passing in a collection uuid?

handlers_test.go
alvin-reyes: Yes. this handlers_test.go is just a placeholder for now for more white-box testing (unit tests).

scripts/samples/add-file-to-collection.sh
```diff
@@ -15,9 +35,7 @@ echo "$fname"
...
 set -x
-#curl --progress-bar -X POST -H "Authorization: Bearer $ESTUARY_TOKEN" -F "data=@$1" -F "name=$fname" $EST_HOST/content/add
-curl --progress-bar -X POST -H "Authorization: Bearer $ESTUARY_TOKEN" -F "data=@$1" -F "name=$fname" $EST_HOST/content/add\?collection="$COLLECTION"
+curl --progress-bar -X POST -H "Authorization: Bearer $ESTUARY_TOKEN" -F "data=@$file" -F "name=$fname" $EST_HOST/content/add\?collection="$COLLECTION"
```
en0ma, line 40: the `name` params is now called `filename` - see changes from https://github.com/application-research/estuary/pull/188/files. This affects the `content/add` and `content/add-ipfs` endpoints

handlers_test.go
```diff
@@ -0,0 +1,152 @@
...
+		c echo.Context
+		u *User
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantErr assert.ErrorAssertionFunc
+	}{
+		// TODO: Add test cases.
```
gmelodie, line 48: I don't think this style of tests will work in this case, because for every test case we would have to set the DB and other things all over again. I'm working on some tests that mock the database, but the roadmap for those are basically: autoretrieve > refactor > tests. So we might want to remove this file for now.

scripts/samples/add-file-to-collection.sh
```diff
@@ -1,11 +1,31 @@
 #!/bin/bash
 
+#!/bin/bash
+
+###################################################################
+#Script Name	: add-file-to-collection.sh
+#Description	: This is a script file that runs a curl command to add a file to the estuary content server.
```
gmelodie, line 7: I think it's better to use Pestuary or any other SDK for this so we could to add the add, remove, list from the same python script.

tests/data/config/run.config
```diff
@@ -0,0 +1,14 @@
+ESTUARY_TOKEN=ESTabcdefghARY
```
gmelodie, line 1: This is not going to pass as a valid estuary token (has to have a valid uuid between `EST` and `ARY`)

tests/chaos/download-concurrent-large-files.sh
```diff
@@ -0,0 +1,21 @@
...
+. ../data/config/run.config
+
+echo '#####################################'
+echo `basename "$0"`
+echo '#####################################'
+echo ''
+
+for i in {0..3}
+do
+  nohup ./download-large-files.sh > logs/$i-download-large-files.log &
```
gmelodie, line 19: I think this `download-large-file` approach will take way too long (also bandwidth if we're thinking of CI/CD on the cloud). A better approach would be to use:
```
sudo fallocate -l 2G bigfile
```
[source](https://unix.stackexchange.com/a/269221/235577)

gmelodie (dismissed):
A couple things to note:
1. We'd probably be better off using Pestuary to add files (deal with tokens and uuids returned by the API)
2. The `handlers_test.go` file will have to change a lot, so we better not commit it now (wait until refactor is done)
3. Use `fallocate` to create large files instead of downloading them

scripts/samples/add-file-to-collection.sh
alvin-reyes: The point of the tests here is to ensure that the estuary endpoints will respond according to the raw request body we provide. I want to run test as raw (low-level) as possible by using `curl` and passing the correct request information (headers, auth, content-type). 

IMO, the scenarios you pointed here should all be unit tests in the `pestuary` repo.

en0ma (approved):

tests/chaos/download-concurrent-large-files.sh
alvin-reyes: Ok. I'm ok using this but just an FYI that this assumes we have root access of the box.

gmelodie: @jimmylee

gmelodie (approved):

PR #223, opened 2022-05-23
Fix collectionRef sql ambiguous

leotons:
sql execution exception：/estuary/handlers.go:4839 ERROR: column reference “path” is ambiguous (SQLSTATE 42702);

PR #231, opened 2022-05-25
Fix collectionRef sql ambiguous

en0ma:
This fixes the error `sql execution exception：/estuary/handlers.go:4839 ERROR: column reference “path” is ambiguous (SQLSTATE 42702);`. 

Thanks to @leotons for working on this (https://github.com/application-research/estuary/pull/223)

closes https://github.com/application-research/estuary/issues/233


alvin-reyes (approved):
LGTM

Issue #135, opened 2022-03-24
Estuary command-line options passed unfiltered to lotus client


softwareplumber:
I was asking myself the question: is the 'repo' flag in the Estuary command line actually used?

Searching the actual estuary code, it isn't, but then I happened upon this line in main.go

		api, closer, err := lcli.GetGatewayAPI(cctx)

Where cctx is the Estuary cli context. The 'repo' flag is (I think) picked up deep inside lotus/cli/util.

But: what if estuary command line options ever conflicted with lotus command line options? How would we ever know?

PR #234, opened 2022-05-25
Use clean context for lotus gateway cli

en0ma:
This PR makes sure the context passed to Lotus CLI is clean and only has intended flags (see https://github.com/application-research/estuary/issues/135). 

It also brings some config refactor to standardize shuttle and estuary. 

I have tested both Estuary and Shuttle and both pick up the gateway API URL and fetches tipsets from the blockchain.

Closes https://github.com/application-research/estuary/issues/135


cmd/estuary-shuttle/main.go
```diff
@@ -170,8 +167,10 @@ func main() {
 
 	app.Flags = []cli.Flag{
 		&cli.StringFlag{
-			Name:  "repo",
-			Value: "~/.lotus",
+			Name:    "node-api-url",
+			Usage:   "lotus api gateway url",
+			Value:   cfg.Node.ApiURL,
+			EnvVars: []string{"FULLNODE_API_INFO"},
```
gmelodie, line 173: Cool! Didn't know you could do that

main.go
```diff
@@ -512,7 +500,14 @@ func main() {
 			return err
 		}
 
-		api, closer, err := lcli.GetGatewayAPI(cctx)
+		// send a clean cctx down into lotus - it needs only the node "api-url"
```
gmelodie, line 503:
```suggestion
		// send a CLI context to lotus that contains only the node "api-url" flag set, so that other flags don't accidentally conflict with lotus cli flags
```

gmelodie (changes requested):
I'd like the comment to be a bit more explicit about why we are creating a brand new cli context. Other than that code looks solid.

gmelodie (approved):

PR #237, opened 2022-05-26
move types package into pinner, it was the only user of it

whyrusleeping:


whyrusleeping:
I'm going to pull the pin manager code out of the estuary repo so i can reuse it elsewhere, this is just prepwork for that

Issue #228, opened 2022-05-24
Shuttle provider context gets canceled too often


en0ma:
After this https://github.com/application-research/estuary/pull/191 change, I can confirm shuttle CID announcement context are canceled too often, so most root CID of files uploaded are not announced.

<img width="1263" alt="Screenshot 2022-05-24 at 19 38 45" src="https://user-images.githubusercontent.com/13554411/170109011-3de90b9e-5462-4de5-9666-3699da362242.png">

PR #239, opened 2022-05-31
increase providing timeout to 15secs

en0ma:
Right now, the providing context timeout of 10 secs, is not sufficient. After testing, it turns out 15secs is just enough. 

Closes https://github.com/application-research/estuary/issues/228

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -1144,7 +1144,7 @@ func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 }
 
 func (s *Shuttle) Provide(ctx context.Context, c cid.Cid) error {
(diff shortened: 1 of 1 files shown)


cmd/estuary-shuttle/main.go
```diff
@@ -1144,7 +1144,7 @@ func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 }
 
 func (s *Shuttle) Provide(ctx context.Context, c cid.Cid) error {
-	subCtx, cancel := context.WithTimeout(ctx, time.Second*10)
+	subCtx, cancel := context.WithTimeout(ctx, time.Second*15)
```
gmelodie, line 1147: This should be a variable like `provideTimeout` (after the refactor this would go into the `constants` package)

gmelodie (changes requested):
Nice fix! Just one comment

PR #249, opened 2022-06-06
Only upload content with size over defaultContentSizeLimit if splitting is enabled

en0ma:
Right now, when users upload contents with size over https://github.com/application-research/estuary/blob/master/replication.go#L55 and splitting is not enabled, such content gets stuck as deals will never be made for them leading to a not so good UX e.g https://filecoinproject.slack.com/archives/C016APFREQK/p1647824002708279

This PR tries to improve the UX by verifying that if splitting is not turned on, and the user is uploading content over the `defaultContentSizeLimit`, then we prevent the upload and notify the user.

Both primary and shuttle have been tested on both `add` and `add-car` endpoints for when content is over and under the defaultContentSizeLimit

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -1098,7 +1098,7 @@ func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 		return err
 	}
 
-	// if splitting is disabled and uploaded content size is over content size limit
+	// if splitting is disabled and uploaded content size is greater than content size limit
 	// reject the upload, as it will only get stuck and deals will never be made for it
 	if !u.FlagSplitContent() && mpf.Size > util.DefaultContentSizeLimit {
 		return &util.HttpError{
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -1220,7 +1220,7 @@ func (s *Shuttle) handleAddCar(c echo.Context, u *User) error {
 		}
 	}
 
-	// if splitting is disabled and uploaded content size is over content size limit
+	// if splitting is disabled and uploaded content size is greater than content size limit
 	// reject the upload, as it will only get stuck and deals will never be made for it
 	if !u.FlagSplitContent() {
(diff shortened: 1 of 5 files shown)


cmd/estuary-shuttle/main.go
```diff
@@ -1091,12 +1098,21 @@ func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 		return err
 	}
 
+	// if splitting is disabled and uploaded content size is over content size limit
```
gmelodie, line 1101:
```suggestion
	// if splitting is disabled and uploaded content size is greater than content size limit
```

cmd/estuary-shuttle/main.go
```diff
@@ -1204,6 +1220,28 @@ func (s *Shuttle) handleAddCar(c echo.Context, u *User) error {
 		}
 	}
 
+	// if splitting is disabled and uploaded content size is over content size limit
```
gmelodie, line 1223:
```suggestion
	// if splitting is disabled and uploaded content size is greater than content size limit
```

handlers.go
```diff
@@ -546,6 +546,28 @@ func (s *Server) handleAddCar(c echo.Context, u *User) error {
 		}
 	}
 
+	// if splitting is disabled and uploaded content size is over content size limit
```
gmelodie, line 549:
```suggestion
	// if splitting is disabled and uploaded content size is greater than content size limit
```

handlers.go
```diff
@@ -702,14 +724,23 @@ func (s *Server) handleAdd(c echo.Context, u *User) error {
...
 	}
-
 	defer form.RemoveAll()
 
 	mpf, err := c.FormFile("data")
 	if err != nil {
 		return err
 	}
 
+	// if splitting is disabled and uploaded content size is over content size limit
```
gmelodie, line 734:
```suggestion
	// if splitting is disabled and uploaded content size is greater than content size limit
```

gmelodie (approved):

alvin-reyes (approved):
LGTM

PR #250, opened 2022-06-07
dont use commp from request params

en0ma:
We have to drop support for uploading `commp` with car file and instead compute it on Estuary. Given Estuary unpacks the car into blocks to the blockstore for transmission; on repacking, it might run into cases where the uploaded `commp` will not match computed (depending on how the user packed it vs how Estuary re-packed it) - This leads to late errors on the miner side. It is best we always compute so we can be sure it will always validate on the miner side.

As for already uploaded `commp`, such contents will be stuck and deals cannot be made for them, except the stored uploaded `commp` are deleted and deals making should be re-attempted.  cc: @brendalee

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -46,7 +46,6 @@ import (
 	"github.com/cenkalti/backoff/v4"
 	"github.com/filecoin-project/go-address"
 	datatransfer "github.com/filecoin-project/go-data-transfer"
-	commcid "github.com/filecoin-project/go-fil-commcid"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/chain/types"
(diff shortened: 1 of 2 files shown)


en0ma:
> What is `commp`?

`pieceCID`

brendalee:
@gmelodie - here is why the commp is important: https://spec.filecoin.io/systems/filecoin_files/piece/#:~:text=The%20piece%20CID%20or%20CommP,Piece%20CID%20on%20their%20side.

gmelodie:
I wonder if we could re-compute it ourselves and, if it's identical to what the user uploaded, we allow the upload to continue, otherwise error out. Of course, if we're doing like that it means we'll compute it anyways so there is an argument to be made along the lines of "why bother".

en0ma:
> I wonder if we could re-compute it ourselves and, if it's identical to what the user uploaded, we allow the upload to continue, otherwise error out. Of course, if we're doing like that it means we'll compute it anyways so there is an argument to be made along the lines of "why bother".

That is why it is easier to ignore the one sent by the user

gmelodie (approved):

alvin-reyes (commented):
Do we need to document the removal of this somewhere? I think we should have a documentation (even on the code) to specify that this was removed and the reason.

en0ma:
> Do we need to document the removal of this somewhere? I think we should have a documentation (even on the code) to specify that this was removed and the reason.
> 
> My other concern is negative impact to existing consumers. I'm assuming we did some smoke tests for this?

Regarding #1 
Not sure why the removal has to be documented. I am sure this was only used to skip the step to compute piece CID (which is still always computed even with this piece of code). 
Regarding #2
There are no side effects on existing contents. In fact, we need to fix existing contents that currently have this mismatch issue. Check Grafana, we will see a bunch of them.

PR #255, opened 2022-06-08
add estuary prep script

anjor:
Moving https://github.com/wfschrec/estuary-stuff/blob/main/prep-script.bash to the estuary repo for better discoverability.


gmelodie (approved):
We should also add a link for that on the README somewhere

Issue #168, opened 2022-04-26
/collections/fs/list occassionally returns null for previously uploaded path 


3456091:
I'm still investigating the specifics, but the gist is that if I run :

`curl -XPOST -H "Authorization: Bearer $AUTH_TOK" "https://upload.estuary.tech/content/add?collection=$COL&collectionPath=/$(basename $f)" -F data=@"$f"`

Then:

`curl  -H "Authorization: Bearer $AUTH_TOK" "https://api.estuary.tech/collections/fs/list?col=$COL&dir=/$(basename "$f")"`

I normally receive something like this back from the latter:

`[{"name":"19971205.gif","type":"file","size":37412,"contId":28528783,"cid":"$CID"}]`

However, every so often I receive `null` instead. This is definitely intermittent, but once it's happened it seems to persist.

gmelodie:
This is interesting... Could you:
1. Fire up a local estuary node with `./estuary --logging 2>&1`
2. Try to reproduce the issue
3. Post the estuary logs that the node outputs here?

corinne-antonia:
Hey @3456091, did you see @gmelodie response? If so, does this solve it for you? Can I close this?

3456091:
Well, the original symptom only occurred sporadically, so I can't judge whether the fix works. 🙂 No objections to closing, though.

PR #258, opened 2022-06-08
Fix collection inconsistencies on /content/add

gmelodie:
Inspired by #168 this PR does the following:
1. fixes `/content/add` so it doesn't create `CollectionRef`s that have a `null` path field
2. changes the database to require a couple fields on `collection_refs` to be not null
3. makes `coluuid` and `colpath` query parameters on `/content/add`, not form values (consistent with other endpoints that use collections)
4. makes `/collections/fs/list` return an empty list instead of `null` when no entries are found

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -1100,8 +1100,8 @@ func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 	defer fi.Close()
 
 	cic := util.ContentInCollection{
-		CollectionID:   c.FormValue("coluuid"),
-		CollectionPath: c.FormValue("colpath"),
+		CollectionID:   c.QueryParam("coluuid"),
+		CollectionPath: c.QueryParam("colpath"),
 	}
 
 	bsid, bs, err := s.StagingMgr.AllocNew()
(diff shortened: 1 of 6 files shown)


en0ma:
>makes coluuid and colpath query parameters on /content/add, not form values (consistent with other endpoints that use collections)
@gmelodie please make sure this change is done across shuttle, shuttle-proxy, barge, and any scripts

cmd/estuary-shuttle/main.go
```diff
@@ -1373,6 +1373,7 @@ func (s *Shuttle) addrsForShuttle() []string {
 }
 
 func (s *Shuttle) createContent(ctx context.Context, u *User, root cid.Cid, fname string, cic util.ContentInCollection) (uint, error) {
+	log.Debugf("createContent> cid: %v, filename: %s, collection: %+v", root, fname, cic)
```
en0ma, line 1376: you forgot this

en0ma:
@gmelodie I don't see shuttle-proxy and barge changes

cmd/estuary-shuttle/main.go
gmelodie: I was intending on letting it there on purpose for instrumentation's sake. Do you want me to use another function or some other format?

gmelodie:
@en0ma shuttle-proxy [propagates query params](https://github.com/application-research/estuary/blob/8c2e39996e41dd1a786fe37e034bffbd9018bfce/cmd/shuttle-proxy/main.go#L152) so we don't need to update it, and barge doesn't seem to support collections (maybe add the support on a separate PR?)

Issue #259, opened 2022-06-09
Implement Peering on Estuary


alvin-reyes:
**Is your feature request related to a problem? Please describe.**
There are multiple instances were in estuary is failing to serve large content and this affects the quality of the core service which is serving large files to multiple clients. 

**Describe the solution you'd like**
We can implement [Peering](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#peering) (from ipfs) to prioritize the connections to a certain list of nodes that will allow estuary to maximize the availability to serve large content from these nodes.

PR #260, opened 2022-06-09
feat: peering implementation

alvin-reyes:
# Changes
Implemented Peering based on [IPFS peering](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#peering)

- Introduce Peering Service
- Introduce Peering Endpoints (add, remove, list, start, stop, get state, check connection)
- Introduce Pre-defined PeeringPeers based on content providers [here](https://docs.ipfs.io/how-to/peering-with-content-providers/#content-provider-list)

# Related Issue
https://github.com/application-research/estuary/issues/259

# How to use
## Add peering.peers via estuary cli
```
./estuary --peering-peers '[{"ID": "12D3KooWEGeZ19Q79NdzS6CJBoCwFZwujqi5hoK8BtRcLa48fJdu", "Addrs": ["/ip4/145.40.96.233/tcp/4001"] },{ "ID": "12D3KooWBnmsaeNRP6SCdNbhzaNHihQQBPDhmDvjVGsR1EbswncV", "Addrs":["/ip4/18.1.1.2/tcp/4001", "/ip4/147.75.87.85/tcp/4001"]}]'
```
## Add peering.peers via estuary-shuttle cli
```
./estuary-shuttle --peering-peers '[{"ID": "12D3KooWEGeZ19Q79NdzS6CJBoCwFZwujqi5hoK8BtRcLa48fJdu", "Addrs": ["/ip4/145.40.96.233/tcp/4001"] },{ "ID": "12D3KooWBnmsaeNRP6SCdNbhzaNHihQQBPDhmDvjVGsR1EbswncV", "Addrs":["/ip4/18.1.1.2/tcp/4001", "/ip4/147.75.87.85/tcp/4001"]}]'
```

### Add peers
```
curl --location --request POST 'localhost:3004/admin/peering/peers/add' \
--header 'Authorization: Bearer [REDACTED]' \
--header 'Content-Type: application/json' \
--data-raw '[{"ID": "12D3KooWEGeZ19Q79NdzS6CJBoCwFZwujqi5hoK8BtRcLa48fJdu", "Addrs": ["/ip4/145.40.96.233/tcp/4001"] },{ "ID": "12D3KooWBnmsaeNRP6SCdNbhzaNHihQQBPDhmDvjVGsR1EbswncV", "Addrs":["/ip4/18.1.1.2/tcp/4001", "/ip4/147.75.87.85/tcp/4001"]}]'
```

### Remove
```
curl --location --request DELETE 'localhost:3004/admin/peering/peers/remove' \
--header 'Authorization: Bearer [REDACTED]' \
--header 'Content-Type: application/json' \
--data-raw '["12D3KooWEGeZ19Q79NdzS6CJBoCwFZwujqi5hoK8BtRcLa48fJdu"]'
```

### List
```
curl --location --request GET 'localhost:3004/admin/peering/peers/list' \
--header 'Authorization: Bearer [REDACTED]'
```

### Start
```
curl --location --request POST 'localhost:3004/admin/peering/start' \
--header 'Authorization: Bearer [REDACTED]'
```
### Stop
```
curl --location --request POST 'localhost:3004/admin/peering/stop' \
--header 'Authorization: Bearer [REDACTED]'
```

# Verification
## Add peers
image

## Remove peers
image
 
## List all peers
image

## Start peering
image

## Stop peering
image


en0ma:
@alvin-reyes why not use the go-ipfs peering package at https://github.com/ipfs/go-ipfs/tree/master/peering

node/modules/peering/peering.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+type PeeringPeer struct {
+	ID        string   `json:"ID"`
+	Addrs     []string `json:"Addrs"`
+	Connected bool     `json:"Connected,omitempty"`
+}
+
+type EstuaryPeeringService struct {
```
en0ma, line 15: why not use `peering.PeeringService` directly? why wrap it in `EstuaryPeeringService` ?

node/modules/peering/peering.go
```diff
@@ -0,0 +1,45 @@
+package peering
```
en0ma, line 1: why not `node/peering/peering.go` why `modules?

node/modules/peering/peering.go
en0ma: nevermind, this wrapper will be good for testing

node/modules/peering/peering.go
en0ma: ignore this, we can always address this when refactor starts

node/node.go
```diff
@@ -168,6 +175,29 @@ func Setup(ctx context.Context, init NodeInitializer) (*Node, error) {
 	}
 
 	h, err := libp2p.New(opts...)
+
+	//	peering service
+	hPeer := peering.NewEstuaryPeeringService(h)
+
+	peeringPeerList := cfg.PeeringPeers
+	peeringPeerList = append(peeringPeerList, bootStrapPeeringPeers...)
```
en0ma, line 183: nit: condense to `peeringPeerList := append(cfg.PeeringPeers, bootStrapPeeringPeers...)`

node/node.go
```diff
@@ -168,6 +175,29 @@ func Setup(ctx context.Context, init NodeInitializer) (*Node, error) {
 	}
 
 	h, err := libp2p.New(opts...)
+
+	//	peering service
+	hPeer := peering.NewEstuaryPeeringService(h)
```
en0ma, line 180: nit: peerServ

node/modules/peering/peering.go
```diff
@@ -0,0 +1,45 @@
...
+type EstuaryPeeringService struct {
+	*peering.PeeringService
+}
+
+type EstuaryPeeringNotifee peering.PeeringService
+
+func NewEstuaryPeeringService(host host.Host) *EstuaryPeeringService {
+	estuaryPeeringService := &EstuaryPeeringService{}
+	estuaryPeeringService.PeeringService = peering.NewPeeringService(host)
+	return estuaryPeeringService
```
en0ma, line 24: nit: condense 22-24 -> `return &EstuaryPeeringService{peering.NewPeeringService(host)}`

node/modules/peering/peering.go
```diff
@@ -0,0 +1,45 @@
...
+	ID        string   `json:"ID"`
+	Addrs     []string `json:"Addrs"`
+	Connected bool     `json:"Connected,omitempty"`
+}
+
+type EstuaryPeeringService struct {
+	*peering.PeeringService
+}
+
+type EstuaryPeeringNotifee peering.PeeringService
```
en0ma, line 19: can remove, not used anywhere

node/modules/peering/peering.go
```diff
@@ -0,0 +1,45 @@
...
+
+func NewEstuaryPeeringService(host host.Host) *EstuaryPeeringService {
+	estuaryPeeringService := &EstuaryPeeringService{}
+	estuaryPeeringService.PeeringService = peering.NewPeeringService(host)
+	return estuaryPeeringService
+}
+
+func (ps *EstuaryPeeringService) Start() error {
+	return ps.PeeringService.Start()
```
en0ma, line 28: can be `return ps.Start()`, same for the others

main.go
```diff
@@ -191,6 +192,17 @@ func overrideSetOptions(flags []cli.Flag, cctx *cli.Context, cfg *config.Estuary
 				return fmt.Errorf("failed to parse announce address %s: %w", cctx.String("announce"), err)
 			}
 			cfg.Node.AnnounceAddrs = []string{cctx.String("announce")}
+		case "peering":
```
en0ma, line 195: `peering-peers` to match with ipfs config

main.go
```diff
@@ -292,6 +304,11 @@ func main() {
 			Usage:   "announce address for the libp2p server to listen on",
 			EnvVars: []string{"ESTUARY_ANNOUNCE"},
 		},
+		&cli.StringFlag{
+			Name:    "peering",
```
en0ma, line 308: `peering-peers` to match with ipfs config

handlers.go
```diff
@@ -432,6 +443,130 @@ func (s *Server) handleStats(c echo.Context, u *User) error {
...
+					Code:    400,
+					Message: util.ERR_PEERING_PEERS_ADD_ERROR,
+				}
+			}
+			multiAddrs = append(multiAddrs, a)
+		}
+
+		peerParamId, _ := peer.Decode(peerParam.ID)
+		//	add the peer(s)
+		s.Node.Peering.AddPeer(
```
en0ma, line 471: why not build/verify the multiaddr before adding it for peering? if it fails halfway, it means partial adding. Except this it your intention

handlers.go
```diff
@@ -432,6 +443,130 @@ func (s *Server) handleStats(c echo.Context, u *User) error {
...
+// handlePeeringStart godoc
+// @Summary      Start Peering
+// @Description  This endpoint can be used to start the Peering Service
+func (s *Server) handlePeeringStart(c echo.Context) error {
+	type GenericResponse struct {
+		Message string `json: "Message"`
+	}
+	err := s.Node.Peering.Start()
+	if err != nil {
+		fmt.Println(err)
```
en0ma, line 534: log instead?

handlers.go
```diff
@@ -432,6 +443,130 @@ func (s *Server) handleStats(c echo.Context, u *User) error {
...
+			Message: util.ERR_PEERING_PEERS_STOP_ERROR,
+		}
+	}
+	return c.JSON(200, GenericResponse{Message: "Peering Stopped."})
+}
+
+// handlePeeringStatus godoc
+// @Summary      Check Peering Status
+// @Description  This endpoint can be used to check the Peering status
+func (s *Server) handlePeeringStatus(c echo.Context) error {
```
en0ma, line 563: how is this endpoint supposed to work?

cmd/estuary-shuttle/main.go
```diff
@@ -273,10 +286,15 @@ func main() {
...
-			Name:  "announce-addr",
-			Usage: "specify multiaddrs that this node can be connected to on",
+			Name: "announce-addr",
+			Usage: "specify multiaddrs that this node can be connected to	",
 			Value: cli.NewStringSlice(cfg.Node.AnnounceAddrs...),
 		},
+		&cli.StringFlag{
+			Name:  "peering-peers",
+			Usage: "specify peering peers that this node can be connected to",
+			Value: cfg.Node.GetPeeringPeersStr(),
```
en0ma, line 296: you did not add the ENV `EnvVars: []string{"PEERING_PEERS"},`

handlers.go
```diff
@@ -432,6 +443,130 @@ func (s *Server) handleStats(c echo.Context, u *User) error {
...
+// handlePeeringPeersRemove godoc
+// @Summary      Remove peers on Peering Service
+// @Description  This endpoint can be used to remove a Peer from the Peering Service
+func (s *Server) handlePeeringPeersRemove(c echo.Context) error {
+	var params []peer.ID
+
+	type PeeringPeerRemoveMessage struct {
+		Message     string    `json: Message`
+		PeersRemove []peer.ID `json: Peers`
+	}
```
gmelodie, line 489: nit: put this on `util/peering.go`

handlers.go
```diff
@@ -432,6 +443,130 @@ func (s *Server) handleStats(c echo.Context, u *User) error {
...
+			Code:    400,
+			Message: util.ERR_PEERING_PEERS_REMOVE_ERROR,
+		}
+	}
+
+	for _, peerId := range params {
+		s.Node.Peering.RemovePeer(peerId)
+	}
+
+	return c.JSON(200, PeeringPeerRemoveMessage{"Removed the following Peers from Peering", params})
```
gmelodie, line 502: Use `http.StatusOK` or, better yet, `http.StatusAccepted` instead of `200`. Also do this to all the other returns on the endpoints

handlers.go
```diff
@@ -432,6 +443,130 @@ func (s *Server) handleStats(c echo.Context, u *User) error {
...
+		for _, addrInfo := range peerAddrInfo.Addrs {
+			peerAddrInfoAddrsStr = append(peerAddrInfoAddrsStr, addrInfo.String())
+		}
+		connectionCheck = append(connectionCheck, peering.PeeringPeer{
+			ID:        peerAddrInfo.ID.Pretty(),
+			Addrs:     peerAddrInfoAddrsStr,
+			Connected: (s.Node.Host.Network().Connectedness(peerAddrInfo.ID) == network.Connected),
+		})
+	}
+	return c.JSON(200, connectionCheck)
```
gmelodie, line 522: Oh I love what you did here!!

handlers.go
```diff
@@ -432,6 +443,130 @@ func (s *Server) handleStats(c echo.Context, u *User) error {
...
+}
+
+// handlePeeringStart godoc
+// @Summary      Start Peering
+// @Description  This endpoint can be used to start the Peering Service
+func (s *Server) handlePeeringStart(c echo.Context) error {
+	type GenericResponse struct {
+		Message string `json: "Message"`
+	}
```
gmelodie, line 531: This is declared twice (here and on `handlePeeringStop`). Maybe leave it on the top of the file

handlers.go
```diff
@@ -432,6 +443,130 @@ func (s *Server) handleStats(c echo.Context, u *User) error {
...
+
+// handlePeeringStop godoc
+// @Summary      Stop Peering
+// @Description  This endpoint can be used to stop the Peering Service
+func (s *Server) handlePeeringStop(c echo.Context) error {
+	type GenericResponse struct {
+		Message string `json: "Message"`
+	}
+	err := s.Node.Peering.Stop()
+	if err != nil {
```
gmelodie, line 551: We should log this error just as in `handlePeeringStart`.

node/modules/peering/peering.go
```diff
@@ -0,0 +1,45 @@
...
+	return ps.PeeringService.ListPeers()
+}
+
+func (ps *EstuaryPeeringService) AddPeer(info peer.AddrInfo) {
+	ps.PeeringService.AddPeer(info)
+}
+
+func (ps *EstuaryPeeringService) RemovePeer(peerId peer.ID) {
+	ps.PeeringService.RemovePeer(peerId)
+}
```
gmelodie, line 45: I'm not sure why we need to wrap everything up under the name of `EstuaryPeeringService`. Can't we simply use `ps.PeeringService` altogether?

node/node.go
```diff
@@ -168,6 +175,29 @@ func Setup(ctx context.Context, init NodeInitializer) (*Node, error) {
...
+			return nil, fmt.Errorf("failed to parse peering peers multi addr: %w", err)
+		}
+		addrInfoId, err := peer.Decode(addrInfo.ID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse peering peers multi addr ID: %w", err)
+		}
+		hPeer.AddPeer(peer.AddrInfo{ID: addrInfoId, Addrs: addrs})
+	}
+
+	//	We only want to start if there are peering_peers configured
```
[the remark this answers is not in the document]
gmelodie: Maybe also add this:
```
err := hPeer.Start()
if err != nil {
    log.Warnf(err)
}

```

node/node.go
```diff
@@ -270,9 +300,27 @@ func Setup(ctx context.Context, init NodeInitializer) (*Node, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse string multi addr: %w", err)
+	}
+	return a, nil
+}
+
+func toMultiAddresses(addrs []string) ([]multiaddr.Multiaddr, error) {
+	var multiAddrs []multiaddr.Multiaddr
+	for _, addr := range addrs {
+		a, _ := toMultiAddress(addr)
```
gmelodie, line 318: Check the error here

gmelodie (dismissed):

handlers.go
en0ma: I meant the complete peer sent for the request - if there are multiple peers, 1 fails and 1 pass - it adds 1, yet returns an error for the request. So my question is, is that intended?

node/modules/peering/peering.go
alvin-reyes: We should always create a wrapper if we want to use an existing functionality so we can extend or add more processing to it. If we decide to log all peers in a DB or make any events, we can just extend the wrapper.

node/node.go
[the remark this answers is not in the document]
alvin-reyes: @en0ma I initially did it this way but I think we should still start the peering regardless if there are pre-defined peers or not.

handlers.go
alvin-reyes: I'll remove instead.

handlers.go
alvin-reyes: https://github.com/ipfs/go-ipfs/pull/9030

node/node.go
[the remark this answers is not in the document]
en0ma: @alvin-reyes I only made the suggestion because of your comment `//	We only want to start if there are peering_peers configured` - ignore my suggestion and remove the comment

main.go
alvin-reyes: I followed the convention of the announce-addr. But I get your point of just using "peering-peers".

handlers.go
```diff
@@ -269,6 +271,15 @@ func (s *Server) ServeAPI() error {
 	admin.POST("/cm/transfer/restart/:chanid", s.handleTransferRestart)
 	admin.POST("/cm/repinall/:shuttle", s.handleShuttleRepinAll)
 
+	//	peering
+	adminPeering := admin.Group("/peering")
+	adminPeering.POST("/peers/add", s.handlePeeringPeersAdd)
+	adminPeering.POST("/peers/remove", s.handlePeeringPeersRemove)
```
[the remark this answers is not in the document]
alvin-reyes: Yes. I have this change already.

handlers.go
alvin-reyes: yeah initially I was thinking of creating 2 separate models for these 2 types of response but I think it make sense to just use a generic one. I'll put this at the top then. Thank you.

node/modules/peering/peering.go
alvin-reyes: We want to make sure that we are loosely coupled from the implementation so that we can extend the functionality. I wrapped it so that can add more features and `decorate` this later if needed.

handlers.go
alvin-reyes: Ok that's a good suggestion. I was following the same response from other endpoints.

main.go
```diff
@@ -292,6 +304,11 @@ func main() {
...
+			Usage:   "peering addresses for the libp2p server to listen on",
+			EnvVars: []string{"ESTUARY_PEERING_PEERS"},
```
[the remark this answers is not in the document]
alvin-reyes: I'd go with the ESTUARY_* prefix since that seems to be the convention according to the other flags.

handlers.go
```diff
@@ -269,6 +276,15 @@ func (s *Server) ServeAPI() error {
 	admin.POST("/cm/transfer/restart/:chanid", s.handleTransferRestart)
 	admin.POST("/cm/repinall/:shuttle", s.handleShuttleRepinAll)
 
+	//	peering
+	adminPeering := admin.Group("/peering")
+	adminPeering.POST("/peers/add", s.handlePeeringPeersAdd)
+	adminPeering.DELETE("/peers/remove", s.handlePeeringPeersRemove)
+	adminPeering.GET("/peers/list", s.handlePeeringPeersList)
```
en0ma, line 283: @alvin-reyes not sure there is a need for the `peer` in the route, same with adding `add`, `remove` and `list` in the route 
POST /peering - is enough for add
GET /Perring - is enough for list
DELETE /peering -  enough for delete

handlers.go
alvin-reyes: This is approved so I'll wait for this to get merged and grab the latest version of IPFS to use the new State() function.

handlers.go
alvin-reyes: Given how `peering/peers` are structured in ipfs, IMO it make sense to have `/peers` because it directly calls the object we are operating in e.i `peers`/

I modified the implementation to the following instead: 

- POST /peering/peers - for adding a list of peers
- DELETE /peering/peers - to remove a peer
- GET /peering/peers - list of peers

And these are the `Peering` level endpoints.

- POST /peering/stop 
- POST /peering/start
- GET /peering/status

handlers.go
alvin-reyes: if one of the address is invalid, then the entire request should be invalid. It would be best for us to impose an absolute correctness of the addresses and inform the consumer if one of them is invalid.

handlers.go
alvin-reyes: done. I think this should be in a model package rather than util. We can revisit once we start `refactoring` the codebase.

node/modules/peering/peering.go
alvin-reyes: I thought about this. I initially think we should at least prepare this if we want estuary to monitor incoming connection but since we don't have anything concrete plans yet, we can remove.

handlers.go
en0ma: @alvin-reyes on line 479, it is not all or nothing - which is what I am pointing at

handlers.go
en0ma: when will this be updated

handlers.go
alvin-reyes: Yes. That's why we return a json response immediately if one of the addrs is invalid.

Line 466 - 475
```
for _, addr := range peerParam.Addrs {
			a, err := multiaddr.NewMultiaddr(addr)
			if err != nil {
				log.Errorf("handlePeeringPeersAdd error: %s", err)
				return c.JSON(http.StatusBadRequest,
					util.PeeringPeerAddMessage{
						"Adding Peer(s) on Peering failed, the addr is invalid" + addr, params})
			}
			multiAddrs = append(multiAddrs, a)
		}
```

If one of them is invalid, the entire request to add will be cancelled with StatusBadRequest response. It'll also show the message and list of invalid addrs.

image

en0ma: you are looking at per addr level. I am talking about per peer level

alvin-reyes: If one of the peers are invalid, then we reject the entire request and raise the error similar to how we handle peer addrs level.

image

It make sense to reject the entire request if one of the peers or peer.addrs are invalid.

handlers.go
alvin-reyes: No dates yet but should be soon, I can wait for the next release and update this PR once we have the new release published.

en0ma (approved):
LGTM

PR #281, opened 2022-06-18
Added docker images to the release pipeline

jlogelin:


alvin-reyes (approved):
LGTM. Just need to take note that we would need to change the `tags` if we eventually decide to create our own dockerhub account.

jlogelin:
> LGTM. Just need to take note that we would need to change the `tags` if we eventually decide to create our own dockerhub account.

That's a good point, maybe I'll factor this out into variable.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,41 @@
...
+          echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+
+      - name: Build and push
+        env:
+          VERSION: ${{ steps.get_version.outputs.VERSION }}
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: true
+          tags: ${{ secrets.DOCKERHUB_ORG }}/estuary:latest , ${{ secrets.DOCKERHUB_ORG }}/estuary:${{ steps.get_version.outputs.VERSION }}
```
alvin-reyes, line 41: LGTM. 🚀

PR #282, opened 2022-06-18
Added client sdk generation script

jlogelin:
Added OpenApi client SDK generation script which can be used by customers to create ad-hoc sdks or via our build pipelines to generate version tracked SDKs.


en0ma (approved):
LGTM

Issue #275, opened 2022-06-16
Improve Estuary pinning service to meet compliance


en0ma:
https://www.notion.so/pl-strflt/202206-Pinning-Service-Compliance-Checker-Initial-Release-b9cdb7ddaa85401e84771fb6957f53bf

en0ma:
@corinne-antonia I am currently working on this

corinne-antonia:
@en0ma Ok great! Thank you

SgtPooki:
Can I use this repo to open any future compliance check failures? w.r.t https://github.com/ipfs-shipyard/pinning-service-compliance/issues/125 ?

alvin-reyes:
> Can I use this repo to open any future compliance check failures? w.r.t [ipfs-shipyard/pinning-service-compliance#125](https://github.com/ipfs-shipyard/pinning-service-compliance/issues/125) ?

Yes of course.

PR #290, opened 2022-06-23
Refactor/improve pinning service for compliance

en0ma:
This PR does a refactor of the pinning implementation to make sure it is compliant with the pinning [spec](https://www.notion.so/pl-strflt/202206-Pinning-Service-Compliance-Checker-Initial-Release-b9cdb7ddaa85401e84771fb6957f53bf) 

NB: I have done the manual testing as well using the compliance audit [tool](https://github.com/ipfs-shipyard/pinning-service-compliance)

closes https://github.com/application-research/estuary/issues/275

### DIFF
--- a/pinning.go
+++ b/pinning.go
@@ -83,7 +83,7 @@ func (cm *ContentManager) pinDelegatesForContent(cont Content) []string {
 		}
 
 		if ai == nil {
-			log.Warnf("no address info for shuttle %s: %s", cont.Location, err)
+			log.Warnf("no address info for shuttle: %s", cont.Location)
 			return nil
 		}
 
@@ -107,7 +107,6 @@ func (s *Server) doPinning(ctx context.Context, op *pinner.PinningOperation, cb 
 
 	bserv := blockservice.New(s.Node.Blockstore, s.Node.Bitswap)
 	dserv := merkledag.NewDAGService(bserv)
-
 	dsess := merkledag.NewSession(ctx, dserv)
 
 	if err := s.CM.addDatabaseTrackingToContent(ctx, op.ContId, dsess, op.Obj, cb); err != nil {
@@ -133,7 +132,6 @@ func (s *Server) doPinning(ctx context.Context, op *pinner.PinningOperation, cb 
 	if err := s.Node.Provider.Provide(op.Obj); err != nil {
 		log.Warnf("providing failed: %s", err)
 	}
-
 	return nil
 }
 
@@ -161,7 +159,6 @@ func (cm *ContentManager) refreshPinQueue() error {
 			}
 		}
 	}
-
 	return nil
 }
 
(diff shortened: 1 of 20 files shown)


jlogelin:
This is awesome. Maybe not for this PR, but do you think it would be possible to include an adhoc Pinning Service Compliance check that we could put into the pipeline?

en0ma:
> This is awesome. Maybe not for this PR, but do you think it would be possible to include an adhoc Pinning Service Compliance check that we could put into the pipeline?

If we are to implement this, It would require cloning the compliance [repo](https://github.com/ipfs-shipyard/pinning-service-compliance) and running the audit against a running estuary server on the pipeline.

Also, the compliance team intends to add a pipeline where, as they improve the requirements and checks, it runs against pinning services prod API and notifies (creates an issue) on the service repo.

pinning.go
```diff
@@ -656,8 +672,8 @@ func (s *Server) handleAddPin(e echo.Context, u *User) error {
 
 	if s.CM.contentAddingDisabled || u.StorageDisabled {
 		return &util.HttpError{
-			Code:    http.StatusBadRequest,
-			Message: util.ERR_CONTENT_ADDING_DISABLED,
+			Code:   http.StatusBadRequest,
+			Reason: util.ERR_CONTENT_ADDING_DISABLED,
```
alvin-reyes, line 676: I'm not sure if we need to change the `Message` field. We can add both `Message` and `Reason`.

pinning.go
```diff
@@ -152,52 +170,65 @@ func (cm *ContentManager) refreshPinQueue() error {
 	// anyways
 	makeDeal := true
 	for _, c := range toPin {
+		var origins []*peer.AddrInfo
+		// when refreshing pinning queue, use content origins if available
+		if c.Origins != "" {
+			_ = json.Unmarshal([]byte(c.Origins), &origins)
```
alvin-reyes, line 176: Handle here
```
if err := json.Unmarshal([]byte(c.Origins), &origins)); err != nil {
        log.Warnf("Content origins is invalid: %s", err)
}
```

en0ma: No need to handle or log the error - it is not actionable, using the origin here is a nice to have. I have added a comment to state why we don't need to

pinning.go
en0ma: Actually, `Reason` is better for the internal error code in `util.XXXXXX_XXXX`. The `Message` field has been replaced with the `details` field - with has more message context. I just updated the error to add details.

alvin-reyes (approved):
LGTM

SgtPooki:
> This is awesome. Maybe not for this PR, but do you think it would be possible to include an adhoc Pinning Service Compliance check that we could put into the pipeline?

This is a great feature request. I opened https://github.com/ipfs-shipyard/pinning-service-compliance/issues/140 to track it

PR #294, opened 2022-06-28
fix: always close request body

en0ma:
Make sure to close the request body before and after content size check


alvin-reyes (approved):
LGTM.

PR #298, opened 2022-06-28
chore: fix the username taken validation

alvin-reyes:
Fix to validate the user pointer AND the actual username upon registration

# Verification

## Test data. Only 3 users available on my DB.
image

## Add new user
Add `alvinreyes2`
image

Added `alvinreyes2` successfully with the invite code
image

Tried adding `alvinreyes2` again
image

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -2880,7 +2880,7 @@ func (s *Server) handleRegisterUser(c echo.Context) error {
 		}
 	}
 
-	if exist != nil {
+	if exist != nil && strings.ToLower(exist.Username) == username {
 		return &util.HttpError{
 			Code:   http.StatusBadRequest,
 			Reason: util.ERR_USERNAME_TAKEN,
--- a/replication.go
+++ b/replication.go
@@ -1588,6 +1588,10 @@ func (cm *ContentManager) checkDeal(ctx context.Context, d *contentDeal) (int, e
 		return DEAL_CHECK_PROGRESS, nil
 	}
 
+	if provds == nil {
+		return DEAL_CHECK_UNKNOWN, fmt.Errorf("failed to lookup provider deal state")
+	}
+
 	if provds.State == storagemarket.StorageDealError {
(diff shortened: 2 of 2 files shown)


jlogelin:
It looks like shuttle-4 is down [causing the tests to fail](https://github.com/application-research/estuary/runs/7103989459?check_suite_focus=true#step:8:16)

handlers.go
```diff
@@ -2880,7 +2880,7 @@ func (s *Server) handleRegisterUser(c echo.Context) error {
 		}
 	}
 
-	if exist != nil {
+	if exist != nil && strings.ToLower(exist.Username) == username {
```
en0ma, line 2883: @alvin-reyes why is this necessary?

alvin-reyes: The `exists` here will always be not nil because it's pointer, e.i it will always hold the reference and even if it doesn't get any record, it will initialize a new instance of the same object.

I added a check to validate if the current record `Username` is equivalent to given `username`.

replication.go
```diff
@@ -1605,12 +1609,12 @@ func (cm *ContentManager) checkDeal(ctx context.Context, d *contentDeal) (int, e
...
 	if provds.DealID != 0 {
 		deal, err := cm.Api.StateMarketStorageDeal(ctx, provds.DealID, types.EmptyTSK)
-		if err != nil {
+		if err != nil || deal == nil {
 			return DEAL_CHECK_UNKNOWN, fmt.Errorf("failed to lookup deal on chain: %w", err)
 		}
 
 		pcr, err := cm.lookupPieceCommRecord(content.Cid.CID)
-		if err != nil {
+		if err != nil || pcr == nil {
```
alvin-reyes, line 1617: I added the check to `pcr` here cause the `lookupPieceCommRecord` can return `nil` pcr record which will cause the dereferencing exception.

handlers.go
en0ma: Oh shit, I see 😞

en0ma (approved):

PR #302, opened 2022-06-29
revert car upload content size validation

en0ma:
I suspect the change to validate the car body is not closing the resource properly and I have to really look into it. Lets revert it for now.


alvin-reyes (approved):
LGTM. We'll revisit this later.

PR #304, opened 2022-06-29
Mark failed deal data transfer as failed and stop restarting it 

en0ma:
Mark failed deal data transfer as `failed` so it won't be picked up perpetually for retry.

closes https://github.com/application-research/estuary/issues/305

### DIFF
--- a/main.go
+++ b/main.go
@@ -820,16 +820,15 @@ func (s *Server) RestartAllTransfersForLocation(ctx context.Context, loc string)
 			continue
 		}
 
-		if err := s.CM.RestartTransfer(ctx, loc, chid); err != nil {
+		if err := s.CM.RestartTransfer(ctx, loc, chid, d.ID); err != nil {
 			log.Errorf("failed to restart transfer: %s", err)
 			continue
 		}
 	}
-
 	return nil
 }
 
-func (cm *ContentManager) RestartTransfer(ctx context.Context, loc string, chanid datatransfer.ChannelID) error {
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -820,29 +820,32 @@ func (s *Server) RestartAllTransfersForLocation(ctx context.Context, loc string)
...
+func (cm *ContentManager) RestartTransfer(ctx context.Context, loc string, chanid datatransfer.ChannelID, dealID uint) error {
 	if loc == "local" {
 		st, err := cm.FilClient.TransferStatus(ctx, &chanid)
 		if err != nil {
 			return err
 		}
 
 		if util.TransferTerminated(st) {
-			return fmt.Errorf("deal in database as being in progress, but data transfer is terminated: %d", st.Status)
+			if err := cm.DB.Model(contentDeal{}).Where("id = ?", dealID).UpdateColumns(map[string]interface{}{
```
alvin-reyes, line 839: Do we need to store the status if we're going to restart the transfer anyway (line 847)? Just concern that we have a DB call here to update it to failed only to do another restart transfer which technically will be "in-progress".

en0ma: look closely, if `TransferTerminated` (meaning is it useless to restart), we mark it as `failed` and return the `deal in database is in progress, but data transfer is terminated` error (line 847 wont get called). The marking is useful in that, whenever estuary is restarted or restart is trigger, deals marked as failed will not be picked from db - improved db calls and less error logs

alvin-reyes (approved):
LGTM

PR #309, opened 2022-07-05
chore: added default peers based on content provider list

alvin-reyes:
# Changes

Added all default content provider list based here https://docs.ipfs.io/how-to/peering-with-content-providers/#content-provider-list

# Verification
image
image

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -57,9 +57,55 @@ var bootstrappers = []string{
 	"/dnsaddr/bootstrap.libp2p.io/p2p/QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt",
 }
 
-var bootStrapPeeringPeers = []peering.PeeringPeer{
+var peeringPeers = []peering.PeeringPeer{
+	//Cloudflare
+	{ID: "QmcfgsJsMtx6qJb74akCw1M24X1zFwgGo11h1cuhwQjtJP", Addrs: []string{"/ip6/2606:4700:60::6/tcp/4009", "/ip4/172.65.0.13/tcp/4009"}},
+
+	//	NFT storage
 	{ID: "12D3KooWEGeZ19Q79NdzS6CJBoCwFZwujqi5hoK8BtRcLa48fJdu", Addrs: []string{"/ip4/145.40.96.233/tcp/4001"}},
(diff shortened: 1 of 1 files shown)


en0ma:
@alvin-reyes I thought these were added as cli config values when standing estuary

en0ma:
> > @alvin-reyes I thought these were added as cli config values when standing estuary
> 
> Not all. We only add a few from the list.
> 
> This PR will force estuary to peer with these providers upon startup.

ok, so we are treating these as a kinda bootstrap peers and any others via the cli are additions

en0ma:
> > > > @alvin-reyes I thought these were added as cli config values when standing estuary
> > > 
> > > 
> > > Not all. We only add a few from the list.
> > > This PR will force estuary to peer with these providers upon startup.
> > 
> > 
> > ok, so we are treating these as a kinda bootstrap peers and any others via the cli are additions
> 
> Yes. This is sort of an official list that's recognized by IPFS ecosystem https://docs.ipfs.io/how-to/peering-with-content-providers/#content-provider-list
> 
> I'm thinking of adding a json file on [ipfs-docs](https://github.com/ipfs/ipfs-docs) repo to make it easy for anyone to consume the peers from an external source.

Hmm, that will be a hard dependency on the availability of that endpoint - not sure that is a good idea

en0ma (approved):
LGTM

alvin-reyes:
> > > > > @alvin-reyes I thought these were added as cli config values when standing estuary
> > > > 
> > > > 
> > > > Not all. We only add a few from the list.
> > > > This PR will force estuary to peer with these providers upon startup.
> > > 
> > > 
> > > ok, so we are treating these as a kinda bootstrap peers and any others via the cli are additions
> > 
> > 
> > Yes. This is sort of an official list that's recognized by IPFS ecosystem https://docs.ipfs.io/how-to/peering-with-content-providers/#content-provider-list
> > I'm thinking of adding a json file on [ipfs-docs](https://github.com/ipfs/ipfs-docs) repo to make it easy for anyone to consume the peers from an external source.
> 
> Hmm, that will be a hard dependency on the availability of that endpoint - not sure that is a good idea

The idea is to make this json file part of the documentation. not sure if ipfs-doc team will agree but it’ll be good to just source this from a external list rather than hard coding it in our code.

Issue #310, opened 2022-07-05
Properly implement Refresh Pin Queue


en0ma:
**Describe the bug**
When Estuary starts, it tries to refresh the pin queue for local and shuttle contents. While local refresh works fines,  shuttle refresh is currently implemented in a hackish way - It just waits for a bit and expects the shuttles to be online and then tries to refresh pins on those shuttles (without any guarantee that they are online).

**Expected behavior**
Estuary should be more intelligent to detect when a shuttle is online and should refresh the pin queue for that shuttle content. Also, when the shuttle is down, it should gracefully stop trying to pin content on that shuttle.

PR #311, opened 2022-07-05
Properly implement refreshPinQueue on node startup

en0ma:
This PR improves the implementation of `refreshPinQueue` to make sure it works for local and shuttle deterministically.

Closes https://github.com/application-research/estuary/issues/310


alvin-reyes (approved):
This is nice. It'll have some level of assurance that the shuttles will try to refresh all `queued | pinning` content. 

LGTM.

PR #175, opened 2022-05-02
Redirect when found preferred upload endpoint on /content/add

gmelodie:
Adds the functionality requested in #153

PR #315, opened 2022-07-09
Replace shuttle-proxy with redirects

gmelodie:
In a discussion with @jlogelin we realized that what `shuttle-proxy` did was not very different from what I implemented in https://github.com/application-research/estuary/pull/175. I personally think that, at least right now, shuttle-proxy adds an unnecessarily big volume of code, so I just made the main node do what shuttle-proxy did: 

When a client hits `/content/add` or `/content/add-car` and `--disable-content-adding` is set on the main node, it fetches available endpoints (other shuttles) and makes the call to them.

Obs: this PR also fixes the biggest issue with https://github.com/application-research/estuary/pull/175: users don't need to worry about redirecting now, since the main node just does everything for them transparently.


alvin-reyes:
Instead of adding it to the main node, we should create a standalone shuttle-proxy that clients can use for their dedicated/self-hosted infra. 

I'm thinking we should:
1 - decouple this "code" as a "proxifier" library
2 - create a standalone shuttle-proxy node with this library
3 - use the same proxifier library when a main and shuttle node sets a flag (like: "enable-proxy"). 

image

gmelodie:
I would agree with that approach if we had to do complicated load balancing, but that's not the case with estuary. It just feels like this is overengineering things, we just need a simple redirect to fix this issue.

alvin-reyes (approved):
LGTM. We should revisit this later on but we can merge it for now.

PR #318, opened 2022-07-13
chore: small change to add total users and sp on public/stats endpoint

alvin-reyes:
# Changes
Small change to add total users and total SPs on the public/stat endpoint. 

# Verification
image

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -3687,6 +3687,8 @@ type publicStatsResponse struct {
 	DealsOnChain       int64 `json:"dealsOnChain"`
 	TotalObjectsRef    int64 `json:"totalObjectsRef"`
 	TotalBytesUploaded int64 `json:"totalBytesUploaded"`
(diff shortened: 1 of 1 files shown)


handlers.go
```diff
@@ -3687,6 +3687,8 @@ type publicStatsResponse struct {
 	DealsOnChain       int64 `json:"dealsOnChain"`
 	TotalObjectsRef    int64 `json:"totalObjectsRef"`
 	TotalBytesUploaded int64 `json:"totalBytesUploaded"`
+	TotalUsers         int64 `json:"totalUsers"`
```
en0ma, line 3690: @alvin-reyes please use `sql.NullInt64` for all of these - it is safer if there are null values

alvin-reyes: done on the latest commit.

en0ma (approved):
LGTM

Issue #185, opened 2022-05-06
Levels for certain loggers are hard-coded


softwareplumber:
Now that we have a config file, I feel it would be appropriate to move the default log configuration into the config file instead of main.go.

This would change behavior slightly because you'd have to run 'estuary configure' to create a config file in order to restore the 'as-is' logging behavior.

corinne-antonia:
@whyrusleeping @alvin-reyes What do you guys think? Should we prioritize this sooner than later?

PR #321, opened 2022-07-13
feat: add support for log-level cli flag

en0ma:
Add support for `log-level` cli flag. Defaults to `info`

`./estuary --datadir=`

Screenshot 2022-07-14 at 18 04 24

`./estuary --log-levl=debug --datadir=`
Screenshot 2022-07-14 at 18 05 04

closes https://github.com/application-research/estuary/issues/185

### DIFF
--- a/main.go
+++ b/main.go
@@ -159,6 +159,27 @@ func (s *Server) updateAutoretrieveIndex(tickInterval time.Duration, quit chan s
 	}
 }
 
+func before(cctx *cli.Context) error {
+	level := "INFO"
+	if util.IsVeryVerbose {
+		level = "DEBUG"
+	}
+
+	_ = logging.SetLogLevel("dt-impl", level)
+	_ = logging.SetLogLevel("estuary", level)
+	_ = logging.SetLogLevel("paych", level)
+	_ = logging.SetLogLevel("filclient", level)
+	_ = logging.SetLogLevel("dt_graphsync", level)
+	_ = logging.SetLogLevel("dt-chanmon", level)
+	_ = logging.SetLogLevel("markets", level)
+	_ = logging.SetLogLevel("data_transfer_network", level)
+	_ = logging.SetLogLevel("rpc", level)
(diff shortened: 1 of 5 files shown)


util/cli.go
```diff
@@ -0,0 +1,12 @@
+package util
+
+import cli "github.com/urfave/cli/v2"
+
+var IsVeryVerbose bool
```
alvin-reyes, line 5: We should just use `debug` instead of `VeryVerbose`. `debug` implies that the configuration is for debugging purposes which is what you are intending to do here.

gmelodie: I agree. I think there's generally two models CLIs use: `-v/-vv/-vvv` (nmap, curl) or `--log-level` (golang things) . Maybe it makes sense to change `--logging` to `--log-level INFO` (maybe make this the default) and just add `--log-level DEBUG`, etc.

alvin-reyes: `--log-level DEBUG/INFO` makes much more sense. Let's go with this.

en0ma: noted

util/cli.go
```diff
@@ -0,0 +1,12 @@
+package util
+
+import cli "github.com/urfave/cli/v2"
+
+var LogLevl string
```
alvin-reyes, line 5: Please change it to `LogLevel`.

util/cli.go
```diff
@@ -0,0 +1,12 @@
...
+
+var FlagLogLevl = &cli.StringFlag{
```
alvin-reyes, line 7: Change to `FlagLogLevel`

main.go
```diff
@@ -159,6 +159,24 @@ func (s *Server) updateAutoretrieveIndex(tickInterval time.Duration, quit chan s
 	}
 }
 
+func before(cctx *cli.Context) error {
+	level := util.LogLevl
```
alvin-reyes, line 163: Change to `LogLevel`

alvin-reyes (approved):
LGTM

PR #322, opened 2022-07-15
Only make shuttle content deals if shuttle is online

en0ma:
Currently, when making deals for shuttle contents, there are no checks to make sure the shuttle for such contents is online. This leads to errors such as `shuttle is not online: SHUTTLE983b494d-a86d-49ca-9630-21a902349e41HANDLE` when a deal proposal is made. This PR makes sure the checks are made before proceeding to make proposals.


alvin-reyes (approved):
LGTM

PR #329, opened 2022-07-19
change name variables to filename

### LINKED ISSUE #262: Change "name" field to "filename" in the Content struct
Mostly a self reminder since I don't have bandwidth to do this just about now. Just rename the field and fix everywhere it breaks.

gmelodie:
Fixes #262


alvin-reyes (commented):
Thanks for this. Please add a few verification steps on this PR.

gmelodie:
@alvin-reyes can you clarify what you mean with verification steps? Do you mean unit tests? If so, what should I unit test specifically?

alvin-reyes:
> @alvin-reyes can you clarify what you mean with verification steps? Do you mean unit tests? If so, what should I unit test specifically?

Just test the endpoints and show that its accepting the new name and returns (response) the new name.

one example is the `/content/create` that accepts a `var req util.ContentCreateBody` which includes the `filename`. This technically means anyone who uses the `/content/create` needs to use `filename` instead of `name` which makes this a breaking API change.

image

image

We need to make sure verify if this work as is (add it to this PR) and also modify the docs with this new param.

alvin-reyes:
# Verification
## Name column on the content table
image

## Run /content/create
image

image

en0ma:
@alvin-reyes @gmelodie those this change require any UI updates?

gmelodie:
@en0ma the API changes are on `/content/add` (it returns the `Content`, which now uses `Filename` instead of `Name`), `createContent` and `shuttleCreateContent`. I am not 100% sure how these last two are used

alvin-reyes:
> @alvin-reyes @gmelodie those this change require any UI updates?

Yes. I'll have a PR that will have this change on `estuary-www` shortly.

alvin-reyes (dismissed):
LGTM

alvin-reyes (approved):
LGTM

PR #337, opened 2022-07-19
chore: added estuary peering peers

alvin-reyes:
# Change

Add estuary peers on the peering list

# Verification (`/admin/peering/peers`)

image


gmelodie (approved):
@alvin-reyes the PR looks solid. Just one question though: what are these peers being added for? So that every estuary node can peer with our hosted main node + shuttles? If so, what benefit does this bring?

alvin-reyes:
Estuary has a `/gw` (gateway) to can be used to retrieve content from other peers. If we have a CID pinned on a shuttle, the the gateway node (in this case the main estuary node) can locate the CID more efficiently.

PR #339, opened 2022-07-20
reduce duplicate code

anjor:


handlers.go
```diff
@@ -711,39 +707,11 @@ func (s *Server) handleAddIpfs(c echo.Context, u *User) error {
...
-				}
-			} else {
-				return &util.HttpError{
-					Code:    http.StatusBadRequest,
-					Reason:  util.ERR_CONTENT_ADDING_DISABLED,
-					Details: "uploading content to this node is not allowed at the moment",
-				}
-			}
-		}
+	if err := s.redirectIfLocalContentAddingDisabled(c, u); err != nil {
```
en0ma, line 710: you should call `ErrorIfContentAddingDisabled` before `redirectIfLocalContentAddingDisabled`

en0ma: also for `redirectIfLocalContentAddingDisabled`, if true, code should redirect and not proceed any further

en0ma: yes, it is a different behavior but that should have been the proper order.

handlers.go
```diff
@@ -1026,6 +968,40 @@ func (s *Server) handleAdd(c echo.Context, u *User) error {
...
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return err
+		}
+
+		c.Response().WriteHeader(resp.StatusCode)
+
+		_, err = io.Copy(c.Response().Writer, resp.Body)
+		if err != nil {
+			log.Errorf("proxying content-add body errored: %s", err)
```
en0ma, line 993: same here, we should return the error

en0ma (approved):
LGTM

PR #342, opened 2022-07-21
Fixes #338 : Salted plaintext passwords 

### LINKED ISSUE #338: Stop storing password in plain text
Currently, in the users' table, passwords are stored as plain text, this is should be the case

kelindi:
Fixes #338 

Added 🧂 to Users schema. Salting at registration and Change password. On Login plaintext is hashed and compared with hashed passwords in DB. This makes the login incompatible with the existing DB of old passwords without a proper migration.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -2852,7 +2852,7 @@ func (s *Server) AuthRequired(level int) echo.MiddlewareFunc {
 
 type registerBody struct {
 	Username     string `json:"username"`
-	PasswordHash string `json:"passwordHash"`
+	Password	 string `json:"passwordHash"`
 	InviteCode   string `json:"inviteCode"`
 }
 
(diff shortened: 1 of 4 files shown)


alvin-reyes (commented):
Looks good. Please add a few verification steps on this PR:

1 - test user registration (with password). provide screenshot of the salt column.
2 - test user password change

Thank you.

alvin-reyes:
LGTM.

image

alvin-reyes (approved):
LGTM

handlers.go
```diff
@@ -2958,7 +2961,9 @@ func (s *Server) handleLoginUser(c echo.Context) error {
 		return err
 	}
 
-	if user.PassHash != body.PassHash {
+	
+	
+	if user.PassHash != util.GetPasswordHash(body.Password, user.Salt) {
```
en0ma, line 2966: @kelindi @alvin-reyes we need a backward compatible check for accounts that are already in plaintext

PR #344, opened 2022-07-21
Change more log levels

en0ma:


handlers.go
```diff
@@ -1484,7 +1484,7 @@ func (s *Server) handleGetContentByCid(c echo.Context) error {
 		return err
 	}
 
-	var out []getContentResponse
+	out := make([]getContentResponse, 0)
```
en0ma, line 1487: added this change so when a cid is not estuary, it does not return null, but an empty array

Screenshot 2022-07-21 at 10 01 59

VS

PR #360, opened 2022-07-26
chore: backward compatibility of password check

alvin-reyes:
# Changes

Now that we have the addition of `salt` as part of the password, we need to be able to handle the old passwords without salts. This ensures that old passwords will still work as is and will be migrated to the new password hashing version when the users decides to change their password.

# Verification (note: verified on localhost)
image

## verified with old account password change
old account: alvinreyesold 
https://user-images.githubusercontent.com/4479171/181120397-b742dc1b-2548-4190-8951-c0dea1fa8c27.mp4

## verified with new account password change
new account: alvinreyes 
https://user-images.githubusercontent.com/4479171/181120580-d9e81194-00f3-431c-961f-b975153316ee.mp4

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -2850,9 +2850,9 @@ func (s *Server) AuthRequired(level int) echo.MiddlewareFunc {
 }
 
 type registerBody struct {
-	Username     string `json:"username"`
-	Password	 string `json:"passwordHash"`
-	InviteCode   string `json:"inviteCode"`
+	Username   string `json:"username"`
+	Password   string `json:"passwordHash"`
+	InviteCode string `json:"inviteCode"`
 }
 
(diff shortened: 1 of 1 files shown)


en0ma (approved):
LGTM

en0ma:
@alvin-reyes it would have been nice to progressively migrate plain password users as they log in, instead of waiting for them to change their password. Ideally, if we had a task runner, we are supposed to use it to migrate all these accounts.

alvin-reyes:
> @alvin-reyes it would have been nice to progressively migrate plain password users as they log in, instead of waiting for them to change their password. Ideally, if we had a task runner, we are supposed to use it to migrate all these accounts.

Ack, can create a quick script for that purpose.

PR #362, opened 2022-07-26
Fixing gosec issues and adding gosec on scan on master push and pull request

mvjq:
Hello, this PR solves all the [issues](https://github.com/application-research/estuary/runs/7095139639?check_suite_focus=true) related to gosec security scan.

The issues are:
- G404 (CWE-338): Use of weak random number generator.
- G108 (CWE-200): Profiling endpoint is automatically exposed on /debug/pprof. 
- G101 (CWE-798): Potential hardcoded credentials.
- G304 (CWE-22): Potential file inclusion via variable.
- G601 (CWE-118): Implicit memory aliasing in for loop.
- G301 (CWE-276): Expect directory permissions to be 0750 or less.
- G307 (CWE-703): Deferring unsafe method "Close" on type "*os.File".
- G104 (CWE-703): Errors unhandled.

Observations:
- G404 is a false positive because the rand number is between [0, len(uep)], there's no need to generate a 'secure' random number.

```
//#nosec G404 - crypto/rand it's not necessary for this use case
// propagate any query params
req, err := http.NewRequest("POST", fmt.Sprintf("%s/content/add", uep[rand.Intn(len(uep))]), c.Request().Body)
```

- G108 is okay because the endpoint is intended to be exposed.
```
//#nosec G108 - exposing the profiling endpoint is expected
	_ "net/http/pprof"
```

-  G101 is a false positive because gosec searches for a string literal with "AUTH" in it. Example:
```
//#nosec G101 -- This is a false positive
const (
	ERR_INVALID_TOKEN              = "ERR_INVALID_TOKEN"
	ERR_TOKEN_EXPIRED              = "ERR_TOKEN_EXPIRED"
```

- G304 I did: `f, err := os.Open(filepath.Clean(filename))`

- G601 I did this to fix it:
```
for i, pd := range params.Deals {
		pd := pd // <-- here
		nd, err := cborutil.AsIpld(&pd)
```

- G301 changed the permission 755 to 750 (let me know if this is a problem):
```
 	if err := os.MkdirAll(dir, 0750); err != nil {
```

- G307 is tricky because when we use defer on functions that return an error, it's ignored. So in a couple of places, I did this: 
```
defer func() {
		if errC := f.Close(); errC != nil {
			err = errC
		}
	}()
```
err is a named return value.
 
- the G104 warning shows in a lot of places, most are solvable by treating the error ( ` err = doSomething(); err != nil {} { return err }`). I put #nosec for this when setLogLevel shows up because its not common to treat log errors (let me know if this is ok). In some places, i used  `xerrors.Errorf` and `fmt.Errorf` as return because the enclosing lines did the same.
```
//#nosec G104 - it's not common to treat SetLogLevel error return
	logging.SetLogLevel(body.System, body.Level)
```

Because I'm adding the GitHub action for `gosec` on push to master/PR, I think it's better to put this scan before doing the push to the remote but I do not know where. Makefile? Which recipe (build)?

All feedback is welcome, thanks!

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	//#nosec G108 - exposing the profiling endpoint is expected
 	_ "net/http/pprof"
 	"os"
 	"path/filepath"
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -807,12 +807,14 @@ func (d *Shuttle) RunRpcConnection() error {
 	}
 }
 
-func (d *Shuttle) runRpc(conn *websocket.Conn) error {
+func (d *Shuttle) runRpc(conn *websocket.Conn) (err error) {
 	conn.MaxPayloadBytes = 128 << 20
 	log.Infof("connecting to primary estuary node")
-	if err := conn.Close(); err != nil {
-		return err
-	}
+	defer func() {
+		if errC := conn.Close(); errC != nil {
+			err = errC
+		}
+	}()
 
 	readDone := make(chan struct{})
 
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -82,6 +82,7 @@ const (
 	ColDir  = "dir"
 )
 
+//#nosec G104 - it's not common to treat SetLogLevel error return
 func before(cctx *cli.Context) error {
 	level := util.LogLevel
 
@@ -1132,6 +1133,7 @@ func (s *Shuttle) handleLogLevel(c echo.Context) error {
 		return err
 	}
 
+	//#nosec G104 - it's not common to treat SetLogLevel error return
 	logging.SetLogLevel(body.System, body.Level)
 
 	return c.JSON(http.StatusOK, map[string]interface{}{})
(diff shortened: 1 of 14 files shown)


cmd/estuary-shuttle/main.go
```diff
@@ -805,7 +810,9 @@ func (d *Shuttle) RunRpcConnection() error {
 func (d *Shuttle) runRpc(conn *websocket.Conn) error {
 	conn.MaxPayloadBytes = 128 << 20
 	log.Infof("connecting to primary estuary node")
-	defer conn.Close()
+	if err := conn.Close(); err != nil {
```
en0ma, line 813: handle in a defer func

handlers.go
```diff
@@ -967,6 +967,7 @@ func (s *Server) redirectContentAdding(c echo.Context, u *User) error {
 		log.Warnf("failed to get preferred upload endpoints: %s", err)
 		return err
 	} else if len(uep) > 0 {
+		//#nosec G404 - crypto/rand it's not necessary for this use case
```
en0ma, line 970: what is the issue here?

mvjq: Hello,

the issue here is that the gosec considers the rand.Intn() insecure for generating random numbers:

image

but in this line, rand.Intn() is used to get a number between [0, len(uep)] so there is no need for a truly secure random number.

cmd/estuary-shuttle/main.go
mvjq: did this: 
```go
	defer func() {
		if errC := conn.Close(); errC != nil {
			err = errC
		}
	}()

```

handlers.go
jlogelin: Agreed @mvjq - we should be using common sense with these flags, this is fine.

en0ma:
@mvjq please resolve conflicts so I can approve

mvjq:
Done @en0ma, thanks for you patience!

mvjq:
@en0ma done, thanks for your patience

en0ma (approved):
LGTM

jlogelin:
nice work @mvjq - appreciate your efforts here!

PR #371, opened 2022-07-28
Fix redirect on content adding

gmelodie:
#315 introduced had an issue that made impossible for the functionality to work. Basically, the function was returning a slightly different endpoint string and we ended up redirecting to an inexistent place (hence the 404s instead of 400s when we tried to upload to  `api.estuary.tech/content/add`).

### Verification
Before (with `--disable-local-content-adding=true`):
image

After (with `--disable-local-content-adding=true`):
Screenshot from 2022-07-29 12-25-21

The behavior with `--disable-content-adding=true` (not `local`) is the same, we always reject:
image

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -3105,7 +3105,7 @@ func (s *Server) handleGetViewer(c echo.Context, u *User) error {
 			DealDuration:          dealDuration,
 			MaxStagingWait:        maxStagingZoneLifetime,
 			FileStagingThreshold:  int64(individualDealThreshold),
-			ContentAddingDisabled: s.CM.contentAddingDisabled || u.StorageDisabled,
+			ContentAddingDisabled: s.CM.globalContentAddingDisabled || u.StorageDisabled,
 			DealMakingDisabled:    s.CM.dealMakingDisabled(),
 			UploadEndpoints:       uep,
 			Flags:                 u.Flags,
@@ -5359,5 +5359,5 @@ func (s *Server) getShuttleConfig(hostname string, authToken string) (interface{
 }
 
 func (s *Server) isContentAddingDisabled(u *User) bool {
-	return s.CM.contentAddingDisabled || u.StorageDisabled
+	return s.CM.globalContentAddingDisabled || u.StorageDisabled
 }
--- a/handlers.go
+++ b/handlers.go
@@ -968,7 +968,7 @@ func (s *Server) redirectContentAdding(c echo.Context, u *User) error {
 		return err
 	} else if len(uep) > 0 {
 		// propagate any query params
-		req, err := http.NewRequest("POST", fmt.Sprintf("%s/content/add", uep[rand.Intn(len(uep))]), c.Request().Body)
+		req, err := http.NewRequest("POST", uep[rand.Intn(len(uep))], c.Request().Body)
 		if err != nil {
 			return err
 		}
--- a/handlers.go
+++ b/handlers.go
@@ -3137,6 +3137,7 @@ func (s *Server) getPreferredUploadEndpoints(u *User) ([]string, error) {
 	var shuttles []Shuttle
 	for hnd, sh := range s.CM.shuttles {
 		if sh.hostname == "" {
+			log.Debugf("shuttle %+v has empty hostname", sh)
 			continue
 		}
 
@@ -3147,6 +3148,7 @@ func (s *Server) getPreferredUploadEndpoints(u *User) ([]string, error) {
 		}
 
 		if !shuttle.Open {
+			log.Debugf("shuttle %+v is not open, skipping", shuttle)
 			continue
 		}
 
--- a/handlers.go
+++ b/handlers.go
@@ -961,39 +961,40 @@ func (s *Server) handleAdd(c echo.Context, u *User) error {
 	})
 }
 
+// redirectContentAdding is called when localContentAddingDisabled is true
+// it finds available shuttles and adds the desired content in one of them
 func (s *Server) redirectContentAdding(c echo.Context, u *User) error {
 	uep, err := s.getPreferredUploadEndpoints(u)
 	if err != nil {
-		log.Warnf("failed to get preferred upload endpoints: %s", err)
-		return err
-	} else if len(uep) > 0 {
-		// propagate any query params
-		req, err := http.NewRequest("POST", uep[rand.Intn(len(uep))], c.Request().Body)
-		if err != nil {
-			return err
-		}
-		req.Header = c.Request().Header.Clone()
-		req.URL.RawQuery = c.Request().URL.Query().Encode()
-
-		resp, err := http.DefaultClient.Do(req)
-		if err != nil {
-			return err
-		}
-
-		c.Response().WriteHeader(resp.StatusCode)
-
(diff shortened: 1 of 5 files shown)


en0ma (approved):
LGTM

en0ma:
> I am not a big fan of us tacking on the /content/add to the shuttle endpoints. But that's a breaking change, and probably to be dealt with separately, if at all.

Me too, and I came to that conclusion as well. As is, car upload will require striping the path out. We could add a backward compatible host and users can opt-in if they want

handlers.go
```diff
@@ -962,40 +962,40 @@ func (s *Server) handleAdd(c echo.Context, u *User) error {
...
 		return &util.HttpError{
 			Code:    http.StatusBadRequest,
 			Reason:  util.ERR_CONTENT_ADDING_DISABLED,
 			Details: "uploading content to this node is not allowed at the moment",
 		}
 	}
+	// propagate any query params
+	req, err := http.NewRequest("POST", uep[rand.Intn(len(uep))], c.Request().Body)
+	if err != nil {
+		return fmt.Errorf("failed to get preferred upload endpoints: %s", err)
```
en0ma, line 982: fix error message

handlers.go
```diff
@@ -962,40 +962,40 @@ func (s *Server) handleAdd(c echo.Context, u *User) error {
...
+	}
+	req.Header = c.Request().Header.Clone()
+	req.URL.RawQuery = c.Request().URL.Query().Encode()
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to get preferred upload endpoints: %s", err)
```
en0ma, line 989: fix error message

handlers.go
```diff
@@ -962,40 +962,40 @@ func (s *Server) handleAdd(c echo.Context, u *User) error {
...
+	}
+
+	c.Response().WriteHeader(resp.StatusCode)
+
+	_, err = io.Copy(c.Response().Writer, resp.Body)
+	if err != nil {
+		return fmt.Errorf("failed to get preferred upload endpoints: %s", err)
```
en0ma, line 996: fixe error message

handlers.go
```diff
@@ -3052,7 +3052,7 @@ func (s *Server) handleGetViewer(c echo.Context, u *User) error {
 			DealDuration:          dealDuration,
 			MaxStagingWait:        maxStagingZoneLifetime,
 			FileStagingThreshold:  int64(individualDealThreshold),
-			ContentAddingDisabled: s.CM.contentAddingDisabled || u.StorageDisabled,
+			ContentAddingDisabled: s.CM.globalContentAddingDisabled || u.StorageDisabled,
```
en0ma, line 3055: this is correct for this case

handlers.go
```diff
@@ -5316,5 +5318,5 @@ func (s *Server) getShuttleConfig(hostname string, authToken string) (interface{
 }
 
 func (s *Server) isContentAddingDisabled(u *User) bool {
-	return s.CM.contentAddingDisabled || u.StorageDisabled
+	return s.CM.globalContentAddingDisabled || u.StorageDisabled
```
en0ma, line 5321: This is not correct. for this case, we are checking if content can be added to this box - so `s.CM.contentAddingDisabled || u.StorageDisabled` is the correct way

main.go
```diff
@@ -699,7 +699,7 @@ func main() {
 		go cm.handleShuttleMessages(cctx.Context, cfg.ShuttleMessageHandlers) // register workers/handlers to process shuttle rpc messages from a channel(queue)
 
 		// refresh pin queue for local contents
-		if !cm.contentAddingDisabled {
+		if !cm.globalContentAddingDisabled {
```
en0ma, line 702: this change is not correct - we do not want to bring in more content via bitswap if local adding is disable

replication.go
```diff
@@ -302,34 +302,34 @@ func NewContentManager(db *gorm.DB, api api.Gateway, fc *filclient.FilClient, tb
...
+		pinJobs:                     make(map[uint]*pinner.PinningOperation),
+		pinMgr:                      pinmgr,
+		remoteTransferStatus:        cache,
+		shuttles:                    make(map[string]*ShuttleConnection),
+		contentSizeLimit:            util.DefaultContentSizeLimit,
+		hostname:                    cfg.Hostname,
+		inflightCids:                make(map[cid.Cid]uint),
+		FailDealOnTransferFailure:   cfg.Deal.FailOnTransferFailure,
+		isDealMakingDisabled:        cfg.Deal.Disable,
+		globalContentAddingDisabled: cfg.Content.DisableGlobalAdding,
```
en0ma, line 326: thanks for making this change, I have been wanting to do it

handlers.go
gmelodie: > we are checking if the user can still upload content to any box
I think the flag `disable-content-adding` is supposed to include local, at least that's how the code seems to assume it (see https://github.com/application-research/estuary/blob/459a166316370e00aee6388e91501f5668a10d7a/handlers.go#L703-L708)

handlers.go
gmelodie: Same as above

handlers.go
en0ma: @gmelodie these are the possible outcomes

- if all 3 `s.CM.globalContentAddingDisabled`, `s.CM.localContentAddingDisabled` and `u.StorageDisabled` are `true` => `errorNoUpload`

-  if both `s.CM.globalContentAddingDisabled` and `s.CM.localContentAddingDisabled` are `true` and  `u.StorageDisabled` is `false`  => `errorNoUpload`

- if `s.CM.globalContentAddingDisabled` is `true` and `s.CM.localContentAddingDisabled` are `false` and `u.StorageDisabled` is `true` => `errorNoUpload`

- if `s.CM.globalContentAddingDisabled` is `false` and `s.CM.localContentAddingDisabled` are `true` and `u.StorageDisabled` is `true` => `errorNoUpload`

- if `s.CM.globalContentAddingDisabled` is `true` and `s.CM.localContentAddingDisabled` are `false` and `u.StorageDisabled` is `false` => `uploads to shuttle`

- if `s.CM.globalContentAddingDisabled` is `false` and `s.CM.localContentAddingDisabled` are `true` and `u.StorageDisabled` is `false` => `uploads to estuary`

en0ma: > is supposed to include local, at least that's how the code seems to assume it see:

That was why I said **any** box - local or shuttle

handlers.go
```diff
@@ -3052,7 +3053,7 @@ func (s *Server) handleGetViewer(c echo.Context, u *User) error {
 			DealDuration:          dealDuration,
 			MaxStagingWait:        maxStagingZoneLifetime,
 			FileStagingThreshold:  int64(individualDealThreshold),
-			ContentAddingDisabled: s.CM.contentAddingDisabled || u.StorageDisabled,
+			ContentAddingDisabled: s.CM.globalContentAddingDisabled || u.StorageDisabled,
```
en0ma, line 3056: @gmelodie you need to use `s.isContentAddingDisabled(u)` here

en0ma (approved):
LGTM

PR #378, opened 2022-07-29
fix pinning status for shuttle contents

en0ma:
Currently, when shuttles add content and fail to signal estuary about pin completion (maybe due to downtime), estuary will not check if this content is pinned. This leaves the content in a stale state on Estuary, but on the shuttle, it will be pinned and active.

This PR correctly sets the pin status on Estuary side when the content is created, and if any downtime occurs, when Estuary is back up, it will retry the pinning and be updated accordingly.

see here for more context https://filecoinproject.slack.com/archives/C033QS35MFA/p1659047543608269

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -1151,12 +1151,8 @@ func (s *Shuttle) handleLogLevel(c echo.Context) error {
 func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 	ctx := c.Request().Context()
 
-	if u.StorageDisabled || s.disableLocalAdding {
-		return &util.HttpError{
-			Code:    http.StatusBadRequest,
-			Reason:  util.ERR_CONTENT_ADDING_DISABLED,
-			Details: "uploading content to this node is not allowed at the moment",
-		}
+	if err := util.ErrorIfContentAddingDisabled(u.StorageDisabled || s.disableLocalAdding); err != nil {
+		return err
 	}
 
 	form, err := c.MultipartForm()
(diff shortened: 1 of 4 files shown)


cmd/estuary-shuttle/main.go
```diff
@@ -1149,12 +1149,8 @@ func (s *Shuttle) handleLogLevel(c echo.Context) error {
 func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 	ctx := c.Request().Context()
 
-	if u.StorageDisabled || s.disableLocalAdding {
-		return &util.HttpError{
-			Code:    http.StatusBadRequest,
-			Reason:  util.ERR_CONTENT_ADDING_DISABLED,
-			Details: "uploading content to this node is not allowed at the moment",
-		}
+	if err := util.ErrorIfContentAddingDisabled(u.StorageDisabled || s.disableLocalAdding); err != nil {
```
gmelodie, line 1152:
```suggestion
	if err := util.ErrorIfContentAddingDisabled(s.isContentAddingDisabled(u)); err != nil {
```

gmelodie (changes requested):

cmd/estuary-shuttle/main.go
en0ma: shuttle cant use `isContentAddingDisabled`

gmelodie: We could just add the function:

```
func (s *Shuttle) isContentAddingDisabled(u *User) bool {
	return s.disableLocalAdding || u.StorageDisabled
}
```

en0ma: thank you

gmelodie (approved):

PR #380, opened 2022-07-29
add default peering peers to config

en0ma:
- before
Screenshot 2022-07-29 at 22 52 33

- after
Screenshot 2022-07-29 at 22 50 15


alvin-reyes (approved):
LGTM

PR #385, opened 2022-08-03
Revert autoretrieve

en0ma:
For some reason, Autoretrieve changes prevent Graphsync data transfer from starting. @gmelodie and I were able to verify this.


alvin-reyes:
Question - do we want to remove the AR table along with this rollback?

alvin-reyes (changes requested):
Look into fixing it forward rather than rolling back.

en0ma:
> Question - do we want to remove the AR table along with this rollback?

I am not sure there is any issue with leaving the table as is - not that there is any way to revert it besides, dropping it from the server via ssh.

> Also, at this point, we should start looking into fixing this forward rather than rolling it back

why? the point of reverting is to get us to a place where we can still be shipping active task, while we figure autoretrieve out

alvin-reyes:
> > Question - do we want to remove the AR table along with this rollback?
> 
> I am not sure there is any issue with leaving the table as is - not that there is any way to revert it besides, dropping it from the server via ssh.
> 
> > Also, at this point, we should start looking into fixing this forward rather than rolling it back
> 
> why? the point of reverting is to get us to a place where we can still be shipping active task, while we figure autoretrieve out

Is there a way for us to verify this rollback change? My concern is rolling it back this way can also cause negative impact. If we just want to backout AR changes, we can cut 0.1.7 to a new branch, add the PRs after AR changes and merge it to master branch. We can then deploy this as 0.1.8.  If we can verify that this rollback PR won't have negative impact, then I'm good to merge this.

en0ma:
ok, closing this. I will focus my test now on how to get graphsync on work with latest master

PR #389, opened 2022-08-04
chore: revert filename to name

alvin-reyes:
# Changes

The change in the `Content` field would force us to rebuild the `contents` table and given that this field mapping is just for naming convenience (using filename instead of name) without any major functionality introduced, we should revert and just use the same field name. 

This PR reverts all the changes we did to rename the `name` field to `filename`.

# Verification
## estuary-www
image

## estuary endpoint tests
image


en0ma (approved):
LGTM

jlogelin (approved):
🚢

PR #393, opened 2022-08-05
Bump version errnotfound

gmelodie:
Update several packages. Here's the chronologic story behind this PR:
1. Autoretrieve integration #159 was merged
2. Deals started failing (the deal was successfully made, but the data could not be transferred)
3. We mapped a version of `go-legs` (introduced by the autoretrieve PR) as the culprit
4. Updating `go-legs` to a version that had the bug fixed proved challenging since a lot of packages were breaking because of [this commit on `go-ipfs-blockstore`](https://github.com/ipfs/go-ipfs-blockstore/commit/b598a29143df44364db9a1b80ddfc3549f7cdce5)
5. This PR updates all the versions needed and makes deals work again.

If this is merged then we no longer need #385


go.mod
```diff
@@ -4,99 +4,102 @@ go 1.17
...
-	github.com/cenkalti/backoff/v4 v4.1.2
+	github.com/application-research/go-bs-autobatch v0.0.0-20220803173502-846f018c8bf9
+	github.com/cenkalti/backoff/v4 v4.1.3
 	github.com/cheggaaa/pb/v3 v3.0.8
 	github.com/docker/go-units v0.4.0
 	github.com/dustin/go-humanize v1.0.0 // indirect
 	github.com/facebookgo/atomicfile v0.0.0-20151019160806-2de1f203e7d5
-	github.com/filecoin-project/boost v0.0.0-20220310131759-cd38741e464d
-	github.com/filecoin-project/go-address v0.0.6
+	github.com/filecoin-project/boost v1.2.1-0.20220803131342-cfcd4837136d
```
alvin-reyes, line 13: v1.2.1 is the master branch of boost. We should use the latest boost release tag instead.

go.mod
```diff
@@ -4,99 +4,102 @@ go 1.17
```
alvin-reyes, line 4: IMO, we need to revisit how we updated these modules. I suggest to re-evaluate each modules here and check if we really need to bump them or not. I can see a lot of non-`//indirect` modules here which means estuary `directly` uses them which can cause regression issues.

go.mod
```diff
@@ -345,43 +333,42 @@ require (
...
-	modernc.org/xc v1.0.0 // indirect
 )
 
 //replace github.com/libp2p/go-libp2p-yamux => github.com/libp2p/go-libp2p-yamux v0.5.1
 
 replace github.com/raulk/go-bs-tests => github.com/whyrusleeping/go-bs-tests v0.1.0
 
 replace github.com/filecoin-project/filecoin-ffi => ./extern/filecoin-ffi
+
+replace github.com/filecoin-project/index-provider => github.com/gmelodie/index-provider v0.8.2
```
alvin-reyes, line 370: We should try to use the original repository instead of forking and replacing it with ours. 

Same with `go-bs-lmdb` and `boost`

trackbs.go
```diff
@@ -163,7 +164,7 @@ func (tbs *TrackingBlockstore) Get(ctx context.Context, c cid.Cid) (blocks.Block
 	tbs.getCh <- c // TODO: should we be tracking all requests? or all successful servings?
 	blk, err := tbs.bs.Get(ctx, c)
 	if err != nil {
-		if xerrors.Is(err, blockstore.ErrNotFound) {
+		if blk == nil { // not found
```
alvin-reyes, line 167: import `github.com/ipfs/go-ipld-formats`

and use `if errors.Is(err, ipld.ErrNotFound{}) {` instead.

alvin-reyes (changes requested):

en0ma:
> This PR updates all the versions needed and makes deals work again.
@gmelodie I have verified deals and made and graphsync data transfers are started?

trackbs.go
Jorropo: @alvin-reyes this is wrapped in.
You can do `if ipld.IsNotFound(err)`, no need to do `errors.Is` directly.

alvin-reyes:
@gmelodie let's use this PR https://github.com/application-research/estuary/pull/413.

PR #397, opened 2022-08-12
Quality of life changes for set up script

snissn:
Changes to make the script idempotent so that it can be run multiple times, also have the script end early and fail if any step fails so that system debugging is easier


alvin-reyes (commented):
LGTM but I'll defer to @anjor or @schreck to review and approve.

snissn:
awesome much appreciated you merging it! Cheers 🎉

PR #417, opened 2022-08-26
Fix swagger annotation on delete miner endpoint

gmelodie:


docs/docs.go
```diff
@@ -1368,7 +1368,7 @@ const docTemplate = `{
                         "type": "string",
                         "description": "Filter by miner",
                         "name": "miner",
-                        "in": "query"
+                        "in": "path"
```
alvin-reyes, line 1371: this is automatic. all the swagger files (docs/*) are auto-generated after the swagger workflow job step.

alvin-reyes (changes requested):
the docs.go will be regenerated after it runs the swag.sh file.

docs/docs.go
gmelodie: I didn't change that file at all. Maybe it was a pre-commit hook or something?

gmelodie: It's a workflow on github: https://github.com/application-research/estuary/blob/a7420a2c22262bd6dcc1edb203eba6bb0f763afe/.github/workflows/swagger.yml#L18-L22

gmelodie:
@alvin-reyes I don't know what you mean. The commit that changed `docs.go` was not mine, it was created by the workflow we have on the repo

alvin-reyes (approved):
LGTM

PR #418, opened 2022-08-26
Document content/add response structure

RangerMauve:
I've added the structure to the docs.go file, but I'm not sure how the JSON and Swagger are supposed to get generated.

Is this the correct approach?

I might submit one for add-car in a bit too.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -815,6 +815,7 @@ func (s *Server) loadCar(ctx context.Context, bs blockstore.Blockstore, r io.Rea
 // @Param        file formData file true "File to upload"
 // @Param        coluuid path string false "Collection UUID"
 // @Param        dir path string false "Directory"
+// @Success      200  {object}  util.ContentAddResponse
 // @Router       /content/add [post]
 func (s *Server) handleAdd(c echo.Context, u *User) error {
 	ctx, span := s.tracer.Start(c.Request().Context(), "handleAdd", trace.WithAttributes(attribute.Int("user", int(u.ID))))


alvin-reyes:
docs.go is autogenerated based on the swagger annotations on the endpoints. Please check first if you can add this response structure on the main definition (top part of handler.go) first.

alvin-reyes (changes requested):
Please add the documentation on the code (annotation).

gmelodie:
Same comment as @alvin-reyes's: you need to annotate the handler function on `handlers.go` with something like:
```golang
// @Success 200 {object} MyResponseObject
```
Where `MyResponseObject` is a struct already defined and annotated as [here](https://github.com/swaggo/swag#user-defined-structure-with-an-array-type)

RangerMauve:
Is this syntax accurate? I'm not sure how it interacts with types from other packages. 😅

gmelodie:
@RangerMauve I think that's the right way of doing it, but @alvin-reyes there seems to be a bug where the responses never get shown on the swagger pages (see image below). Any idea what that is? I did some digging but couldn't find anything  
image

RangerMauve:
Yeah I think I noticed that too about responses. 😅 I found that the swagger docs were nice for finding what paths exist for the API, but ultimately I had to dig into the source code to actually use anything.

alvin-reyes:
I think there is a bug on the parses. The /pinning/pins has the failure responses [here](https://docs.estuary.tech/swagger-ui-page#/pinning/get_pinning_pins).

alvin-reyes:
Try putting the @success annotation after @Produce or @Param but before the @Router.

RangerMauve:
> Try putting the @success annotation after https://github.com/produce or @param but before the @router.

Like this?

gmelodie (approved):

PR #423, opened 2022-08-31
Add /public/miners/storage/query/:miner route to handleQueryAsk

gmelodie:


handlers.go
```diff
@@ -1537,6 +1537,7 @@ func (s *Server) handleGetContentByCid(c echo.Context) error {
 // @Produce      json
 // @Param 		 miner path string true "CID"
 // @Router       /deal/query/{miner} [get]
+// @Router       /public/miners/storage/query/miner} [get]
```
en0ma, line 1540:
```suggestion
// @Router       /public/miners/storage/query/{miner} [get]
```

docs/swagger.yaml
```diff
@@ -1043,6 +1043,21 @@ paths:
       tags:
         - public
         - miner
+  /public/miners/storage/query/miner}:
```
alvin-reyes, line 1046: FYI: This will be automatically removed by the swagger automation script. You should put any of these changes to the handle function annotation.

gmelodie: This was added by the swagger workflow on commit 811a8eb2838da3a3bb416d79535438a32c120e8d

PR #425, opened 2022-09-04
Add API endpoint for removing content from collection

pythonicode:
Hi, I thought I'd get this started. 

Definitely not finished (I couldn't test the endpoint on WSL).

@gmelodie


gmelodie:
Hey there! Thank you for the PR. I think we will have to change the approach first (that's the reason why I hadn't started on this yet as well). I'm thinking we'd have two endpoints: one for removing a content from a collection, and one for removing a content from the database. These two would kinda work together but I'm still debating on how to implement this. Please always be sure you're clear to work on an issue before starting to avoid miscommunication issues. Again, thank you very much for contributing!

PR #432, opened 2022-09-11
fix(auth): change passwordHash to base64 -> ensures compatibility with postgres

fabioespinosa:
### The error:

When trying to use the project with a postgres database, running setup with:

`./estuary setup --username=<uname> --password=<pword> --database=postgres=postgresql://[user[:password]@][host][:port][/dbname]
`

I get:

`FATAL	estuary	estuary/main.go:665	could not run estuary app: admin user creation failed: ERROR: invalid byte sequence for encoding "UTF8": 0xee 0x0d 0x26 (SQLSTATE 22021)	{"app_version": "v0.1.8-22-ga7420a2"}`

### Explanation:

When converting the password hash byte sequence to a string (to be saved in the db as a string) here: https://github.com/application-research/estuary/blob/master/util/auth.go#L30
The output contains characters that if tried to be encoded in UTF-8, would look like:

`��PK�uU����իagc����T0x`

The SQLite database will save it no problem, but postgres will complain with the error above making user registration and first user admin setup not possible. 

Already tried changing encoding of the postgres db to other formats without luck. Changing it to base64 made it work. 

### Fix (in this PR):

Save the hash byte sequence encoded in a base64 string. This way postgres will have no problem saving the hash as a UTF-8 string.

### Potential errors:
* **Backwards compatibility of passwords**
    * _Explanation_: If already existing users try to log in with their credentials they will not be able to do so since their saved hashed passwords are not in base64
    * _Mitigation_: Offer now two verifications: either it matches via `GetPasswordHashBase` (previous method) or `GetPasswordHashBase64` (proposed method). Included in this PR.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -2867,7 +2867,7 @@ func (s *Server) handleRegisterUser(c echo.Context) error {
 		Username: username,
 		UUID:     uuid.New().String(),
 		Salt:     salt,
-		PassHash: util.GetPasswordHash(reg.Password, salt),
+		PassHash: util.GetPasswordHashBase64(reg.Password, salt),
 		Perm:     util.PermLevelUser,
 	}
 
@@ -2927,7 +2927,7 @@ func (s *Server) handleLoginUser(c echo.Context) error {
 	}
 
 	//	validate password
-	if ((user.Salt != "") && user.PassHash != util.GetPasswordHash(body.Password, user.Salt)) ||
+	if ((user.Salt != "") && (user.PassHash != util.GetPasswordHash(body.Password, user.Salt)) || (user.PassHash != util.GetPasswordHashBase64(body.Password, user.Salt))) ||
 		((user.Salt == "") && (user.PassHash != body.Password)) {
 		return &util.HttpError{
 			Code:   http.StatusForbidden,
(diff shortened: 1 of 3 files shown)


alvin-reyes:
Great! Please create a PR against the `dev` branch instead. I'd like to include this on the next release.

alvin-reyes:
@fabioespinosa can you post your verification steps please? I tried running some tests on my local and it's failing to authenticate the password. 

I use `alvin` as password and the comparison checks is returning a different hash

```
PassHash from DB: 0V9V9RyP8ZMMTUoleGnED3OHFQXHrlxWbc40QAzMMFI=
util.GetPasswordHashBase64: lk1S67MDDiirtlAIozzt9f0PE3rAN02d8OQdI/b+Hjk=
```

fabioespinosa:
Are you using the first user that is created to authenticate? 

I believe using those credentials is not working because of this https://github.com/application-research/estuary/pull/433

However if you add an invite_code manually in the db and create a user, they should match

alvin-reyes (approved):
Let's merge this.

We will switch to postgres and drop the support for SQLlite.

PR #433, opened 2022-09-11
docs(readme): Add clarification of first user credentials

fabioespinosa:
### The problem

Users via the front-end when providing credentials are using a predefined salt defined here: https://github.com/application-research/estuary-www/blob/master/common/constants.ts#L28  for their password. The resulting hash is then safely sent to authenticate. This goes for signing-up/logging-in.

Now when creating the first user via:

`./estuary setup --username=<uname> --password=<pword> --database=XXXXX`

The provided password is saved by salting it with a randomly generated `uuid` not the salt provided above. Therefore when trying to authenticate using this same credentials, the resulting hash will not match and the first user is not able to authenticate. The user can use a workaround by using the token and logging in with it. However it might be confusing to see the provided credentials are not working for first time users in estuary-www.

### The Fix

It could be possible to salt the first user's password in the backend with the same salt, not sure if maintaining a copy of the same salt (in estuary-www and estuary) desired, since at some point it might diverge.

However, providing some documentation provided in this PR will at least avoid confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ To run locally in a 'dev' environment, first run:
 ```
 
 Save the auth token that this outputs, you will need it for interacting with
-and controlling the node.
+and controlling the node. This username and password won't work to log in using the front end (estuary-www), but the auth token will.
 
 NOTE: if you want to use a different database than a sqlite instance stored in your local directory, you will need to configure that with the `--database` flag, like so: `./estuary setup --username=<uname> --password=<pword> --database=XXXXX`
 
@@ -89,15 +89,18 @@ npm run dev
 
 And then head to [localhost:4444/staging](http://localhost:4444/staging) to see the status of your deal.
 
-
 ## Contributing
+
 See [`CONTRIBUTING.md`](CONTRIBUTING.md) for contributing and development instructions.
 
 ## Troubleshooting
+
 Make sure to install all dependencies as indicated above. Here are a few issues that one can encounter while building estuary
 
 ### Guide for: `route ip+net: netlinkrib: too many open files`
+
 #### Error
+
 If you get the following error:
 
 ```sh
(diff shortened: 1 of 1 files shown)


en0ma:
@fabioespinosa I think it's best to fix the CLI setup to use the salt instead

alvin-reyes:
Please switch the PR to use the `dev` branch.

fabioespinosa:
> Please switch the PR to use the `dev` branch.

Sure!

alvin-reyes (approved):
LGTM

en0ma:
@alvin-reyes @fabioespinosa we still need to fix the CLI setup to use the default salt

fabioespinosa:
@en0ma Yes, absolutely. Let me try to take some time next weekend to investigate more about how we can use the default salt in the CLI. 

We can then remove the comment from the readme. But in the meantime, I think it's useful for new users to be advised of this small detail.

alvin-reyes:
Yes. I'm actually trying to fix this with a new approach (that's why I merged it on `dev`). My approach is to use the dialect as the basis of the hashing function.

```
func GetPasswordHash(password, salt string, dialector string) string {
	fmt.Println("dialector", dialector)
	switch dialector { // can be "postgres", "sqlite"
	case "postgres": // for postgres and other rdbms
		return GetPasswordHashBase64(password, salt)
	default: // for sqllite, embedded db.
		return GetPasswordHashSum256(password, salt)
	}
}
```

I'm also looking to remove the `salt` on the frontend and let the backend handle all the validation. Having the `salt` exposed is not desirable. The password is transported via HTTPS and TLS so the transportation of the password is protected anyway.

fabioespinosa:
That should totally fix it. 

The only issue I can think of that approach is if there was a decision to migrate from databases and after the migration, validating the existing password hashes won't work. But if this is something that is assured won't happen, then it should be all good. 

If it was all kept in Base64 we'd be blinded to that scenario though, I'm curious what you think about just keeping it in Base64 🤔

alvin-reyes:
PRs for this:

https://github.com/application-research/estuary/pull/441
https://github.com/application-research/estuary-www/pull/66

I added a simple change on the login screen to tick if the current user is created via estuary cli.

PR #434, opened 2022-09-12
Full endpoint paths on echo

gmelodie:
As discussed with @jimmylee, we should use full endpoint paths on the code, which makes it easier to locate ourselves when developing.


alvin-reyes:
I have some reservations in this PR. 

1 - the purpose of echo having a grouping is to `Group` the endpoints. The proposed change seems to underutilize the grouping mechanism. 

2 - I'm not entirely sure the difference in terms of readability e.i being able to `search` endpoints easily. I don't see a huge difference in this regard. We can collectively vote if we want to group or not but I prefer not to change any existing functionality that already works. 

3 - If we are to change the routes, please add some verification tests on this. I think this is a good opportunity to build the unit-tests. I have a branch where I already started working on each of the unit tests for each handler. `unit-tests`

gmelodie:
To address 1 and 2: @jimmylee told me he liked this way better and I agreed in terms of knowing which what is the entire path of an endpoint just by looking at the `handler.go` file.

jimmylee:
We'll probably need to talk about this more because it effects a bunch of people. I do like experimenting here on readability and trying new things out. But maybe we can have @whyrusleeping + others weigh in too.

jlogelin:
> We'll probably need to talk about this more because it effects a bunch of people.

Agreed, we can't keep changing the API contract, as there are downstream effects on apps. It might be a good idea for us to snapshot our current REST api with a version prefix (`https://upload.estuary.tech/v1/content/add`). At the very least, we could also just leave the current "version 1 api" url alone and simply add a `v2` namespace for the updated api:

Current `https://upload.estuary.tech/content/add`
(also maps to `https://upload.estuary.tech/v1/content/add`)

Updated `https://upload.estuary.tech/v2/content/add`

alvin-reyes (changes requested):
We need more discussion before changing this.

alvin-reyes:
@gmelodie We should revisit this with the Versioning. Your changes here are close to what is required there.

gmelodie:
Closing this as stale. We need to revisit this problem after the `api-v2` branch is merged.

PR #439, opened 2022-09-15
add rustup to requirements and a note about libhwloc-dev on linux

44203:
just a couple of readme updates when setting up from scratch on ubuntu


en0ma (dismissed):

README.md
```diff
@@ -128,6 +129,8 @@ For M1 Macs, here's the following steps needed
 - Step 2: export LIBRARY_PATH=/opt/homebrew/lib
 - Step 3: Follow the steps as per the docs.
 
+On Ubuntu, install `libhwloc-dev`.
```
gmelodie, line 132: This should be in the same line as `- [hwloc](https://www.open-mpi.org/projects/hwloc/)`

gmelodie (changes requested):

gmelodie:
Also why do we need `rustup`?

44203:
> Also why do we need `rustup`?

Unsure, but running `make clean all` fails if it isn't installed and asks for it.

alvin-reyes:
Please move use `dev` as the upstream branch for this PR.

PR #443, opened 2022-09-20
Ignore trailing slash on API endpoints

### LINKED ISSUE #438: Ignore trailing slash on API endpoints
If we use `/collections/` with the trailing slash it works:
```bash
$ curl -X POST -H "Authorization: Bearer EST599c28b4-fd89-4b44-b579-74af0997a5ecARY " http://localhost:3004/collections/ -d '{ "name": "A new collection", "description": "A new collection test" }' -s | jq
{
  "createdAt": "2022-09-14T19:54:44.174049754Z",
  "uuid": "0c222e8d-007c-486e-b526-0c4b716ff031",
  "name": "A new collection",
  "description": "A new collection test",
  "userId": 1,
  "cid": ""
}
```

But without the trailing slash it doesn't find the endpoint:
```bash
$ curl -X POST -H "Authorization: Bearer EST599c28b4-fd89-4b44-b579-74af0997a5ecARY " http://localhost:3004/collections -d '{ "name": "A new collection", "description": "A new collection test" }' -s | jq 
{
  "error": {
    "code": 404,
    "reason": "Not Found",
    "details": "Not Found"
  }
}
```

I would like estuary to treat both as valid endpoints.

jcace:
Closes #438 

This change will strip the trailing slash from any incoming request. The caller can either make a request with or without it (ex, `http://localhost:3004/health` or `http://localhost:3004/health/`),  both will work now.


en0ma:
@jcace please change destination branch to dev, and also apply this middle to this echo instance https://github.com/application-research/estuary/blob/master/cmd/estuary-shuttle/main.go#L1050

jcace:
Thanks @en0ma , made those changes!

en0ma (approved):

alvin-reyes (changes requested):
we need to discuss the impact of this before we merge.

jcace:
> This impacts `collections` api. `collections` api uses a trailing slash to work. Without these, it'll return 404 and can impact the current users.

I modified it [here](https://github.com/application-research/estuary/pull/443/files#diff-56307bdb7ded67ebe516f562b90384407b21b2d34e8b5766b42602e1fb9705f0L196) - removed the "/" at the root of the `collections ` endpoint, and tested it out locally. It seemed to work with both a trailing / and without one

alvin-reyes:
It doesn’t need a trailing hash but as it stands today, it requires a trailing hash. We need to make it backward compatible and communicate these changes to the existing users.

gmelodie (dismissed):
LGTM

neelvirdy:
@alvin-reyes do you still have concerns around this? i dont think this is a break since the PR makes both options work (with or without trailing slash). can also ask @anjor to do a quick test since he's probably the collections API's biggest power user right now.

jcace:
Fixed merge conflicts - let me know if any other concerns guys

alvin-reyes (approved):
LGTM

Issue #331, opened 2022-07-19
Missing objects and object refs for contents


en0ma:
**Describe the bug**
While investigating the commp issue, I noticed content with id `30545853` in Estuary is missing its `objects` and `object_refs` in Estuary and in its shuttle.  I suspect this kind of issue must have happened due to downtime on primary when content is being created or something. 

**Expected behavior**
The Content should have its object and object_refs database records in both primary and shuttle.

**Actual behavior**
The Content is missing its object and object_refs database records in both primary and shuttle.

- [x] find and fix the bug that causes this
- [ ] A backfill/verification tool should be written to detect and backfill such contents.

en0ma:
> On the verification side, are you thinking a tool that goes through the database, and finds any consistencies?
> 
> So potentially, we can configure some "invariants" that the data model should respect, and this tool checks whether any of those invariants are broken and flags them.
> 
> Fixing these inconsistencies would probably look like (1) backfilling if we can or (2) dropping the rows where it is not possible to backfill -- this will involve data loss.
> 
> Is my understanding roughly correct?

yes @anjor

PR #444, opened 2022-09-23
Split, Aggregate, deal checking and some other improvements

en0ma:
This PR brings;

- Support to validate `content type` (JSON) for the pinning endpoint
- A catch-all payload validation error to return 400 HTTP error instead of 500 (this will be improved in another PR)
- Use appropriate log levels where necessary and quiet filclient to reduce logging noise
- Fix shuttle and primary content split request to set the correct columns in DB
- Fix shuttle split request that does not send objects of the split, leading to missing objects in primary DB - https://github.com/application-research/estuary/issues/331#issue-1309844260
- Improve content aggregation to respect (not select) shuttles with `disable-content-adding` flag
- Improve data transfer state retrieval to display in the UI correctly
- Refactored deal check to improve readability and make it more deterministic 

closes https://github.com/application-research/estuary/issues/331#issue-1309844260


handlers.go
```diff
@@ -214,10 +209,11 @@ func (s *Server) ServeAPI() error {
 	pinning.Use(openApiMiddleware)
 	pinning.Use(s.AuthRequired(util.PermLevelUser))
 	pinning.GET("/pins", withUser(s.handleListPins))
-	pinning.POST("/pins", withUser(s.handleAddPin))
 	pinning.GET("/pins/:pinid", withUser(s.handleGetPin))
-	pinning.POST("/pins/:pinid", withUser(s.handleReplacePin))
 	pinning.DELETE("/pins/:pinid", withUser(s.handleDeletePin))
+	pinning.Use(util.JSONPayloadMiddleware)
```
alvin-reyes, line 214: What's the difference between the previous way we handle json payload? 

If possible, please run and document tests to ensure this is backward compatible.

en0ma: I noticed many users do not read the IPFS pinning spec to know those endpoints only expect a JSON payload (content type). Ideally, we should verify our APIs only allows content types we support. This change will help communicate the error so users can make the change instead of coming to slack to figure out why their request failed.

I have tested this with the IPFS pinning compliance tool.

When standardizing our API to be more restful, both content-type and accept types should be verified if there are supported.

alvin-reyes:
LGTM. Please fix the conflicts. I'd like to include this on the next release.

alvin-reyes (dismissed):

jlogelin (dismissed):
LGTM

gmelodie (approved):

PR #454, opened 2022-10-08
bugfix for mac silicon - makes sure homebrew path is in library path

snissn:
seems homebrew's library path not being added to LIBRARY_PATH is a recurring issue and this should fix it

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,9 @@ FFI_PATH:=extern/filecoin-ffi/
 FFI_DEPS:=.install-filcrypto
 FFI_DEPS:=$(addprefix $(FFI_PATH),$(FFI_DEPS))
 
-#For mac silicon
-export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)
+ifeq ($(OS),Darwin)
+	export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)
+endif
 
 $(FFI_DEPS): build/.filecoin-install ;
 
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,9 @@ FFI_PATH:=extern/filecoin-ffi/
 FFI_DEPS:=.install-filcrypto
 FFI_DEPS:=$(addprefix $(FFI_PATH),$(FFI_DEPS))
 
+#For mac silicon
+export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)
+
 $(FFI_DEPS): build/.filecoin-install ;
 
 build/.filecoin-install: $(FFI_PATH)


jimmylee (approved):
LGTM, seems fine.

alvin-reyes (changes requested):
Looks good. Can we instead have a flag here like `BUILD_ENV=mac` or `INCLUDE_HOMEBREW=true`? 

then add the following on the Makefile
```
ifeq ($(BUILD_ENV),"mac")
	export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)
endif
```

OR

```
ifeq ($(INCLUDE_HOMEBREW),"true")
	export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)
endif
```

snissn:
> Looks good. Can we instead have a flag here like `BUILD_ENV=mac` or `INCLUDE_HOMEBREW=true`?
> 
> then add the following on the Makefile
> 
> ```
> ifeq ($(BUILD_ENV),"mac")
> 	export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)
> endif
> ```
> 
> OR
> 
> ```
> ifeq ($(INCLUDE_HOMEBREW),"true")
> 	export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)
> endif
> ```

I don't understand -- is BUILD_ENV or INCLUDE_HOMEBREW a variable that the make chain understands? Thanks

alvin-reyes:
It'll be a variable that the make chain will understand when you pass it. You then parse the variable and set the LIBRARY path based on the value.

The goal is to allow the developer to build it specifically for mac/m1 and not force any non-mac users to export a non-existent path.
```
make all BUILD_ENV=mac 

OR 

make all INCLUDE_HOMEBREW=true
```

> > Looks good. Can we instead have a flag here like `BUILD_ENV=mac` or `INCLUDE_HOMEBREW=true`?
> > then add the following on the Makefile
> > ```
> > ifeq ($(BUILD_ENV),"mac")
> > 	export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)
> > endif
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > OR
> > ```
> > ifeq ($(INCLUDE_HOMEBREW),"true")
> > 	export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)
> > endif
> > ```
> 
> I don't understand -- is BUILD_ENV or INCLUDE_HOMEBREW a variable that the make chain understands? Thanks

snissn:
Hey! I didn't notice your comment until now, sorry! 

So is the idea that a user who's building estuary will know to add an environment flag when compiling estuary and will type `make all BUILD_ENV=mac` into the shell? That feels like a source of error and confusion for people compiling this, and is about as confusing as requiring a user anyway to know that they need to add the homebrew path to LIBRARY_PATH

> not force any non-mac users to export a non-existent path.

My understanding of the unix philosophy and how path works is that it's not a problem to append non existence paths to the LIBRARY_PATH variable, this PR just simply says to the make chain in addition to all the random folders you're checking that may or may not have the library you're looking for also check /opt/homebrew/lib

https://unix.stackexchange.com/questions/96577/is-there-a-problem-if-i-add-a-non-existing-path-to-path-environment-variable

the code appends  /opt/homebrew/lib to LIBRARY_PATH it doesn't replace it 

```export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)``` 

IMO this PR just makes a build problem go away automatically with essentially 0 downside except an extra few system calls in the build chain to potentially learn a directory doesn't exist

**Comment by alvin-reyes** [comment]
`
So is the idea that a user who's building estuary will know to add an environment flag when compiling estuary and will type make all BUILD_ENV=mac into the shell? That feels like a source of error and confusion for people compiling this, and is about as confusing as requiring a user anyway to know that they need to add the homebrew path to LIBRARY_PATH
`

I don't see why this would confusing. If you are building on a mac, it make sense to check if there are mac specific libraries you need to add. Same if we will eventually support windows. 

`
My understanding of the unix philosophy and how path works is that it's not a problem to append non existence paths to the LIBRARY_PATH variable, this PR just simply says to the make chain in addition to all the random folders you're checking that may or may not have the library you're looking for also check /opt/homebrew/lib
`

It won't be an issue if the path doesn't exist. It will be when it actually exist. Given `homebrew` might never be on any linux system other than Mac. 

I think its better to auto-detect it then rather that forcing it. It's much cleaner and we can add more conditions to it per operating system.

```
ifeq ($(OS),Darwin)
	export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)
endif
```

**Comment by snissn** [comment]
wow awsome!! thanks 
```
ifeq ($(OS),Darwin)
	export LIBRARY_PATH := /opt/homebrew/lib:$(LIBRARY_PATH)
endif
```

is perfect! I'll update the PR

en0ma (approved):

en0ma:
@alvin-reyes can you have a final look at this, so we can get it in

snissn (commented):
I think my second commit https://github.com/application-research/estuary/pull/454/commits/b0a1efca9ca61de57091332a29b9958df47610e6 addresses this -- 

@alvin-reyes  can you please confirm?

snissn:
@alvin-reyes I think that this has been ready to be merged and your concerns were addressed

alvin-reyes (approved):

Issue #408, opened 2022-08-18
Queued pins show as pinning


lidel:
There is a bug in Pinning Service API implementation, `queued` pins show as `pinning` (tested with Kubo 0.14):

```console
$ ipfs pin remote ls --service=estuary --status=queued
$ ipfs pin remote ls --service=estuary --status=pinning
QmbCpjNfbwbMvJqMUUHAoDAjNqdoZFo7zWEGBkCCWEZ6nA	queued	test-pin-25
QmfQvgMkYj77amUviAqCT2kf3QkwsiCFRYc21SfPXSkpn8	queued	test-pin-24
QmbqbK7pakRu2rmp38rKpp1EwnDZ9ikQbQrYiQ3sAMeofR	queued	test-pin-23
QmTPZEhFu9tRcxNEvexkhz3Zoi4U9t9mmGncfhKqt1VyGf	queued	test-pin-22
QmNThD2GkZMtzWoDvuHjb2XQnSH5NEFgWwzR5h6cqVZkB9	queued	test-pin-21
QmUqgwLckMg5QvdtXgd9WES726TqfoTX3fRpsV57cy79YU	queued	test-pin-20
QmQTzvHwJZ8N3G8tQdiZ9ykGD3Nnxb5cyUh9tJRSXLeAmT	queued	test-pin-19
QmfPN5b8nZdGNeqBtZtQu7V2GSnfUKM3fg1HGFnzfB9GoY	queued	test-pin-18
QmZqzBwfCLSqN7VauZHQHnQPt925CGFy3XoT6HxJ32cZS5	queued	test-pin-17
QmekSUExxjF1pycq1DtaQZyHtMq6kCNjVB3ZogSdgEguB2	queued	test-pin-16
QmekSUExxjF1pycq1DtaQZyHtMq6kCNjVB3ZogSdgEguB2	queued	test-pin-16
QmaHq597QaoiqjHjwbFz2DKxjuezkPVhQ5iwGF5gxfNHb4	queued	test-pin-15
QmRAaaZxYDdWjTERbPENBtqoC5aDnc5vf4Qg472ssnyCic	queued	test-pin-14
QmU718ekePHsiEVhbRRvmRhkGb4aJwqeCoqGSD6mvdXSG6	queued	test-pin-13
QmdoX2NpQU7YpZzGqEAXjX4xAzkWAmFMtrHvwEgrN7vQpk	queued	test-pin-12
[...]

PR #457, opened 2022-10-13
fix pin mgr status and send peers info to shuttles for pinning

en0ma:
This PR;

1. Fixes the issue of wrong pin status raised in https://github.com/application-research/estuary/issues/408 and https://filecoinproject.slack.com/archives/C016APFREQK/p1665650121296659 by keeping only one pin status state (the one in DB, no need having 2 states in pinMgr and Db)
2. Improves shuttle pinning success rate by propagating origins from pin request to shuttle for peers' discoverability
3. Fixes shuttle pinning status for consolidated contents
4. Properly trigger deal-making for shuttle aggregated contents
5. Fixes consolidation bugs
 
These have been tested

closes https://github.com/application-research/estuary/issues/408

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -1532,25 +1532,8 @@ func (d *Shuttle) doPinning(ctx context.Context, op *pinner.PinningOperation, cb
 	dsess := dserv.Session(ctx)
 
 	if err := d.addDatabaseTrackingToContent(ctx, op.ContId, dsess, d.Node.Blockstore, op.Obj, cb); err != nil {
-		// pinning failed, we wont try again. mark pin as dead
-		/* maybe its fine if we retry later?
-		if err := d.DB.Model(Pin{}).Where("content = ?", op.ContId).UpdateColumns(map[string]interface{}{
-			"pinning": false,
-		}).Error; err != nil {
-			log.Errorf("failed to update failed pin status: %s", err)
-		}
-		*/
-
 		return errors.Wrapf(err, "failed to addDatabaseTrackingToContent - contID(%d), cid(%s)", op.ContId, op.Obj.String())
 	}
-
-	/*
-		if op.Replace > 0 {
-			if err := s.CM.RemoveContent(ctx, op.Replace, true); err != nil {
-				log.Infof("failed to remove content in replacement: %d", op.Replace)
-			}
-		}
-	*/
 
 	if err := d.Provide(ctx, op.Obj); err != nil {
 		return errors.Wrapf(err, "failed to provide - contID(%d), cid(%s)", op.ContId, op.Obj.String())
@@ -1725,7 +1708,6 @@ func (s *Shuttle) refreshPinQueue() error {
 	for _, c := range toPin {
 		s.addPinToQueue(c, nil, 0)
 	}
-
 	return nil
 }
 
--- a/cmd/estuary-shuttle/rpc.go
+++ b/cmd/estuary-shuttle/rpc.go
@@ -16,6 +16,7 @@ import (
 	"github.com/ipfs/go-cid"
 	cbor "github.com/ipfs/go-ipld-cbor"
 	"github.com/ipfs/go-merkledag"
+	"github.com/libp2p/go-libp2p-core/peer"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/xerrors"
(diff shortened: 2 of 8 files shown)


cmd/estuary-shuttle/rpc.go
```diff
@@ -278,18 +280,17 @@ func (d *Shuttle) handleRpcTakeContent(ctx context.Context, cmd *drpc.TakeConten
...
-			if count > 1 {
-				log.Errorf("have multiple pins for same content: %d", c.ID)
-			}
 			continue
 		}
 
-		if err := d.addPin(ctx, c.ID, c.Cid, c.UserID, true); err != nil {
-			return err
-		}
+		go func(c drpc.ContentFetch) {
```
alvin-reyes, line 288: why did we need to create a go routine here?

cmd/estuary-shuttle/rpc.go
```diff
@@ -354,7 +355,8 @@ func (d *Shuttle) handleRpcAggregateContent(ctx context.Context, cmd *drpc.Aggre
 	}
 
 	if err := d.DB.Model(Pin{}).Where("id = ?", pin.ID).UpdateColumns(map[string]interface{}{
-		"active": true,
+		"active":  true,
+		"pinning": false,
```
alvin-reyes, line 359: Can you please add a description on the statuses here on this PR?

cmd/estuary-shuttle/rpc.go
en0ma: to make sure each pin request does not block each other, even if 1 errors - best effort

cmd/estuary-shuttle/rpc.go
en0ma: done

cmd/estuary-shuttle/main.go
```diff
@@ -1532,26 +1532,9 @@ func (d *Shuttle) doPinning(ctx context.Context, op *pinner.PinningOperation, cb
 	dsess := dserv.Session(ctx)
 
 	if err := d.addDatabaseTrackingToContent(ctx, op.ContId, dsess, d.Node.Blockstore, op.Obj, cb); err != nil {
-		// pinning failed, we wont try again. mark pin as dead
```
alvin-reyes, line 1535: this means that the pinning will still be "true" which in this case will be retried?

cmd/estuary-shuttle/rpc.go
```diff
@@ -79,10 +80,10 @@ func (d *Shuttle) sendRpcMessage(ctx context.Context, msg *drpc.Message) error {
 func (d *Shuttle) handleRpcAddPin(ctx context.Context, apo *drpc.AddPin) error {
 	d.addPinLk.Lock()
 	defer d.addPinLk.Unlock()
-	return d.addPin(ctx, apo.DBID, apo.Cid, apo.UserId, false)
+	return d.addPin(ctx, apo.DBID, apo.Cid, apo.UserId, apo.Peers, false)
 }
 
-func (d *Shuttle) addPin(ctx context.Context, contid uint, data cid.Cid, user uint, skipLimiter bool) error {
+func (d *Shuttle) addPin(ctx context.Context, contid uint, data cid.Cid, user uint, peers []*peer.AddrInfo, skipLimiter bool) error {
```
alvin-reyes, line 86: Looks good. So just to re-iterate here: You passed the peers here so that every time you fetch the content via the new go routine call you made, you'll look it up to the defined peers (origins)?

replication.go
```diff
@@ -252,6 +251,12 @@ func (cb *contentStagingZone) isReady() bool {
 func (cm *ContentManager) tryAddContent(cb *contentStagingZone, c util.Content) (bool, error) {
 	cb.lk.Lock()
 	defer cb.lk.Unlock()
+
+	// if this bucket is being consolidated, do not add anymoe content
```
alvin-reyes, line 255: nit: `anymore`

alvin-reyes (dismissed):
LGTM. I'd like to merge this along with the fix on the split/aggregate. I'd like to test this out on a new environment.

cmd/estuary-shuttle/rpc.go
en0ma: yes, this will help improve pins discoverability for shuttles

cmd/estuary-shuttle/main.go
en0ma: yes, I thought about uncommenting, for the time being, to help with the pressure. Later in the future, we add support for retries - what do you think?

en0ma: Ah, I remember now, I removed this comment because I already made sure the pinning status is marked false if it fails in some PR in the past. the `onPinStatusUpdate` func is called anything a pin attempt fails from here - https://github.com/application-research/estuary/blob/dev/cmd/estuary-shuttle/main.go#L506

Which marks changes the pinning status to false - so no reties

alvin-reyes (approved):

PR #458, opened 2022-10-13
guard against duplicate Pin Queue

snissn:
work in progress

### DIFF
--- a/pinner/pinmgr.go
+++ b/pinner/pinmgr.go
@@ -37,6 +37,7 @@ func NewPinManager(pinfunc PinFunc, scf PinStatusFunc, opts *PinManagerOpts) *Pi
 		pinQueueIn:       make(chan *PinningOperation, 64),
 		pinQueueOut:      make(chan *PinningOperation),
 		pinComplete:      make(chan *PinningOperation, 64),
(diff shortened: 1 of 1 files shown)


snissn:
added feature to also remove work from the guard, no needs code review and testing

pinner/pinmgr.go
```diff
@@ -95,6 +97,38 @@ type PinningOperation struct {
...
+	Location    string
+	SkipLimiter bool
+	MakeDeal    bool
+}
+
+func getPinningData(po *PinningOperation) PinningOperationData {
+	return PinningOperationData{
+		Obj:  po.Obj,
+		Name: po.Name,
+		//Peers:       po.Peers,
```
en0ma, line 119: the peers are needed

snissn: thanks! That was my kind of only question - any tips on how to serialize the value for it ? 

also i'm debating whether the work queue guard is necessary or if backing the queue to disk is sufficient.

it would depend on whether there is erroneously duplicated data in the queue or not.

Also, no worries if it's too much but it would be helpful if you could you share a quick explanation of what the Peers datastructure is

en0ma: I think the `ContId` is sufficient enough to use for dups check

PR #459, opened 2022-10-13
checkin first version of durable queue pin manager

snissn:


en0ma:
@snissn please have a look at this thread on some issues that might help guide your implementation https://filecoinproject.slack.com/archives/C016APFREQK/p1660258369066179

alvin-reyes:
@snissn great start man. :) 

Let's isolate this durable queue pin manager approach and isolate this on its own branch until we are ready to put it in prod.

next steps:
1 - since you already have it on your own branch, I can create a small equinix server that we can use to isolate the durable queue pin manager.
2 - let's use your branch and the environment to test it. I just want to isolate this change for now especially with all the improvements made by @en0ma.

PR #463, opened 2022-10-15
Pin mgr test

snissn:
this test exposes a bug in the pin manager where if the number of workers > 1 some pins will be lost and never worked on

### DIFF
--- a/pinner/pinmgr_test.go
+++ b/pinner/pinmgr_test.go
@@ -2,6 +2,7 @@ package pinner
 
 import (
 	"context"
+	"time"
 
 	"testing"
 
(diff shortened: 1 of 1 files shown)


snissn:
the number is small though, if there are 1000 pins that should be worked on 996 will get worked on

jlogelin (dismissed):
Awesome, thanks @snissn - looks great!

en0ma (approved):

alvin-reyes (dismissed):
LGTM. 

I would prefer to use the ginkgo framework but this works. If you have some cycles, please convert this to gingko 

https://github.com/application-research/estuary/blob/unit-tests/tests/unit/handler_content_test.go

snissn:
cool! Definitely can & will convert to ginko in the backlog! Feeling good now about the state of the test and also about the state of the disk based dequeue, checking into the filter/guard branch next

snissn:
btw i pushed a few fixes to this branch so it needs another approval

alvin-reyes (approved):
LGTM

PR #464, opened 2022-10-15
Add get content by content ID endpoint

en0ma:
There have been numerous asks for this, and I am surprised it was never there in the first place.

Also improve some API response and fix some minor bugs


jlogelin (approved):
LGTM

Issue #462, opened 2022-10-14
Uploading to upload.estuary.tech does not work as expected


jlogelin:
**Describe the bug**
When I push content up through `https://upload.estuary.tech/content/add` data does seem to be pushed up to one of the servers, as reflected by a correct ipfs.dweb.link. However the content metadata is not reflected in the user interface, or through the api (via https://api.estuary.tech/content/by-cid/<CID>)

**To Reproduce**
1. Load content via
```
curl --request POST 'https://upload.estuary.tech/content/add' \
--header 'Authorization: Bearer XXX' \
--form 'data=@"/foo/YOUR-DATA.png"'
``` 
capture the resulting CID
Note: the https://CID.ipfs.dweb.link works
2. In the web app navigate to https://estuary.tech/home and attempt to query your data, you will not find it.
3. From the command line attempt to query the content metadata with:
```
curl --request GET 'https://api.estuary.tech/content/by-cid/CID' --header 'Authorization: Bearer XXX'
```
Note the incorrect, empty resultset.

**Expected behavior**
Content should show up in both UI and via API calls

**Actual behavior**
Described above

**Additional context**
NA

en0ma:
@jlogelin any sample CID to look at?

jlogelin:
> @jlogelin any sample CID to look at?

`bafkreigreegsiyihdx3ugen2obrdns3jixpuobhbkz7mw4bxzlwsdm3ehi`

It is easy to reproduce, but this is the response I got back with the last test I ran:

```
{
   "cid":"bafkreigreegsiyihdx3ugen2obrdns3jixpuobhbkz7mw4bxzlwsdm3ehi",
   "retrieval_url":"https://dweb.link/ipfs/bafkreigreegsiyihdx3ugen2obrdns3jixpuobhbkz7mw4bxzlwsdm3ehi",
   "estuaryId":39989033,
   "providers":[
      "/ip4/139.178.81.141/tcp/6745/p2p/12D3KooWHe2fZ6jJJ2K2oX3EjMiXMRzvHVA5mvgYYvZvi7tEYNMe",
      "/ip4/127.0.0.1/tcp/6745/p2p/12D3KooWHe2fZ6jJJ2K2oX3EjMiXMRzvHVA5mvgYYvZvi7tEYNMe",
[...]

PR #467, opened 2022-10-16
List full contents for UI response, Only return content with deals for UI deals response, and Fix dropped RPC messages 

en0ma:
This PR brings some improvements and bug fixes;

- The API response for the contents UI, will now return all contents (this will address the UX issue raised in https://github.com/application-research/estuary/issues/462 and https://filecoinproject.slack.com/archives/C016APFREQK/p1665824335920899?thread_ts=1665650121.296659&cid=C016APFREQK

**before**
Screenshot 2022-10-16 at 07 13 18

**after**
Screenshot 2022-10-16 at 07 10 51

- The UI deals page will now only show the deals for contents that actually have deals being made - see https://filecoinproject.slack.com/archives/C016APFREQK/p1665674619728969. It intermittently picks up newly added contents, even small ones just before they are added to the staging zone due to race conditions. This fix, will do away with race checks and really directly on the deal on the database.

**before** (only content 18 has deal attempt, but it shows contents 18 and 21 - 21 should not be there)
Screenshot 2022-10-16 at 07 54 54
Screenshot 2022-10-16 at 07 55 11

**after** (only 18 now shows)
Screenshot 2022-10-16 at 07 56 03

- Improve the shuttle API response, noticed some HTTP error codes were not properly reported

**before**
Screenshot 2022-10-16 at 05 27 51

**after**
Screenshot 2022-10-16 at 05 34 08

- Fixed the first part of dropped RPC messages, by using a buffered queue for primary (this will fix such bugs reported https://filecoinproject.slack.com/archives/C016APFREQK/p1665771946484999?thread_ts=1665623630.317789&cid=C016APFREQK and https://github.com/application-research/estuary/issues/462.
 The second PR will be created to handle resending pin RPC messages for situations where the shuttle was down or crashed before receiving/processing RPC request

closes https://github.com/application-research/estuary/issues/462

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -1244,24 +1244,31 @@ func (s *Server) handleListContentWithDeals(c echo.Context, u *util.User) error 
 	}
 
 	var contents []util.Content
-	if err := s.DB.Limit(limit).Offset(offset).Order("id desc").Find(&contents, "active and user_id = ? and not aggregated_in > 0", u.ID).Error; err != nil {
+	err := s.DB.Model(&util.Content{}).
+		Limit(limit).
+		Offset(offset).
+		Order("contents.id desc").
+		Joins("inner join content_deals on contents.id = content_deals.content").
+		Where("contents.active and contents.user_id = ? and not contents.aggregated_in > 0", u.ID).
+		Group("contents.id").
+		Scan(&contents).Error
+
+	if err != nil {
 		return err
 	}
 
 	out := make([]expandedContent, 0, len(contents))
 	for _, cont := range contents {
-		if !s.CM.contentInStagingZone(c.Request().Context(), cont) {
-			ec := expandedContent{
-				Content: cont,
-			}
-			if cont.Aggregate {
-				if err := s.DB.Model(util.Content{}).Where("aggregated_in = ?", cont.ID).Count(&ec.AggregatedFiles).Error; err != nil {
-					return err
-				}
-
-			}
-			out = append(out, ec)
+		ec := expandedContent{
+			Content: cont,
 		}
+		if cont.Aggregate {
+			if err := s.DB.Model(util.Content{}).Where("aggregated_in = ?", cont.ID).Count(&ec.AggregatedFiles).Error; err != nil {
+				return err
+			}
+
+		}
+		out = append(out, ec)
 	}
 
 	return c.JSON(http.StatusOK, out)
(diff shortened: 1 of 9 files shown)


handlers.go
```diff
@@ -409,7 +409,7 @@ func (s *Server) handleStats(c echo.Context, u *util.User) error {
 	}
 
 	var contents []util.Content
-	if err := s.DB.Limit(limit).Offset(offset).Order("created_at desc").Find(&contents, "user_id = ? and active", u.ID).Error; err != nil {
+	if err := s.DB.Limit(limit).Offset(offset).Order("created_at desc").Find(&contents, "user_id = ?", u.ID).Error; err != nil {
```
alvin-reyes, line 412: @kelindi can you create a separate table for active and inactive contents for the user? Create a tab here for active and inactive

image

replication.go
```diff
@@ -330,14 +330,13 @@ func NewContentManager(db *gorm.DB, api api.Gateway, fc *filclient.FilClient, tb
 		Replication:                  cfg.Replication,
 		tracer:                       otel.Tracer("replicator"),
 		DisableFilecoinStorage:       cfg.DisableFilecoinStorage,
-		IncomingRPCMessages:          make(chan *drpc.Message),
+		IncomingRPCMessages:          make(chan *drpc.Message, 200000),
```
alvin-reyes, line 333: Any reason why it's fixed as `200000`? Can we parameterize this and use `200000` by default.

shuttle.go
```diff
@@ -92,7 +92,7 @@ func (cm *ContentManager) registerShuttleConnection(handle string, hello *drpc.H
 		address:               hello.Address,
 		addrInfo:              hello.AddrInfo,
 		hostname:              hello.Host,
-		cmds:                  make(chan *drpc.Command, 32),
+		cmds:                  make(chan *drpc.Command, 100000),
```
alvin-reyes, line 95: Same as comment above.

replication.go
en0ma: no reason actually, I will move it to be configurable, that way we can play around a variable size

handlers.go
gmelodie, line 412: Do we not want the queried `contents` to be active here?

gmelodie (approved):

handlers.go
en0ma: no, that is the point of the change - return content for all states

en0ma:
@alvin-reyes can I get approval for this, so I can finalize https://github.com/application-research/estuary/pull/444

alvin-reyes (approved):
LGTM

PR #472, opened 2022-10-18
Pin improvements rc

snissn:
Two fixes/band aids are added here for the pin manager along with working tests

1. A Set exists in front of the work manager so that duplicate work is not added to the queue if it is sent to the shuttle multiple times
2. The in memory Deque is replaced with a leveldb based Deque so that if the Deque is very large it does not overly use too much system memory
3. There are working tests that confirm the features are valid and correct

Want to run this branch on a node and see if performance is valid.


en0ma:
Finally reviewed this. great work @snissn

snissn:
thanks! will ping on slack regarding merging and testing

jlogelin (approved):
This looks great @snissn - assuming that you've ran the nicely written unit test you provided here :) Thank you for your contribution to the improved pinning.

alvin-reyes (approved):
Great work! LGTM.

PR #476, opened 2022-10-19
Add /admin/fil-address endpoint for when account has no balance yet

neelvirdy:
**Problem**: When setting up a new estuary node, you can’t find your FIL address through the estuary-www UI because the /admin/balance endpoint fails until the address has some balance. This can cause confusion if you miss the FIL address in the CLI output, making it trickier to get your account funded and start using your estuary node.

**Proposed Solution**: Add a dedicated /admin/fil-address endpoint so we can always display the FIL address in the estuary-www Balance page.

**Note**: Will follow-up with an estuary-www PR to consume this endpoint. (Edit: See https://github.com/application-research/estuary-www/pull/81)

This is what /admin/balance throws in this initial state:
`{"error":{"code":500,"reason":"Internal Server Error","details":"resolution lookup failed (<address>): resolve address <address>: actor not found"}}`

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -245,6 +245,7 @@ func (s *Server) ServeAPI() error {
 
 	admin := e.Group("/admin")
 	admin.Use(s.AuthRequired(util.PermLevelAdmin))
+	admin.GET("/fil-address", s.handleAdminFilAddress)
 	admin.GET("/balance", s.handleAdminBalance)
 	admin.POST("/add-escrow/:amt", s.handleAdminAddEscrow)
 	admin.GET("/dealstats", s.handleDealStats)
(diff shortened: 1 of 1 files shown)


jlogelin (approved):
LGTM

alvin-reyes (approved):
Looks good. If possible, please add some verification tests. E.i show the response output.

neelvirdy:
@alvin-reyes Want to make sure I'm interpreting verification test correctly - I didn't see a place to write tests for handler logic in the project, but I was able to test with the corresponding estuary-www change in the linked PR (https://github.com/application-research/estuary-www/pull/81)

Here's a screenshot of the network request and its response appearing as expected:

image

image

Is this sufficient or did you have something more robust in mind, like what's suggested here https://echo.labstack.com/guide/testing/?

en0ma:
this is a nice feature @neelvirdy

PR #479, opened 2022-10-20
Track staging zone readiness and readiness reason

neelvirdy:
Problem
---
It can be unclear why staging zones are not turning into deals and how to get them unstuck. The logic used for staging zone readiness is buried in code and that is the only place it's discoverable. This can be especially frustrating for new users and deter them from following through on using estuary. 

Solution
---
Expose a readiness flag and readiness reason on the staging zone struct, available via /content/staging-zones. This will be displayed on estuary-www's Staging page so the end user can understand how we're evaluating their staging zone.

Example
---
image

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -143,6 +143,11 @@ func (cm *ContentManager) isInflight(c cid.Cid) bool {
 	return ok && v > 0
 }
 
+type stagingZoneReadiness struct {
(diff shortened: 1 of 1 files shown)


en0ma (approved):
LGTM - Thanks @neelvirdy

Issue #424, opened 2022-09-02
RESTful API Request For Comments


gmelodie:
I think it's about time we redesign the API endpoints. In terms of actual code that should be easy enough (changing routes, swagger annotations and possibly some arguments). I'm thinking we can change incrementally:
1. update collection endpoints
- /collections
  - GET: list all collections for current user
  - POST: create new collections (body: `collectionName`, `collection description`)
- /collections/:coluuid
  - GET: list all files in this collection (query param: `dir` to filter for files only in this dir)
  - POST: ???
  - DELETE: remove file from collection
  - UPDATE: change name/description (body: `collectionName` or `collection description`) 
2. update **pinning** endpoints
 - TODO
3. update **content** endpoints
- `/content/`
  - GET: list all contents for current user  
  - POST: upload file/content 
    - optional query param: `type` (`ipfs` for `ipfs-add`, or `car` for `add-car`)
    - optional query param: `coluuid` to include this file in a collection (in the `/` path by default)
    - optional query param: `dir` to include this file in a specific dir of a collection (`coluuid` must be present)
- `/content/:contentID`
  - GET: download this content
  - DELETE: remove this content
- TODO the rest
5. update **deal** endpoints
 - TODO
6. update the rest
 - TODO
 
 Other ideas:
 - change `/api-keys/...` to `/keys/...`
 - `/content/by-cid` needs to be replaced by something more RESTful

elijaharita:
this is my favorite type of update, looks great so far!

should `/content` not be `/contents`? i guess this is debatable since content _can_ be used as an uncountable noun in some contexts, but to me this usage of the term seems like it should be plural countable noun form since each content is a unique clearly defined object. maybe that's too pedantic, lol.

what is UPDATE btw?

jimmylee:
Looks like this was resolved in Slack
<img width="479" alt="Screen Shot 2022-09-06 at 9 08 48 PM" src="https://user-images.githubusercontent.com/310223/188786737-89ed75fe-05b3-4131-bd9b-9e6f5cc79182.png">

gmelodie:
@anjor as soon as we got the new API design, I'll create a new branch and a bunch of issues that people can grab. When all issues are done, we merge the api branch to dev and do testing. Then release.

neelvirdy:
few thoughts:
[...]

PR #485, opened 2022-10-23
create a v1 prefix for the api

snissn:
create a v1 prefix for the api i, also install all of the v1 api endpoints to the router without the v1 prefix in a way that can easily be deprecated in the future

testing it looks correct--

```
% curl -X POST http://localhost:3004/user/api-keys -H "Authorization: Bearer ESTxx0ARY" -H "Accept: application/json" 

{"token":"ESTxxARY","expiry":"2022-11-22T22:25:03.73497Z"} 

% curl -X POST http://localhost:3004/v1/user/api-keys -H "Authorization: Bearer ESTxxRY" -H "Accept: application/json" 

{"token":"ESTxxRY","expiry":"2022-11-22T22:25:12.289277Z"}
```

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -80,7 +80,7 @@ import (
 )
 
 // @title Estuary API
-// @version 0.0.0
+// @version 1.0.0
 // @description This is the API for the Estuary application.
 // @termsOfService http://estuary.tech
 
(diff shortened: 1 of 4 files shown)


handlers.go
```diff
@@ -91,23 +91,24 @@ import (
...
-// @BasePath  /
+// @BasePath  /v1/
 // @securityDefinitions.Bearer
 // @securityDefinitions.Bearer.type apiKey
 // @securityDefinitions.Bearer.in header
 // @securityDefinitions.Bearer.name Authorization
 func (s *Server) ServeAPI() error {
-	e := echo.New()
-	e.Binder = new(util.Binder)
+	myecho := echo.New()
```
alvin-reyes, line 100: can we rename this to `echoEngine` instead or any other name.

handlers.go
```diff
@@ -317,7 +318,23 @@ func (s *Server) ServeAPI() error {
...
+
+	addRoutesWithNoPrefix(myecho)
+
+	return myecho.Start(s.cfg.ApiListen)
+}
+
+func addRoutesWithNoPrefix(myecho *echo.Echo) {
+	for _, value := range myecho.Routes() {
+		var newpath string
+		if strings.HasPrefix(value.Path, "/v1/") {
```
alvin-reyes, line 330: if we introduce `v2`, this means we would need to add a condition here right? I suggest we add the following on the docs: 

1 - adding a new version programmatically. Add an instruction that a new version requires a new group example `v2`.
2 - how to separate the new version routes to a new file instead of putting them all in one handler.go.

snissn: I think when we do V2 we'll want v2 to be in it's own separate file. I thought that was what you had suggested. So when we do v2 this whole handlers.go will be the v1 handler and we'll delete the addRoutesWithNoPrefix function and it's call and the v1 or v2 prefix will be required. Does that sound right?

handlers.go
snissn: pushed a commit with the rename, thanks for catching that!

gmelodie:
I don't think we want to merge this before we do the API refactor. We should have a plan for the new API, then we can implement it part by part using the V1 prefix. These are the steps I think we should follow:
1. agree on a new api design in https://github.com/application-research/estuary/issues/424
2. create issues for each part of the api (all already using v1 as a prefix) so that people can work async on this
3. release it when all issues are fixed and merged

snissn:
https://github.com/application-research/estuary/pull/596/files is better - deleting this PR cc @en0ma

PR #487, opened 2022-10-24
Dev fix

jlogelin:


alvin-reyes (approved):
LGTM

replication.go
```diff
@@ -1185,6 +1185,10 @@ func (cm *ContentManager) SetDataTransferStartedOrFinished(ctx context.Context,
 		return err
 	}
 
+	if chanst == nil {
```
en0ma, line 1188: @jlogelin @alvin-reyes this should not be the case, this state should exist. I am working on a task to always have a remote state, the current implementation is not sufficient

shuttle.go
```diff
@@ -392,7 +393,7 @@ func (cm *ContentManager) handleRpcTransferStatus(ctx context.Context, handle st
 		}
 
 		param.State = &filclient.ChannelState{
-			Status:  param.State.Status,
+			Status:  datatransfer.Failed,
```
en0ma, line 396: @alvin-reyes @jlogelin why this change? I used `state.status` because a failed state could be either `canceled` or `failed` - I wanted to retain the actual failed status.

Issue #343, opened 2022-07-21
Stop storing API Tokens in Plain Text


kelindi:
API Tokens should only be visible upon creation, and should be hashed in the Database otherwise.

jimmylee:
Thanks for reporting this, its a good idea.

neelvirdy:
Started looking into this. The approach i'm starting with is the following:
### Phase 1 (estuary PR) [mostly complete]:
- Add TokenHash and Label to auth token DB table and HTTP responses for api-key endpoints. Populate them for all new user keys. Label will be an empty string for all manually generated keys until PR 2.
- Change api-key queries to check on TokenHash when present, and fallback to Token otherwise. This ensures ongoing tokens remain active. (This could also happen in PR 3 if preferred)
### Phase 2 (estuary-www PR):
- Require a label to be provided when creating a new key.
- Display Label instead of Token in keys UI, if present. Otherwise, continue displaying raw Token.
- Display a modal window with the new key when created, communicating that it must be stored on the user's end since it will not be recoverable once the modal is closed. This isn't technically true until Phase 3 is live.
### Phase 3 (estuary PR):
- Omit Token field for all new writes to the auth token DB table. 
### Phase 4 (estuary-www PR):
- Display a warning indicator next to any token that is stored in the DB saying the user should revoke and replace the token.

Note: Any tokens generated between Phase 2 and Phase 3 will end up receiving a warning indicator to be replaced, despite the user getting the UX of the completed workflow at generation time.

@kelindi @jimmylee let me know if you have any feedback on this!

en0ma:
@neelvirdy I updated your checklist to make this easy to track

Issue #471, opened 2022-10-18
Add a feature to tag an API key to a "custom" name


alvin-reyes:
**Is your feature request related to a problem? Please describe.**
No. This is general UX improvement.

**Describe the solution you'd like**
We need to add a way for the user to associate a name to an API key.
Changes will include:

1. name field on the auth_token table
2. name field on the API key generation (we should just generate a random name for each key and let the user either use it or define their own
3. the name is just an identifier so we shouldn't treat it as a lookup for an API key.

**Describe alternatives you've considered**
N/A

**Additional context**
This will require changes on the backend (estuary) and frontend (www)

sasicodes:
+1

PR #492, opened 2022-10-26
Add TokenHash and Label to AuthToken, allow revoke by hash

neelvirdy:
Implements Phase 1/4 of https://github.com/application-research/estuary/issues/343, which tracks the switch to only store hashes of API Tokens in DB.

The end-user experiences no change in behavior from this PR, but newly created tokens will have a TokenHash and Label stored in DB. The Label is an arbitrary string that does not need to be unique, but allows the user to track the keys that are in use and know which to revoke when needed.

Allowing revoke by hash is necessary to ensure keys can still be revoked after we stop storing the raw token itself.

For info on follow-up PRs to complete the switch, see the issue linked above.

---

Demonstrating creating a new key with a label, as well as revoking keys by hash and by key. The listed tokens with the censored token are the keys that have no token value stored in the db, and are being revoked via their hash.

https://user-images.githubusercontent.com/2850013/200372130-89420bc0-91fe-404b-bff0-bbb5fe743222.mov

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -3071,16 +3071,8 @@ func (s *Server) handleRegisterUser(c echo.Context) error {
 		}
 	}
 
-	token := "EST" + uuid.New().String() + "ARY"
-	authToken := &util.AuthToken{
-		Token:     token,
-		TokenHash: util.GetTokenHash(token),
-		Label:     TOKEN_LABEL_ON_REGISTER,
-		User:      newUser.ID,
-		Expiry:    time.Now().Add(constants.TokenExpiryDurationRegister),
-	}
-
-	if err := s.DB.Create(authToken).Error; err != nil {
+	authToken, err := s.newAuthTokenForUser(newUser, time.Now().Add(constants.TokenExpiryDurationRegister), nil, TOKEN_LABEL_ON_REGISTER)
+	if err != nil {
 		return err
 	}
 
--- a/handlers.go
+++ b/handlers.go
@@ -3008,6 +3008,8 @@ type registerBody struct {
 	InviteCode string `json:"inviteCode"`
 }
 
+const TOKEN_LABEL_ON_REGISTER = "on-register"
+
 func (s *Server) handleRegisterUser(c echo.Context) error {
 	var reg registerBody
 	if err := c.Bind(&reg); err != nil {
@@ -3073,7 +3075,7 @@ func (s *Server) handleRegisterUser(c echo.Context) error {
 	authToken := &util.AuthToken{
 		Token:     token,
 		TokenHash: util.GetTokenHash(token),
-		Label:     "on-register",
+		Label:     TOKEN_LABEL_ON_REGISTER,
 		User:      newUser.ID,
 		Expiry:    time.Now().Add(constants.TokenExpiryDurationRegister),
 	}
@@ -3102,6 +3104,8 @@ type loginResponse struct {
 	Token  string    `json:"token"`
 	Expiry time.Time `json:"expiry"`
 }
+
+const TOKEN_LABEL_ON_LOGIN = "on-login"
 
 func (s *Server) handleLoginUser(c echo.Context) error {
 	var body loginBody
@@ -3142,7 +3146,7 @@ func (s *Server) handleLoginUser(c echo.Context) error {
 		}
 	}
 
-	authToken, err := s.newAuthTokenForUser(&user, time.Now().Add(constants.TokenExpiryDurationLogin), nil, "login")
+	authToken, err := s.newAuthTokenForUser(&user, time.Now().Add(constants.TokenExpiryDurationLogin), nil, TOKEN_LABEL_ON_LOGIN)
 	if err != nil {
 		return err
 	}
--- a/handlers.go
+++ b/handlers.go
@@ -3075,7 +3075,7 @@ func (s *Server) handleRegisterUser(c echo.Context) error {
 		TokenHash: util.GetTokenHash(token),
 		Label:     "on-register",
 		User:      newUser.ID,
-		Expiry:    time.Now().Add(time.Hour * 24 * 7),
+		Expiry:    time.Now().Add(constants.TokenExpiryDurationRegister),
 	}
 
 	if err := s.DB.Create(authToken).Error; err != nil {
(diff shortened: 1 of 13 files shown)


en0ma:
@neelvirdy can the label address this requirement https://github.com/application-research/estuary/issues/471 ?

handlers.go
```diff
@@ -138,7 +138,7 @@ func (s *Server) ServeAPI() error {
 	user.Use(s.AuthRequired(util.PermLevelUser))
 	user.GET("/api-keys", withUser(s.handleUserGetApiKeys))
 	user.POST("/api-keys", withUser(s.handleUserCreateApiKey))
-	user.DELETE("/api-keys/:key", withUser(s.handleUserRevokeApiKey))
+	user.DELETE("/api-keys/:key-or-hash", withUser(s.handleUserRevokeApiKey))
```
en0ma, line 141: nitpick: can we stay consistent `key_or_hash`

handlers.go
```diff
@@ -2881,10 +2882,13 @@ func (s *Server) handleRegisterUser(c echo.Context) error {
 		}
 	}
 
+	token := "EST" + uuid.New().String() + "ARY"
 	authToken := &AuthToken{
```
en0ma, line 2886: I think we have to stop creating API keys on user registration, and instead let the users create a key after registering. It will improve the UX and allow them properly label their keys

handlers.go
```diff
@@ -2938,7 +2942,7 @@ func (s *Server) handleLoginUser(c echo.Context) error {
 		}
 	}
 
-	authToken, err := s.newAuthTokenForUser(&user, time.Now().Add(time.Hour*24*30), nil)
+	authToken, err := s.newAuthTokenForUser(&user, time.Now().Add(time.Hour*24*30), nil, "login")
```
en0ma, line 2945: @alvin-reyes why do we create new API keys every time a user logs in?

handlers.go
neelvirdy: :+1:

neelvirdy:
> @neelvirdy can the label address this requirement #471 ?

@en0ma yep! didn't see that issue till now but it seems identical except these 2 differences:
1. Label vs name for field name - i have no strong preference with which way we go. Label is slightly more intuitive to me but clearly not for everyone! I don't see any convention from a quick google search. Open to adjusting it if someone shouts
2. I intend to make the label a required input when generating a key, so the user is encouraged to provide a meaningful label and remember what the key is for, rather than auto-generating a random label and relying on the user to do their due diligence proactively to have good key management.

handlers.go
neelvirdy: i think this is necessary since the key management endpoints require a bearer token. this seems to give the newly registered user a token for their browser session upon successful register.

on cleanliness - it seems we could reduce code duplication here and have all new tokens be created the newAuthTokenForUser method. happy to do that cleanup in a separate PR

handlers.go
neelvirdy: i believe this only happens if the user's cookie no longer has a token or has an expired token, so they get the login screen and have to login with their credentials. then upon successful login they get a new token and shouldn't hit the login endpoints anymore as long as their token is valid

en0ma: ok

en0ma:
> 2\. I intend to make the label a required input when generating a key, so the user is encouraged to provide a meaningful label and remember what the key is for, rather than auto-generating a random label and relying on the user to do their due diligence proactively to have good key management.

If this is done, I think it will completely address that issue. I am ok with using "label" too.

en0ma:
@alvin-reyes can you have a look at this PR

handlers.go
```diff
@@ -136,7 +136,7 @@ func (s *Server) ServeAPI() error {
 	user.Use(s.AuthRequired(util.PermLevelUser))
 	user.GET("/api-keys", withUser(s.handleUserGetApiKeys))
 	user.POST("/api-keys", withUser(s.handleUserCreateApiKey))
-	user.DELETE("/api-keys/:key", withUser(s.handleUserRevokeApiKey))
+	user.DELETE("/api-keys/:key_or_hash", withUser(s.handleUserRevokeApiKey))
```
alvin-reyes, line 139: This is a breaking change. Retain the old endpoint and use the same handler.

alvin-reyes: I realized that this is not a breaking change. This looks good to me.

handlers.go
```diff
@@ -3068,10 +3069,13 @@ func (s *Server) handleRegisterUser(c echo.Context) error {
...
 	}
 
+	token := "EST" + uuid.New().String() + "ARY"
 	authToken := &util.AuthToken{
-		Token:  "EST" + uuid.New().String() + "ARY",
-		User:   newUser.ID,
-		Expiry: time.Now().Add(time.Hour * 24 * 7),
+		Token:     token,
+		TokenHash: util.GetTokenHash(token),
+		Label:     "on-register",
```
alvin-reyes, line 3076: Are these default labels? We should put it on a `const` enumerator.

handlers.go
```diff
@@ -3366,9 +3374,8 @@ type getApiKeysResp struct {
 // @Param        key path string true "Key"
```
alvin-reyes, line 3374: Let’s add a comment here that this endpoint will completely delete the key and will not be recoverable. We can add that to the description part of the swagger doc.

main.go
```diff
@@ -504,10 +504,13 @@ func main() {
...
+				token := "EST" + uuid.New().String() + "ARY"
 				authToken := &util.AuthToken{
-					Token:  "EST" + uuid.New().String() + "ARY",
-					User:   newUser.ID,
-					Expiry: time.Now().Add(time.Hour * 24 * 365),
+					Token:     token,
+					TokenHash: util.GetTokenHash(token),
+					Label:     "admin",
+					User:      newUser.ID,
+					Expiry:    time.Now().Add(time.Hour * 24 * 365),
```
alvin-reyes, line 513: Let’s put the expiry value on the constants.go and use it across the estuary code base.

util/users.go
```diff
@@ -30,7 +30,9 @@ func (u *User) FlagSplitContent() bool {
 
 type AuthToken struct {
 	gorm.Model
-	Token      string `gorm:"unique"`
+	Token      string
```
alvin-reyes, line 33: the token should still be `gorm:”unique”`.

alvin-reyes (commented):
Great work @neelvirdy! 

Can you add some verification tests and post it in this PR? Just add a new section on the description part 

```
## Verification 
### revoke user by hash
<screen shot of the UI or the cli>
### revoke user by key
<screen shot of the UI or the cli>
```

neelvirdy:
Demonstrating creating a new key with a label, as well as revoking keys by hash and by key. The listed tokens with the censored token are the keys that have no token value stored in the db, and are being revoked via their hash.

https://user-images.githubusercontent.com/2850013/200372130-89420bc0-91fe-404b-bff0-bbb5fe743222.mov

alvin-reyes (approved):
LGTM

PR #495, opened 2022-10-29
fix: only send pin completion message after blocks are completely copied

en0ma:
When files are uploaded to a shuttle, it sends a pin completion RPC message to the primary even before the staging blocks are copied to the main blockstore. 

The Primary immediately starts deal processing and request for commp from the shuttle, given the copy is not completed, the results in commp error.

This PR addresses this by making sure the RPC message is only sent after the copy is done


alvin-reyes (approved):
LGTM

PR #505, opened 2022-11-04
Absolute paths when rendering gateway dir hrefs

neelvirdy:
When rendering a dir via the estuary gateway, renders hrefs as `/gw/ipfs/<cid>/<child-name>` instead of `./<child-name>`.

This fixes browsing cids like the following: 
- https://api.estuary.tech/gw/ipfs/bafybeihadprcvcygjv7znlpdmrekbu6moyykceg4gb2jhq2obpjf4p6mbm (no trailing slash case)
- https://api.estuary.tech/gw/ipfs/bafybeihadprcvcygjv7znlpdmrekbu6moyykceg4gb2jhq2obpjf4p6mbm/outputs/ (nested paths case)

This also mirrors the behavior of the kubo gateway implementation (see https://strn.pl/ipfs/bafybeihadprcvcygjv7znlpdmrekbu6moyykceg4gb2jhq2obpjf4p6mbm as a reference), hrefs are rendered as `/ipfs/<cid>/<child-name>`.

### DIFF
--- a/util/gateway/gateway.go
+++ b/util/gateway/gateway.go
@@ -182,10 +182,9 @@ func (gw *GatewayHandler) serveUnixfsDir(ctx context.Context, n mdagipld.Node, w
 	fmt.Fprintf(w, "<html><body><ul>")
 
 	requestURI, err := url.ParseRequestURI(req.RequestURI)
-	originalURLPath := requestURI.Path
 
 	if err := dir.ForEachLink(ctx, func(lnk *mdagipld.Link) error {
-		href := gopath.Join(originalURLPath, lnk.Name)
+		href := gopath.Join(requestURI.Path, lnk.Name)
 		fmt.Fprintf(w, "<li><a href=\"%s\">%s</a></li>", href, lnk.Name)
 		return nil
 	}); err != nil {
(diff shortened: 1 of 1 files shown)


alvin-reyes (commented):
can we beautify the list on this PR? I think we should just list it as if it’s a directory or a file list.

https://codemyui.com/directory-list-with-collapsible-nested-folders-and-files/ 

Use the above reference and stretch the body of the list to 100% with 10px padding.

alvin-reyes:
I spoke to Andy about building a proper UI (that we need to convert to html/js/css) for this. We should try getting that in with this PR.

en0ma:
> I spoke to Andy about building a proper UI (that we need to convert to html/js/css) for this. We should try getting that in with this PR.

@alvin-reyes let's get this PR in to address the bug and @neelvirdy can work on a follow-up PR for the UI upgrade

jlogelin (approved):
LGTM

neelvirdy:
Agreed with merging this PR before tracking the UI changes. I'd also like to have a conversation about whether we should invest in our own gateway front end (and gw implementation altogether) or just leverage the kubo gateway implementation.

Issue #506, opened 2022-11-05
Bearer API prefix is confusing


snissn:
**Is your feature request related to a problem? Please describe.**
I find it really confusing, especially to new users that the API KEY needs to have "Bearer " prefix when making requests

**Describe the solution you'd like**
Short patch of code that detects the api key is of the form "EST-xxx-ARY" instead of "Bearer EST-xxx-ARY" and just accepts either

**Describe alternatives you've considered**
Alternative would be to remove the Bearer at all - but that would break things for users

**Additional context**
I have a PR ready i'll link in

snissn:
PR - https://github.com/application-research/estuary/pull/507

whyrusleeping:
This is HTTP standard stuff, The authorization header is intended to have a specific format, see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

PR #507, opened 2022-11-05
add Bearer prefix to auth header if missing

snissn:
corresponding issue 

https://github.com/application-research/estuary/issues/506


alvin-reyes (commented):
Please add a verification test on the description of this PR.

neelvirdy (changes requested):
echoing the slack thread that we should stick with the http standards and keep the Bearer prefix https://filecoinproject.slack.com/archives/C016APFREQK/p1667608204892989

PR #509, opened 2022-11-07
Fix links in gateway

jlogelin:


en0ma:
@jlogelin see https://github.com/application-research/estuary/pull/505

PR #512, opened 2022-11-07
Change some per-provider logs to warn where appropriate

neelvirdy:
Failures with individual SPs in dealmaking should be warns instead of errors, since it doesnt mean that estuary itself is in an error state.


alvin-reyes (approved):
LGTM

PR #513, opened 2022-11-08
Update replace pin swagger annotations

gmelodie:
We forgot to mention some parameters on the swagger annotations for replacing a pin

Reminder: the `docs: update swagger docs` commit is a post commit hook, not me


alvin-reyes (approved):
LGTM

PR #514, opened 2022-11-08
docs: Update annotations for Pinning API endpoints

simonwo:
This is necessary because otherwise the `estuary-clients` package won't generate the correct Go types to unmarshal responses into, hence the client doesn't work.


alvin-reyes (approved):
LGTM

PR #515, opened 2022-11-08
Fix collections add content

gmelodie:
The `POST /collections/:coluuid` had a bug where it would not correctly parse the `[]uint` list on the request body. This fixes it.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -3626,7 +3626,7 @@ func (s *Server) handleListCollections(c echo.Context, u *util.User) error {
 // @Tags         collections
 // @Accept       json
 // @Produce      json
-// @Param        coluuid     path      string  true  "coluuid"
+// @Param        coluuid     path      string  true  "Collection UUID"
 // @Param        contentIDs  body      []uint  true  "Content IDs to add to collection"
 // @Success      200         {object}  string
 // @Failure      400  {object}  util.HttpError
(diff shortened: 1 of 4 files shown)


handlers.go
```diff
@@ -3635,11 +3639,35 @@ func (s *Server) handleListCollections(c echo.Context, u *util.User) error {
 func (s *Server) handleAddContentsToCollection(c echo.Context, u *util.User) error {
 	coluuid := c.Param("coluuid")
 
-	var contentIDs []uint
-	if err := c.Bind(&contentIDs); err != nil {
+	// we accept both {"contentids": [1, 2]} and [1, 2] as json payloads
```
en0ma, line 3642: @gmelodie why do we accept 2 formats, is it due to backward compatibility?

gmelodie: See [Slack discussion](https://filecoinproject.slack.com/archives/C016APFREQK/p1668026579464679?thread_ts=1668005715.880389&cid=C016APFREQK)

gmelodie: @alvin-reyes weight in

alvin-reyes: Yes. backward compatibility. We want to support the new json format without negatively impacting any existing users of this endpoint.

alvin-reyes (approved):
If you can, please add some verification steps here before merging.

LGTM.

Issue #516, opened 2022-11-09
Improve API Response error message with invalid API Key


snissn:
**Describe the bug**
User sends api request with http header with `Authorization: APIKEY` instead of `Authorization: Bearer APIKEY` they receive an error message that says the `

**To Reproduce**
```
curl -X 'GET' \
  'https://api.estuary.tech/collections/' \
  -H 'accept: application/json' \
  -H 'Authorization: APIKEY'```


**Expected behavior**
{
  "error": {
    "code": 401,
    "reason": "ERR_INVALID_AUTH",
    "details": "Unsupported authorization scheme: Bearer is a required prefix, also make sure you have a valid and unexpired API key. An API key is required to use this endpoint."
  }
}

**Actual behavior**
Response body
```
{
  "error": {
    "code": 401,
    "reason": "ERR_INVALID_AUTH",
    "details": "invalid api key was specified"
  }
}
```

**Additional context**
discussion on slack https://filecoinproject.slack.com/archives/C016APFREQK/p1667608204892989


please have @whyrusleeping @jimmylee and/or @neelvirdy chime in and approve before we merge

en0ma:
closing this has it has been addressed

PR #517, opened 2022-11-09
update response message with Cakes suggested text

snissn:
related issue https://github.com/application-research/estuary/issues/516
via our discussion on slack

Please only merge if the reviewers approve.

Thanks!

### DIFF
--- a/util/http.go
+++ b/util/http.go
@@ -111,7 +111,7 @@ func ExtractAuth(c echo.Context) (string, error) {
 		return "", &HttpError{
 			Code:    http.StatusUnauthorized,
 			Reason:  ERR_INVALID_AUTH,
-			Details: "invalid api key was specified",
+			Details: "Unsupported authorization scheme: Bearer is a required prefix, also make sure you have a valid and unexpired API key. An API key is required to use this endpoint.",
 		}
 	}
 
@@ -119,7 +119,7 @@ func ExtractAuth(c echo.Context) (string, error) {
 		return "", &HttpError{
 			Code:    http.StatusUnauthorized,
 			Reason:  ERR_AUTH_MISSING_BEARER,
-			Details: "invalid api key was specified",
+			Details: "Unsupported authorization scheme: Bearer is a required prefix, also make sure you have a valid and unexpired API key. An API key is required to use this endpoint.",
 		}
 	}
 	return parts[1], nil
--- a/util/http.go
+++ b/util/http.go
@@ -51,8 +51,8 @@ const (
 
 const (
 	ERR_AUTH_MISSING_DETAILS        = "no api key was specified"
-	ERR_AUTH_MISSING_BEARER_DETAILS = "Unsupported authorization scheme: Bearer is a required prefix. The Authorization HTTP Header should be in the format \"Authorization: Bearer ESTxxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxxARY\"."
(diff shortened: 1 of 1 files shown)


util/http.go
```diff
@@ -111,15 +111,15 @@ func ExtractAuth(c echo.Context) (string, error) {
 		return "", &HttpError{
 			Code:    http.StatusUnauthorized,
 			Reason:  ERR_INVALID_AUTH,
-			Details: "invalid api key was specified",
+			Details: "Unsupported authorization scheme: Bearer is a required prefix, also make sure you have a valid and unexpired API key. An API key is required to use this endpoint.",
 		}
 	}
 
 	if parts[0] != "Bearer" {
```
neelvirdy, line 118: can we put this check before the one above and write the condition as `if len(parts) > 0 & parts[0] != "Bearer"`? so they get the ERR_AUTH_MISSING_BEARER reason when they don't provide bearer, instead of ERR_INVALID_AUTH. then you can make the language more targeted and say "Bearer prefix is required but missing" for that case

neelvirdy (changes requested):

util/http.go
```diff
@@ -111,15 +111,15 @@ func ExtractAuth(c echo.Context) (string, error) {
...
 		return "", &HttpError{
 			Code:    http.StatusUnauthorized,
 			Reason:  ERR_AUTH_MISSING_BEARER,
-			Details: "invalid api key was specified",
+			Details: "Unsupported authorization scheme: Bearer is a required prefix, also make sure you have a valid and unexpired API key. An API key is required to use this endpoint.",
```
alvin-reyes, line 122: Can we just put all hardcoded strings in a constant? This so we we can edit or add the message in a single source and call the same constant variable instead of repeatedly hardcoding the same text over.

alvin-reyes (changes requested):

util/http.go
alvin-reyes: And please merge this against `dev`.

snissn:
@neelvirdy @alvin-reyes 

I believe i've incorporated your requests into a new commit. I also tested the code and see the expected result:

```
 ~ % curl -X 'GET' \
  'http://localhost:3004/collections/' \
  -H 'accept: application/json' \
  -H 'Authorization: APIKEY'``
{"error":{"code":401,"reason":"ERR_AUTH_MISSING_BEARER","details":"Unsupported authorization scheme: Bearer is a required prefix. The Authorization HTTP Header should be in the format \"Authorization: Bearer YOUR_API_KEY_HERE\"."}}

% curl -X 'GET' \
  'http://localhost:3004/collections/' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer'``
{"error":{"code":401,"reason":"ERR_INVALID_AUTH","details":"Invalid Auth: An Invalid API Key was specified. The Authorization HTTP Header should be in the format \"Authorization: Bearer YOUR_API_KEY_HERE\". You have the Bearer prefix but your API Key is empty or missing."}}
```

I think that @whyrusleeping's main concerns on slack are met as we're not changing anything around how we use the HTTP spec/protocol, we're simply updating error messages. And Cake seems to approve the improvement. 

If @neelvirdy and @alvin-reyes both agree let's merge this. Also if there are any other additional blockers or changes i'm very happy to make them

en0ma:
@snissn can you fix the Gosec failing build? Also thanks for doing this, these will greatly improve the auth UX

snissn:
Will do!

alvin-reyes (approved):
LGTM

neelvirdy (approved):

alvin-reyes (dismissed):
LGTM

snissn:
@alvin-reyes updated the PR so that it passes gosec!

alvin-reyes (approved):
LGTM

PR #520, opened 2022-11-11
Fix minor swagger inconsistencies

gmelodie:
I found a couple inconsistencies on our swagger annotations:
- `path` variables not being required
- multiple `@Param body`s (which is not allowed)
- an extra unused annotated variable 
- `/admin/peering` was a subgroup but we don't do that on other admin subgroups like `/admin/cm`
- remove `peering,peers` tags from `/admin/peering` endpoints (cause them to show up duplicated on swagger)

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -2606,7 +2606,7 @@ func (s *Server) handleEstimateDealCost(c echo.Context) error {
 // @Success      200  {object}  string
 // @Failure      400  {object}  util.HttpError
 // @Failure      500  {object}  util.HttpError
-// @Param        miner  path      string  false  "Filter by miner"
+// @Param        miner  path      string  true  "Filter by miner"
 // @Router       /public/miners/failures/{miner} [get]
 func (s *Server) handleGetMinerFailures(c echo.Context) error {
 	maddr, err := address.NewFromString(c.Param("miner"))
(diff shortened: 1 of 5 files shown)


handlers.go
```diff
@@ -278,13 +278,12 @@ func (s *Server) ServeAPI() error {
...
 
 	//	peering
-	adminPeering := admin.Group("/peering")
-	adminPeering.POST("/peers", s.handlePeeringPeersAdd)
-	adminPeering.DELETE("/peers", s.handlePeeringPeersRemove)
-	adminPeering.GET("/peers", s.handlePeeringPeersList)
-	adminPeering.POST("/start", s.handlePeeringStart)
-	adminPeering.POST("/stop", s.handlePeeringStop)
-	adminPeering.GET("/status", s.handlePeeringStatus)
+	admin.POST("/peering/peers", s.handlePeeringPeersAdd)
```
alvin-reyes, line 281: Is there a reason why we needed to remove the group here? It make sense to utilize the grouping in this case since we have multiple functions for a `peering`.

en0ma:
@gmelodie any update on this?

en0ma (approved):
LGTM

en0ma:
@alvin-reyes @snissn @gmelodie let's get this in

PR #523, opened 2022-11-11
Update dependencies, incl. migrate to rcmgr.LimitConfig

neelvirdy:


neelvirdy:
quic-go dropped support for go 1.17 so we'll have to require go 1.18 when this goes through. looks like CI build is failing because it is running on go 1.17

node/modules/lp2p/rcmgr.go
```diff
@@ -19,15 +19,11 @@ import (
...
 
-func NewDefaultLimiter() *rcmgr.BasicLimiter {
-	return rcmgr.NewDefaultLimiter()
-}
-
-func NewResourceManager(limiter *rcmgr.BasicLimiter) (network.ResourceManager, error) {
+func NewResourceManager(limiter *rcmgr.Limiter) (network.ResourceManager, error) {
 	var opts []rcmgr.Option
-	libp2p.SetDefaultServiceLimits(limiter)
+	libp2p.SetDefaultServiceLimits(&rcmgr.DefaultLimits)
```
neelvirdy, line 24: still need to figure out what's going on here with the limit configs. not sure if we're trying to pull libp2p's defaults or use our own

config/node.go
```diff
@@ -23,12 +23,11 @@ type Node struct {
 	WalletDir                 string                `json:"wallet_dir"`
 	ApiURL                    string                `json:"api_url"`
 	Bitswap                   Bitswap               `json:"bitswap"`
-	Limits                    Limits                `json:"limits"`
+	LimitConfig               rcmgr.LimitConfig     `json:"limit_config"`
```
en0ma, line 26: best you call it `limiter`, no need to add the `config` suffix

en0ma: `Limiter  rcmgr.LimitConfig     json:"limiter"`

node/modules/lp2p/rcmgr.go
en0ma: using our own

config/node.go
neelvirdy: Limiter refers to a different type in rcmgr. so instead to keep it consistent with the existing naming i put it back to `Limits rcmgr.ScalingLimitConfig json:"limits"`

config/node.go
```diff
@@ -2,33 +2,27 @@ package config
...
+	NoLimiter                 bool                     `json:"no_limiter"`
+	IndexerURL                string                   `json:"indexer_url"`
+	Blockstore                string                   `json:"blockstore"`
+	WriteLogDir               string                   `json:"write_log_dir"`
+	Libp2pKeyFile             string                   `json:"libp2p_key_file"`
+	DatastoreDir              string                   `json:"datastore_dir"`
+	WalletDir                 string                   `json:"wallet_dir"`
+	ApiURL                    string                   `json:"api_url"`
+	Bitswap                   Bitswap                  `json:"bitswap"`
+	Limits                    rcmgr.ScalingLimitConfig `json:"limits"`
```
neelvirdy, line 26: this is a break to the json structure of node config, if NoLimiter is set to false. from a brief scan it looks like this is hardcoded to true, so the limits codepath doesn't even get executed currently. but if we ever switch this back to true we'd need to ensure the `limits` json blob matches the `ScalingLimitConfig` struct rather than the custom `Limits` struct we had previously - which no longer maps cleanly into the `ScalingLimitConfig` which is why i removed it entirely.

neelvirdy:
going to remake this PR to clean it up a bit

neelvirdy:
Closing in favor of https://github.com/application-research/estuary/pull/529

PR #526, opened 2022-11-14
Add pinning retry worker

en0ma:
This PR replaces the existing `refreshPinQueue` implementation with a pinning monitor/retry worker. This new approach will not only refresh the pin queue when Estuary is restarted, but it will also handle missing pin requests while it runs.

**How it works**

- It runs every X hour interval (default is 1)
- It selects contents that have been created over 1hr that are still in the `pinning` state. So as not to stress the database; it batches the selection of the contents (default is 1000) to retry 

**Verifications**

- I have verified it picks up only contents that are an hour old that are still in `pinning` state
- I have verified it batches the selections correctly (only picks 1000 per loop until no more contents)
- I have verified contents are picked up and properly queued for re-pinning

### DIFF
--- a/config/estuary.go
+++ b/config/estuary.go
@@ -104,8 +104,9 @@ func NewEstuary(appVersion string) *Estuary {
 
 		Pinning: Pinning{
 			RetryWorker: RetryWorker{
-				Interval:   1,
-				BatchLimit: 1000,
+				Interval:               1 * time.Hour, // check every 1hr
+				BatchSelectionLimit:    1000,
+				BatchSelectionDuration: time.Hour * 24 * 30 * 2, // select pins from 60 days ago only
 			},
 		},
 
--- a/config/estuary.go
+++ b/config/estuary.go
@@ -32,6 +32,7 @@ type Estuary struct {
 	StagingBucket          StagingBucket `json:"staging_bucket"`
 	Replication            int           `json:"replication"`
 	RPCMessage             RPCMessage    `json:"rpc_message"`
+	Pinning                Pinning       `json:"pinning"`
 }
 
 func (cfg *Estuary) Load(filename string) error {
@@ -99,6 +100,13 @@ func NewEstuary(appVersion string) *Estuary {
 			MinDealSize:             256 << 20,                                               //0.25 Gib
 			IndividualDealThreshold: int64((abi.PaddedPieceSize(4<<30).Unpadded() * 9) / 10), // 90% of the unpadded data size for a 4GB piece, the 10% gap is to accommodate car file packing overhead, can probably do this better
 			AggregateInterval:       time.Minute * 5,                                         // aggregate staging buckets every 5 minutes
+		},
+
+		Pinning: Pinning{
+			RetryWorker: RetryWorker{
+				Interval:   1,
+				BatchLimit: 1000,
+			},
 		},
 
 		Jaeger: Jaeger{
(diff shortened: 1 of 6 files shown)


ohmpatel1997 (approved):
looks good, thanks for taking this up.

main.go
```diff
@@ -714,6 +714,7 @@ func main() {
 
 		go cm.Run(cctx.Context)                                                 // deal making and deal reconciliation
 		go cm.handleShuttleMessages(cctx.Context, cfg.RPCMessage.QueueHandlers) // register workers/handlers to process shuttle rpc messages from a channel(queue)
+		go cm.RunPinningRetryWorker(cctx.Context)                               // pinning retry worker, re-attempt pinning contents, not yet pinned after a period of time
```
alvin-reyes, line 717: Neat that you separated this.

pinning_retry_worker.go
```diff
@@ -0,0 +1,36 @@
...
+	timer := time.NewTicker(time.Hour * cm.cfg.Pinning.RetryWorker.Interval)
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-timer.C:
+			startContentID := 0
+			for {
+				var contents []util.Content
+				if err := cm.DB.Limit(cm.cfg.Pinning.RetryWorker.BatchLimit).Order("id ASC").Find(&contents, "pinning and not active and not failed and not aggregate and id > ? and created_at > NOW() + INTERVAL '1 HOURS'", startContentID).Error; err != nil {
```
alvin-reyes, line 21: I'm thinking we should add a date here. We should just re-try the last 2 to 3 months of pins. We will need to run another GC routine to clean up old data.

en0ma: ok

config/estuary.go
```diff
@@ -101,6 +102,13 @@ func NewEstuary(appVersion string) *Estuary {
 			AggregateInterval:       time.Minute * 5,                                         // aggregate staging buckets every 5 minutes
 		},
 
+		Pinning: Pinning{
+			RetryWorker: RetryWorker{
+				Interval:   1,
```
gmelodie, line 107: This should be of type `time.Duration` and set to `1*time.Hour` (more descriptive, tells us this is in hours)

gmelodie (dismissed):
Nice!

pinning_retry_worker.go
jcace: `and created_at > NOW() + INTERVAL '1 HOURS'` 

Why is this hard coded to `1 HOURS` ? Wouldn't it need to be dynamic based on `cm.cfg.Pinning.RetryWorker.Interval` ?

en0ma: `cm.cfg.Pinning.RetryWorker.Interval` is how often it should check,, `and created_at > NOW() + INTERVAL '1 HOURS'` is what data it should select at the interval. I will make this dynamic with a default of 2 months back

alvin-reyes (approved):
LGTM. 

I think we should have a separate configuration constants for this limits. I'm thinking we need to have a separate GC process to clean up the data that we don't process anymore.

en0ma:
> LGTM.
> 
> I think we should have a separate configuration constants for this limits. I'm thinking we need to have a separate GC process to clean up the data that we don't process anymore.

we can have another worker for GC. what limits are you talking about?

ohmpatel1997:
@en0ma  @alvin-reyes  can we please migrate the existing stuck pins from last may be 6 months instead of 60 days?

en0ma:
@alvin-reyes I think 6months can work too, let's do this

ohmpatel1997:
> @alvin-reyes I think 6months can work too, let's do this

thanks

ohmpatel1997:
@en0ma @alvin-reyes  just wanted to check if the above is deployed or not?

ohmpatel1997:
Ohk, thanks

PR #441, opened 2022-09-19
feat: authorization/password handler

alvin-reyes:
# Changes

This introduces the following:
- bcrypt for handling users created using CLI.
- proper hashing based on DB dialect (database type). 
- introduce base and base64
- [introduce login (on the frontend) using users created using CLI.](https://github.com/application-research/estuary-www/pull/66)

PR #527, opened 2022-11-14
Chore/drop support for sqlite

en0ma:
This PR drops support for SQLite, it is a breaking change and should be stated in the release note.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,13 +22,11 @@ Requirements:
 To run locally in a 'dev' environment, first run:
 
 ```sh
-./estuary setup --username=<uname> --password=<pword>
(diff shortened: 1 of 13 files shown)


jlogelin:
@en0ma what's the thinking around this? (I like having a local sqlite db to test estuary locally)

alvin-reyes (commented):
@en0ma I think we should not drop support for SQLite since it's really been useful for anyone who wants to do some quick demonstration of Estuary. 

We solved the encoding issue by introducing a DB dialect detection in this PR https://github.com/application-research/estuary/pull/441

IMO, we should make postgres as primary DB but we don't need to drop support for SQLite.

PR #529, opened 2022-11-14
Update dependencies. Requires go 1.18 and BasicLimiter -> ScalingLimitConfig

neelvirdy:
quic-go dropped support for go 1.17 in [0.29.0](https://github.com/lucas-clemente/quic-go/releases/tag/v0.29.0) so we have to require go 1.18 moving forward.

libp2p's rcmgr lib did a migration from BasicLimiter to ScalingLimitConfig in [0.21.0](https://github.com/libp2p/go-libp2p/pull/1648). This hardly effects us since we run with the NoLimiter flag set to true, so we dont even go down the impacted code path. But the migration is done and tested in case we switch. If we have any json-loading of the Node config struct we will need to adapt that wherever that lives to fit ScalingLimitConfig's schema.

Next blocker for getting to latest is filclient: https://github.com/application-research/filclient/blob/master/go.mod#L39-L41. Will track a follow-up to get filclient updated and then continue updating estuary.


en0ma:
> If we have any json-loading of the Node config struct we will need to adapt that wherever that lives to fit ScalingLimitConfig's schema.

We do not

en0ma (dismissed):

config/shuttle.go
```diff
@@ -116,32 +117,39 @@ func NewShuttle(appVersion string) *Shuttle {
...
-					ConnsOutbound: 256,
-					Conns:         512,
-
-					FD: 1024,
+					ConnsInbound:    256,
+					ConnsOutbound:   256,
+					Conns:           512,
+					FD:              1024,
+				},
+				// can be removed after https://github.com/libp2p/go-libp2p/pull/1878 is released
```
alvin-reyes, line 141: Add a `// TODO:`here so we can get back to it once the libp2p PR is merged/released.

go.mod
```diff
@@ -98,7 +95,7 @@ require (
 	github.com/ipfs/go-ipfs v0.13.1
 	github.com/libp2p/go-libp2p-pubsub v0.7.1
 	github.com/pkg/errors v0.9.1
-	github.com/syndtr/goleveldb v1.0.0
+	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
```
alvin-reyes, line 98: Do we need to upgrade the goleveldb? The commit above seems to be from an outdated commit? 

https://github.com/syndtr/goleveldb/commit/2ae1ddf74ef7020251ff1ff0fe8daac21a157761

alvin-reyes (approved):
I ran the build on my local and executed a few regression tests and this seems to work as expected. I have a few minimal feedback but mostly nits. Great work on adding more assertions. 

One thing I'm concerned with this update is if it's somehow impacts `Graphsync` and/or boost transfers. It shouldn't but it be good to test it.

Let's just merge this to dev and see if we can test those out. If it's ok, maybe let's run some graphsync and boost transfers with this bumps? It would be interested to know if this somehow impacts AR/filclient. 

@en0ma

go.mod
neelvirdy: this version gets bumped when i run `go mod tidy`, and if i try to manually put it back to 1.0.0, `go test -v ./...` complains and asks me to run `go mod tidy`. i also tried to `go install github.com/syndtr/goleveldb@v1.0.0` but i get the following:

```
go install github.com/syndtr/goleveldb@v1.0.0
go: github.com/syndtr/goleveldb@v1.0.0: module github.com/syndtr/goleveldb@v1.0.0 found, but does not contain package github.com/syndtr/goleveldb
```

not really sure why this is happening, but figured we could let it be as long as tests are working as expected

neelvirdy: @alvin-reyes goleveldb seems generally unmaintained fwiw. v1.0.0 was from feb 2019 and the latest commit on master was 4 months ago with failing checks. wonder if we should think about migrating off of it or even coming up with our own ownership story for it https://github.com/syndtr/goleveldb/issues/321

alvin-reyes (approved):
LGTM

PR #531, opened 2022-11-15
Added datecreated to the query at the GET /invites endpoint

kelindi:
Datecreated is now returned when querying GET /invites. Implemented this inorder to show analytics on how many invites are sent out each month on estuary-www.


en0ma (approved):

alvin-reyes (dismissed):
LGTM

en0ma (approved):

PR #533, opened 2022-11-16
miner selection improvements - part 1 (extract miner logic into a separate package)

en0ma:
The miner selection improvement will happen in multiple PRs to make code review easy.
This first PR makes no miner selection behavior changes, it basically extracted the miner logic into a separate package. 

A follow-up PR will bring the actual selection behavior changes and it will be easy to code review as there will be less distraction from moving things around

**Verification**
I have verified deals are made, and current miner behavior was unchanged


jcace:
Looks good! Much cleaner with the `miner` code separated out like this.

handlers.go
```diff
@@ -2107,22 +2108,22 @@ type adminStatsResponse struct {
 func (s *Server) handleAdminStats(c echo.Context) error {
 
 	var dealsTotal int64
-	if err := s.DB.Model(&contentDeal{}).Count(&dealsTotal).Error; err != nil {
+	if err := s.DB.Model(&util.ContentDeal{}).Count(&dealsTotal).Error; err != nil {
```
alvin-reyes, line 2111: If we are going to improve this, we should move the objects in a model.go file and reserve the util to utilities objects.

miner/manager.go
```diff
@@ -0,0 +1,184 @@
...
+	_, nrand := pickMinerDist(n)
+
+	out, err := mgr.randomMinerListForDeal(ctx, nrand, pieceSize, exclude, filterByPrice)
+	if err != nil {
+		return nil, err
+	}
+	return mgr.sortedMinersForDeal(ctx, out, n, pieceSize, exclude, filterByPrice)
+}
+
+// TODO - this is currently not used, if we choose to use it,
```
jcace, line 131: this comment is confusing because the function definitely is being used!

Can the comment just be removed?

handlers.go
en0ma: ok, I will start with just content_deal for this PR only

miner/manager.go
en0ma: sure, noticed it too

jcace (approved):

alvin-reyes (approved):
LGTM

a few things to note: 
- we should abstract the miner selection. 
- think about adding geolocation (miners <> shuttle <> users)
- reputation to include number of deals, location, age, uptime (heartbeat mechanism)

We can do this on another set of PRs.

Great work as usual @en0ma!

PR #535, opened 2022-11-17
Simplify staging zone readiness to size only

neelvirdy:
1. Prevents a situation where we attempt to make deals with too small of content by adding a per-file size minimum
2. Simplifies staging zone readiness to: is it big enough?
3. Staging zones live indefinitely until they become ready
4. Changes accepted staging zone sizes from 13.4GB - 14.4GB to 3.6GB - 14.4GB

We definitely will need to move staging buckets into persisted storage to avoid bucket accumulation. This may be a blocker for this change to go through, but keeping this PR open to track + for discussion on the proposal.

I revisited the tuning of minStagingZoneSizeLimit and maxStagingZoneSizeLimit. It seemed like a fairly tight range that would be hard to hit (13.4GB-14.4GB). IMO the range should correlate to the typical piece size range an SP would accept. From some brief slack searching, it seems like 32GiB should be our absolute max, and some common min piece sizes are 16GB, 1GB, and 256 bytes. I settled on 3.6GB since it will yield consistent deal sizes by matching the individual deal threshold (also 3.6GB), and it strikes a balance between usable in FE but not irrelevantly small for SPs. This way, SPs should see deals from estuary consistently in the range of 3.6GB-14.4GB, likely skewed towards 3.6GB.

Associated FE PR since some metadata that is rendered in FE is removed by this PR: https://github.com/application-research/estuary-www/pull/97

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -18,15 +18,16 @@ const MaxBucketItems = 10000
 // miners who start their deals early don't run into issues
 const DealDuration = 1555200 - (2880 * 21)
 
-// IndividualDealThreshold 90% of the un-padded data size for a 4GB piece
+// IndividualDealThreshold 3.6 GB
+// 90% of the un-padded data size for a 4GB piece
 // the 10% gap is to accommodate car file packing overhead, can probably do this better
-var IndividualDealThreshold = int64(abi.PaddedPieceSize(4<<30).Unpadded() * 9 / 10)
+var IndividualDealThreshold = int64((4 << 30) * 9 / 10)
 
-// MaxStagingZoneSizeLimit 14.29 GB
-var MaxStagingZoneSizeLimit = int64(abi.PaddedPieceSize(16<<30).Unpadded() * 9 / 10)
+// MaxStagingZoneSizeLimit 14.4 GB
+var MaxStagingZoneSizeLimit = int64((16 << 30) * 9 / 10)
 
-// MinStagingZoneSizeLimit 1.81 GB
-var MinStagingZoneSizeLimit = int64(abi.PaddedPieceSize(2<<30).Unpadded() * 9 / 10)
+// MinStagingZoneSizeLimit 3.6 GB
+var MinStagingZoneSizeLimit = int64((4 << 30) * 9 / 10)
 
 const TokenExpiryDurationAdmin = time.Hour * 24 * 365           // 1 year
 const TokenExpiryDurationRegister = time.Hour * 24 * 7          // 1 week
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -7,7 +7,7 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 )
 
-const DefaultContentSizeLimit = 34_000_000_000
+const DefaultContentSizeLimit = int64(34_000_000_000)
 const ContentLocationLocal = "local"
 const TopMinerSel = 15
 const MinSafeDealLifetime = 2880 * 21 // three weeks
@@ -24,10 +24,10 @@ const DealDuration = 1555200 - (2880 * 21)
 var IndividualDealThreshold = int64((4 << 30) * 9 / 10)
 
 // MaxStagingZoneSizeLimit 14.4 GB
-var MaxStagingZoneSizeLimit = int64((16 << 30) * 9 / 10)
+var MaxStagingZoneSizeLimit = DefaultContentSizeLimit
 
 // MinStagingZoneSizeLimit 3.6 GB
-var MinStagingZoneSizeLimit = int64((4 << 30) * 9 / 10)
+var MinStagingZoneSizeLimit = IndividualDealThreshold
 
 const TokenExpiryDurationAdmin = time.Hour * 24 * 365           // 1 year
 const TokenExpiryDurationRegister = time.Hour * 24 * 7          // 1 week
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -22,11 +22,11 @@ const DealDuration = 1555200 - (2880 * 21)
 // the 10% gap is to accommodate car file packing overhead, can probably do this better
 var IndividualDealThreshold = int64(abi.PaddedPieceSize(4<<30).Unpadded() * 9 / 10)
 
-// MaxStagingZoneSizeLimit 14.29 Gib
+// MaxStagingZoneSizeLimit 14.29 GB
 var MaxStagingZoneSizeLimit = int64(abi.PaddedPieceSize(16<<30).Unpadded() * 9 / 10)
 
-// MinStagingZoneSizeLimit 13.29 GiB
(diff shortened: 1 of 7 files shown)


replication.go
```diff
@@ -292,15 +244,15 @@ func (cm *ContentManager) tryAddContent(cb *contentStagingZone, c util.Content)
...
 	}
 
 	if cb.CurSize+c.Size > cb.MaxSize {
-		return false, nil
+		return false, errors.Errorf("cannot add content because resulting size would exceed max bucket size of %d bytes", cb.MaxSize)
 	}
 
 	if len(cb.Contents) >= cb.MaxItems {
-		return false, nil
+		return false, errors.Errorf("cannot add content because resulting number of items would exceed max bucket limit of %d items", cb.MaxItems)
```
en0ma, line 255: no, don't return errors in these places - we don't want to completely fail here, we want a new staging zone to be created if content cannot be added to the current zone

constants/constants.go
```diff
@@ -2,42 +2,29 @@ package constants
...
 
-// 14.29 Gib
-var MaxStagingZoneSizeLimit = int64((abi.PaddedPieceSize(16<<30).Unpadded() * 9) / 10)
+// MaxStagingZoneSizeLimit 14.4 GB
+var MaxStagingZoneSizeLimit = int64((16 << 30) * 9 / 10)
 
-// 13.29 GiB
-var MinStagingZoneSizeLimit = int64(MaxStagingZoneSizeLimit - (1 << 30))
+// MinStagingZoneSizeLimit 3.6 GB
+var MinStagingZoneSizeLimit = int64((4 << 30) * 9 / 10)
```
en0ma, line 27: nit: `var MinStagingZoneSizeLimit = IndividualDealThreshold`

replication.go
```diff
@@ -2089,7 +2026,7 @@ func (cm *ContentManager) makeDealsForContent(ctx context.Context, content util.
 	))
 	defer span.End()
 
-	if content.Size < (256 << 10) {
+	if content.Size < constants.IndividualDealThreshold {
```
en0ma, line 2029: update the error message with the fact that the content size is below the deal sie threshold

en0ma:
Thank you @neelvirdy for doing this, I have always wanted some form of deterministic behavior for the staging zone, and this PR does just that - awesome!

I have always been of the opinion that the min staging bucket size should match the individual deal threshold size - it keeps things consistent (guarantees the same min deal size across aggregated and non-aggregated content deals), intuitive, and deterministic. 

I agree with removing the time-based (closing and keep-alive) rules, as it can easily yield content sizes that would not satisfy the min ask piece size and/or not give SPs consistent min deal size (which I believe Estuary from the onset, has always wanted to give SPs some guarantee that there get almost 3.6Gb per deal).

**Some suggestions**
- Make the staging zone maximum size to be equal to `DefaultContentSizeLimit`. This way, staging buckets will be aggregated above the minimum bucket size, but also gives a wider range to add more contents up to `DefaultContentSizeLimit` to have consistent behavior as non-aggregated contents.

- Change `DefaultContentSizeLimit` to `MaxDealContentSize` and `IndividualDealThreshold` to `MinDealContentSize` to make things more intuitive

handlers.go
```diff
@@ -869,6 +869,17 @@ func (s *Server) handleAdd(c echo.Context, u *util.User) error {
 		return err
 	}
 
+	// if upload size is smaller than (min deal size / staging bucket file count limit)
```
en0ma, line 872: I made a suggestion on how to deal with this [here](https://github.com/application-research/estuary/pull/535#issuecomment-1318197048), I think it is better than rejecting uploads

constants/constants.go
```diff
@@ -21,10 +21,10 @@ const DealDuration = 1555200 - (2880 * 21)
 var IndividualDealThreshold = int64((4 << 30) * 9 / 10)
 
 // MaxStagingZoneSizeLimit 14.4 GB
```
en0ma, line 23: fix description

neelvirdy:
Thanks for the thorough review @en0ma! I incorporated most of your suggestions and have successfully uploaded 20k small files (190KB each) on my local estuary and reached `TransferFinish` deal state. The stats on the deal are 19404 files totaling 3.62 GB. The rest went into a new staging bucket.

I also realized that the staging buckets are modeled and auto-recovered in the contents table [here](https://github.com/application-research/estuary/blob/44d02f7c7fa76a492fc2754c7e6485c0e4562abb/replication.go#L767-L810). I believe we can avoid having a `staging_zones` table entirely and still avoid any reboot issues with the staging zones state by treating every change to the buckets as the following:
- Lock bucketLk
- Write change to contents dB
- Write change to buckets in memory
- Unlock bucketLk

Lmk your thoughts! Ideally I'd like to make the bucket persistence changes in a follow up PR so we can improve SP experience ASAP.

Will look into some cleanup tweaks tomorrow as well, particularly around constant names and redundancy.

image

replication.go
```diff
@@ -690,24 +685,34 @@ func (cm *ContentManager) aggregateContent(ctx context.Context, b *contentStagin
 	}
 }
 
-func (cm *ContentManager) createAggregate(ctx context.Context, conts []util.Content) (*merkledag.ProtoNode, error) {
+func (cm *ContentManager) createAggregate(ctx context.Context, blockstore blockstore.Blockstore, exch exchange.Interface, conts []util.Content) (ipld.Node, error) {
```
en0ma, line 688: no need to pass these params, you can access them from `cm` in this func

handlers.go
```diff
@@ -4594,6 +4583,11 @@ func (s *Server) handleContentHealthCheck(c echo.Context) error {
 		return err
 	}
 
+	var exch exchange.Interface
```
en0ma, line 4586: please revert this, the intention is for the walk below. You can get the exchange in the `createAggregate` func

en0ma:
> Thanks for the thorough review @en0ma! I incorporated most of your suggestions and have successfully uploaded 20k small files (190KB each) on my local estuary and reached `TransferFinish` deal state. The stats on the deal are 19404 files totaling 3.62 GB. The rest went into a new staging bucket.

Great to see that it worked, awesome job!

> I also realized that the staging buckets are modeled and auto-recovered in the contents table [here](https://github.com/application-research/estuary/blob/44d02f7c7fa76a492fc2754c7e6485c0e4562abb/replication.go#L767-L810). I believe we can avoid having a staging_zones table entirely and still avoid any reboot issues with the staging zones state by treating every change to the buckets as the following:
> Write change to contents dB
Lock bucketLk
Write change to buckets in memory
Unlock bucketLk
Lmk your thoughts! Ideally I'd like to make the bucket persistence changes in a follow up PR so we can improve SP experience ASAP.

Let's discuss this in the follow-up PR, but in summary, this PR has address SPs pain, the follow-up PR is to address RAM usage and improve deployment by not rebuilding staging-zones and keeping its metadata in memory is it can quickly build up - we have no control.

en0ma (approved):
LGTM - great job!

PR #537, opened 2022-11-17
allow setting max deal prices config via cli flag

en0ma:
Some node operators would like to set these values as they see fit

**verifications**
- I have verified the defaults are picked up correctly
- I have verified both flags when set are picked up corretcly
- I have verified no changes in code behavior for defaults
- I have verified SPs are filtered correctly when prices are set

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/lotus/chain/types"
 )
 
 const DefaultContentSizeLimit = 34_000_000_000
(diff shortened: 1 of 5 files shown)


jimmylee (approved):
LGTM, lets get another +1 before landing it though.

alvin-reyes:
@en0ma this looks good. 

Please add the following on this PR:

- verification tests - it doesn't need to be a formal doc. Just screenshots and results of non-negative impact is ok.
- OR regression test/edge cases - can we run a few deal making test? What if the miner sets a relatively higher price?
- OR unit tests

Honestly any verification test would be good to include here.

constants/constants.go
```diff
@@ -44,3 +45,14 @@ const TokenExpiryDurationRegister = time.Hour * 24 * 7          // 1 week
 const TokenExpiryDurationLogin = time.Hour * 24 * 30            // 30 days
 const TokenExpiryDurationDefault = time.Hour * 24 * 30          // 30 days
 const TokenExpiryDurationPermanent = time.Hour * 24 * 365 * 100 // 100 years
+
+var DealMaxPrice abi.TokenAmount
+var VerifiedDealMaxPrice = abi.NewTokenAmount(0)
+
+func init() {
```
alvin-reyes, line 52: Why did we need to move this init() to constants?

en0ma: no particular reason besides colocating default values (given all it does is set the default price)

alvin-reyes: Ok I see that you're trying to initialize the value of the dealmax price so this change seems reasonable

alvin-reyes (approved):
LGTM

PR #539, opened 2022-11-17
docs: Correct response type for AddCar API.

simonwo:
Without this, the `estuary-clients` don't understand how to parse the response from the Add CAR API, so that API doesn't work.


alvin-reyes (approved):
LGTM

PR #541, opened 2022-11-17
Omit token field from new auth token db entries

neelvirdy:
Only store token hash for newly created API tokens. Phase 3 of https://github.com/application-research/estuary/issues/343

Anywhere we query the AuthToken DB for `"token = ?", token` has already been replaced with `"token = ? or token_hash = ?", token, GetTokenHash(token)`, so this does not effect the validity of any tokens, existing or new.

---

Demoing creating a token, using it to re-authenticate, revoke another token, and create yet another token.

https://user-images.githubusercontent.com/2850013/202584414-970c618a-14c8-4d6f-a0fd-9b11a4ed1513.mov

This is the state of the DB after this flow:
image

### DIFF
--- a/dbmgr/dbmgr.go
+++ b/dbmgr/dbmgr.go
@@ -328,7 +328,7 @@ func NewAuthTokensQuery(db *gorm.DB) *AuthTokensQuery {
 }
 
 func (q *AuthTokensQuery) Create(authToken AuthToken) error {
-	return q.DB.Create(&authToken).Error
+	return q.DB.Omit("token").Create(&authToken).Error
 }
 
 // CONTENTS
--- a/dbmgr/dbmgr.go
+++ b/dbmgr/dbmgr.go
@@ -3,6 +3,7 @@ package dbmgr
 import (
 	"errors"
 	"fmt"
+	"github.com/application-research/estuary/constants"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


dbmgr/dbmgr.go
```diff
@@ -328,7 +329,7 @@ func NewAuthTokensQuery(db *gorm.DB) *AuthTokensQuery {
 }
 
 func (q *AuthTokensQuery) Create(authToken AuthToken) error {
-	return q.DB.Create(&authToken).Error
+	return q.DB.Omit(constants.TokenColumnName).Create(&authToken).Error
```
en0ma, line 332: instead of using omit, which requires too many changes, why not ignore this field directly in the mode using `gorm: "-"` ?

neelvirdy: We still want to respect existing tokens, and simply warn users in the FE if their token is stored raw and should be replaced for increased security. `gorm: "-"` would prevent reading any existing tokens and make them effectively inaccessible. However, because you asked i discovered `gorm:"->"` which marks a field as read only and achieves what we want without using omit. Updated the PR to use the `->` tag.

en0ma: nice, I didn't know about `->` 😸

en0ma (approved):
LGTM

PR #543, opened 2022-11-18
chore: use breezy and test arm64 build

alvin-reyes:
# Changes

- fixed the dockerfile
- commented out the arm64 build
- use breezy instead of breeze for macos build


en0ma (approved):
LGTM. What was wrong with the old build (how is this an improvement)?

PR #549, opened 2022-11-21
extract content manager into its own package

en0ma:
This PR does not change any of the application behavior, it simply moves things around. This is required so we can start adding unit and integration tests. It will also help set the stage for quick innovation as we slowly learn the boundaries in the project

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -19,7 +19,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/application-research/estuary/constants"
 	"github.com/application-research/estuary/node/modules/peering"
 	"github.com/application-research/estuary/pinner/types"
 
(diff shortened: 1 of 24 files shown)


jlogelin (dismissed):
LGTM, let's continue restructuring/isolating subsystems like this

alvin-reyes (commented):
LGTM.

But we should add the core modules into a core package. This includes, collections, autoretrieve, contentmgr and pinner.
also: 
- the handlers api endpoint should be isolated and put into the API package. 
- api package will have the model, api endpoints.

@en0ma I suggest you document or we document any changes here first before implementing it. I'm also working on a re-arranging the packages.

en0ma:
> LGTM.
> 
> But we should add the core modules into a core package. This includes, collections, autoretrieve, contentmgr and pinner. also:
> 
> * the handlers api endpoint should be isolated and put into the API package.
> * api package will have the model, api endpoints.
> 
> @en0ma I suggest you document or we document any changes here first before implementing it. I'm also working on a re-arranging the packages.

The aim of the PR is not to get things in their perfect state at the moment. The aim is to only extract boundaries we are sure of and do it without regression. We can later adjust as we see fit,, and it will be easy to PR as it will most likely be a proper refactor. 

Also, the API extraction is coming next. I needed to this this first, so as to safely extract the api

alvin-reyes:
> > LGTM.
> > But we should add the core modules into a core package. This includes, collections, autoretrieve, contentmgr and pinner. also:
> > 
> > * the handlers api endpoint should be isolated and put into the API package.
> > * api package will have the model, api endpoints.
> > 
> > @en0ma I suggest you document or we document any changes here first before implementing it. I'm also working on a re-arranging the packages.
> 
> The aim of the PR is not to get things in their perfect state at the moment. The aim is to only extract boundaries we are sure of and do it without regression. We can later adjust as we see fit,, and it will be easy to PR as it will most likely be a proper refactor.
> 
> Also, the API extraction is coming next. I needed to this this first, so as to safely extract the api

I don't think we should be putting all core modules on the root folder package.

`/api`
`/internal` (for all core functions) - this should have the core modules.
`/pkg`
`/config`
`/cmd`

alvin-reyes (approved):
LGTM

PR #554, opened 2022-11-21
remove dead package (dbmgr)

en0ma:


alvin-reyes (approved):
I do think we need a database package for all our DB objects / model / dao. We can add that later after this PR. 

LGTM.

Issue #236, opened 2022-05-25
Staged content still remains in staging zone after it is deleted/unpinned


en0ma:
When staged content is deleted (unpinned), it still remains in the staging zone and shows up in the UI

corinne-antonia:
@en0ma Do you have everything you need to work on this?

neelvirdy:
@en0ma i just started looking into this. is the motivation behind this that dealmaking would fail if the staging zone is attempting to aggregate a content that is unpinned? so if a user added contents, it went into a staging zone, then called DELETE /pinning/pins/:pinid on that pin, they would essentially corrupt their staging zone unless we also removed the associated content from the staging zone?

en0ma:
> @en0ma i just started looking into this. is the motivation behind this that dealmaking would fail if the staging zone is attempting to aggregate a content that is unpinned? so if a user added contents, it went into a staging zone, then called DELETE /pinning/pins/:pinid on that pin, they would essentially corrupt their staging zone unless we also removed the associated content from the staging zone?

yes

PR #555, opened 2022-11-21
When deleting pins, also remove from staging buckets

neelvirdy:
Fixes https://github.com/application-research/estuary/issues/236

This is the resulting staging bucket after adding 6 files and deleting 2 of them via:
`curl -X DELETE "http://localhost:3004/pinning/pins/2" -H "Authorization: Bearer <TOKEN>"`
`curl -X DELETE "http://localhost:3004/pinning/pins/6" -H "Authorization: Bearer <TOKEN>"`

Content ID 3 is taken by the bucket itself.

I also tested restarting my server and verified the staging bucket state remained consistent, so the DB and in-memory buckets were both handled accurately throughout this process. 

image

### DIFF
--- a/contentmgr/gc.go
+++ b/contentmgr/gc.go
@@ -171,10 +171,12 @@ func (cm *ContentManager) UnpinContent(ctx context.Context, contid uint) error {
 		}
 	}
 
-	buckets, _ := cm.Buckets[pin.UserID]
-	for _, bucket := range buckets {
-		if _, err := cm.tryRemoveContent(bucket, pin); err != nil {
-			return err
+	buckets, ok := cm.Buckets[pin.UserID]
+	if ok {
+		for _, bucket := range buckets {
+			if _, err := cm.tryRemoveContent(bucket, pin); err != nil {
+				return err
+			}
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


contentmgr/replication.go
```diff
@@ -223,6 +223,35 @@ func (cm *ContentManager) tryAddContent(cb *contentStagingZone, c util.Content)
...
+func (cm *ContentManager) tryRemoveContent(cb *contentStagingZone, c util.Content) (bool, error) {
+	cb.lk.Lock()
+	defer cb.lk.Unlock()
+
+	// if this bucket is being consolidated, do not add anymore content
+	if cb.IsConsolidating {
+		return false, nil
+	}
+
+	if err := cm.DB.Model(util.Content{}).
```
en0ma, line 235: no need to modify the db state

en0ma (dismissed):

contentmgr/gc.go
```diff
@@ -170,6 +170,13 @@ func (cm *ContentManager) UnpinContent(ctx context.Context, contid uint) error {
 			return err
 		}
 	}
+
+	buckets, _ := cm.Buckets[pin.UserID]
```
alvin-reyes, line 174: Just a suggestion: 

```
buckets, ok := cm.Buckets
if ok {
  for _, bucket := range buckets {
    if _, err := cm.tryRemoveContent(bucket, pin); err != nil {
      return err
    }
  }
}
```

en0ma (approved):

PR #560, opened 2022-11-23
fix: use the default links per block for importing file

alvin-reyes:
# Changes

I noticed that the dag builder generates a different cid when from IPFS. 

This PR will use the same default number of links per block configuration on the dag builder which will make it par with IPFS/Kubo. 

I have tested it with several files and it works as expected. @en0ma please test this against your files as well.

### DIFF
--- a/util/unixfs.go
+++ b/util/unixfs.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 
-	"github.com/ipfs/go-cidutil"
 	chunker "github.com/ipfs/go-ipfs-chunker"
 	ipld "github.com/ipfs/go-ipld-format"
 	"github.com/ipfs/go-merkledag"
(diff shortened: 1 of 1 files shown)


jlogelin:
LGTM

jlogelin (approved):
LGTM

en0ma:
@alvin-reyes It did not work for me

gmelodie:
@alvin-reyes I got both equal and different CIDs. The difference seems to be when the file is larger in size.

alvin-reyes:
I realized that we don’t need to change the way we compute our CIDs since it’ll work either anyway. We encountered a weird behaviour last week where in the CIDs are different from the ones uploaded on api.estuary.tech and on shuttles.

I think @en0ma already fixed this on his commit (e.i revert the logic to dump the content to the blockstore).

PR #562, opened 2022-11-23
make redirect upload work for add and add-car

en0ma:
The current redirect upload does not work for car uploads. This PR makes sure it proxy properly to the right endpoint and supports both `content/add` and `content/add-car`

**verification**

I used the `--disable-local-content-adding` to force the primary to redirect to my local shuttle. Below are the response directly from the shuttle. 

Screenshot 2022-11-23 at 13 33 35
Screenshot 2022-11-23 at 13 35 11

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"net/http"
+	"net/http/httputil"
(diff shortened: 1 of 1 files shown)


alvin-reyes (approved):
LGTM

PR #565, opened 2022-11-23
create sharded dir in shuttle for shuttle contents

en0ma:
This PR moves shuttle content aggregation to shuttle (its faster, cleaner, and more deterministic).

I tested the previous implementation where aggregation happens on primary, for shuttle contents, sharded dir requires bitswap so it takes time.

**verification**

I have tested this, it works as expected - super fast!

### DIFF
--- a/cmd/estuary-shuttle/rpc.go
+++ b/cmd/estuary-shuttle/rpc.go
@@ -47,7 +47,7 @@ func (d *Shuttle) handleRpcCmd(cmd *drpc.Command) error {
 	case drpc.CMD_TakeContent:
 		return d.handleRpcTakeContent(ctx, cmd.Params.TakeContent)
 	case drpc.CMD_AggregateContent:
-		return d.handleRpcAggregateStagedContent(ctx, cmd.Params.AggregateContents)
+		return d.handleRpcAggregateStagedContent(ctx, cmd.Params.AggregateContent)
 	case drpc.CMD_StartTransfer:
 		return d.handleRpcStartTransfer(ctx, cmd.Params.StartTransfer)
 	case drpc.CMD_PrepareForDataRequest:
--- a/cmd/estuary-shuttle/rpc.go
+++ b/cmd/estuary-shuttle/rpc.go
@@ -302,21 +302,21 @@ func (d *Shuttle) handleRpcTakeContent(ctx context.Context, cmd *drpc.TakeConten
 	return nil
 }
 
-func (s *Shuttle) handleRpcAggregateStagedContent(ctx context.Context, cmd *drpc.AggregateContents) error {
+func (s *Shuttle) handleRpcAggregateStagedContent(ctx context.Context, aggregate *drpc.AggregateContents) error {
 	// only progress if aggr is not allready in progress
-	if !s.markStartAggr(cmd.DBID) {
+	if !s.markStartAggr(aggregate.DBID) {
 		return nil
 	}
-	defer s.finishAggr(cmd.DBID)
+	defer s.finishAggr(aggregate.DBID)
 
 	ctx, span := s.Tracer.Start(ctx, "handleAggregateContent", trace.WithAttributes(
-		attribute.Int64("dbID", int64(cmd.DBID)),
-		attribute.Int64("userId", int64(cmd.UserID)),
+		attribute.Int64("dbID", int64(aggregate.DBID)),
+		attribute.Int64("userId", int64(aggregate.UserID)),
 	))
 	defer span.End()
 
 	var p Pin
-	err := s.DB.First(&p, "content = ?", cmd.DBID).Error
+	err := s.DB.First(&p, "content = ?", aggregate.DBID).Error
 	switch err {
 	default:
 		return err
@@ -330,7 +330,7 @@ func (s *Shuttle) handleRpcAggregateStagedContent(ctx context.Context, cmd *drpc
 		// normal case
 	}
 
-	for _, c := range cmd.Contents {
+	for _, c := range aggregate.Contents {
 		var cont Pin
 		if err := s.DB.First(&cont, "content = ?", c.ID).Error; err != nil {
 			// TODO: implies we dont have all the content locally we are being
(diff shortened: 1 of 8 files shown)


cmd/estuary-shuttle/rpc.go
```diff
@@ -319,61 +321,81 @@ func (s *Shuttle) handleRpcAggregateStagedContent(ctx context.Context, cmd *drpc
 	default:
 		return err
 	case nil:
-		// exists already
+		// if it was aggregated successfully but estuary was not notified,, resend message
+		if p.Active {
+			return s.resendPinComplete(ctx, p)
+		}
 		return nil
```
neelvirdy, line 328: nit: should we log a warning if we're hitting this `return nil` case? sounds like we'd only hit it if we're sending redundant aggregate cmds, but would be good to make sure we're not doing it

cmd/estuary-shuttle/rpc.go
```diff
@@ -319,61 +321,81 @@ func (s *Shuttle) handleRpcAggregateStagedContent(ctx context.Context, cmd *drpc
...
 	case gorm.ErrRecordNotFound:
 		// normal case
 	}
 
-	totalSize := int64(len(cmd.ObjData))
 	for _, c := range cmd.Contents {
```
neelvirdy, line 333: nit: can we consider some different var names here than `c` and `cmd.Contents` since they're the `AggregateContent` type? would help readability imo. maybe `for _, aggrContent := range cmd.AggrContents`. or we could rename `AggregateContent` to `ContentMetadata`, but plurality gets confusing with that one, so defer to you

cmd/estuary-shuttle/rpc.go
```diff
@@ -319,61 +321,81 @@ func (s *Shuttle) handleRpcAggregateStagedContent(ctx context.Context, cmd *drpc
...
 
-	pin := &Pin{
-		Content:   cmd.DBID,
-		Cid:       util.DbCID{CID: cmd.Root},
-		UserID:    cmd.UserID,
-		Size:      totalSize,
-		Active:    false,
-		Pinning:   true,
-		Aggregate: true,
+	bserv := blockservice.New(s.Node.Blockstore, s.Node.Bitswap)
```
neelvirdy, line 346: seems like we can make a helper to do this to de-dupe with `CreateAggregate` in replication.go

contentmgr/replication.go
```diff
@@ -196,9 +189,10 @@ func (cb *contentStagingZone) updateReadiness() {
 		cb.MinSize)
 }
 
-func (cm *ContentManager) tryAddContent(cb *contentStagingZone, c util.Content) (bool, error) {
+func (cm *ContentManager) tryAddContent(cb *ContentStagingZone, c util.Content) (bool, error) {
 	cb.lk.Lock()
 	defer cb.lk.Unlock()
+	defer cb.updateReadiness()
```
neelvirdy, line 195: nice!

cmd/estuary-shuttle/rpc.go
en0ma: yeah, we can do that on a later PR

cmd/estuary-shuttle/rpc.go
en0ma: the naming is ok, the issue I see here is `cmd`, it should have been `aggregate`, so it reads `aggregate.Contents`

cmd/estuary-shuttle/rpc.go
en0ma: yeah, I warning will be good here

cmd/estuary-shuttle/rpc.go
en0ma: refactoring this part between estuary and shuttle is a bit ticky at the moment

contentmgr/pinning.go
```diff
@@ -400,10 +400,31 @@ func (cm *ContentManager) handlePinningComplete(ctx context.Context, handle stri
 
 	if cont.Aggregate {
 		// this is used by staging content aggregate
+		if len(pincomp.Objects) != 1 {
```
alvin-reyes, line 403: Aggregate are supposed to have more than 1 objects right? 

This will also pass if `len` = 0. Use `> 1` instead.

contentmgr/pinning.go
```diff
@@ -400,10 +400,31 @@ func (cm *ContentManager) handlePinningComplete(ctx context.Context, handle stri
...
+			return fmt.Errorf("aggregate has more than 1 objects")
+		}
+
+		obj := &util.Object{
+			Cid:  util.DbCID{CID: pincomp.Objects[0].Cid},
+			Size: pincomp.Objects[0].Size,
+		}
+		if err := cm.DB.Create(obj).Error; err != nil {
+			return err
```
alvin-reyes, line 412: return xerrors.Errorf("failed to create Object: %w", err)

contentmgr/pinning.go
```diff
@@ -400,10 +400,31 @@ func (cm *ContentManager) handlePinningComplete(ctx context.Context, handle stri
...
+		}
+
+		if err := cm.DB.Create(&util.ObjRef{
+			Content: cont.ID,
+			Object:  obj.ID,
+		}).Error; err != nil {
+			return err
```
alvin-reyes, line 419: return xerrors.Errorf("failed to create Object reference: %w", err)

contentmgr/pinning.go
en0ma: aggregates should onnly ever 1 one object

contentmgr/replication.go
```diff
@@ -111,32 +111,25 @@ type stagingZoneReadiness struct {
...
+type ContentStagingZone struct {
+	ZoneOpened      time.Time            `json:"zoneOpened"`
+	Contents        []util.Content       `json:"contents"`
+	MinSize         int64                `json:"minSize"`
+	MaxSize         int64                `json:"maxSize"`
+	CurSize         int64                `json:"curSize"`
+	User            uint                 `json:"user"`
+	ContID          uint                 `json:"contentID"`
+	Location        string               `json:"location"`
+	IsConsolidating bool                 `json:"is_consolidating"`
```
alvin-reyes, line 123: Is this something we return as part of an HTTP response? if not, lets use `isConsolidating` instead. If yes, leave it as is.

en0ma: It is not, I can fix it

alvin-reyes (approved):
LGTM

PR #568, opened 2022-11-24
catch panic from http request

en0ma:
This makes sure HTTP request does not crash the servers. I have verified this by adding a panic on an endpoint, calling the endpoint, and got a 500 response as expected


alvin-reyes (approved):
LGTM

PR #569, opened 2022-11-24
Boost throttler

anjor:


config/estuary.go
```diff
@@ -181,6 +181,7 @@ func NewEstuary(appVersion string) *Estuary {
 				LowWater:  2000,
 				HighWater: 3000,
 			},
+			Libp2pThrottleLimit: 10,
```
magik6k, line 184: Intuitively that seems like a low default; You probably want something like 50-100, just not 10000, tho this is definitely one of those knows you need to turn a bunch to see at what setting you get most bandwidth / lowest fail rate

en0ma (approved):
lgtm

PR #588, opened 2022-11-24
/buckets (collections v2)

gmelodie:
This is how the `/collections` API will look. We'll rebrand it as `buckets` since this is much more of an industry-standard term.

#570


en0ma (approved):
LGTM as a first step for name change from collections to buckets

PR #589, opened 2022-11-24
[draft] feat: stop creating deals for unreachable content

jcace:
Check if the content is accessible before creating a deal with SP. 

TODO: Persist this in some metadata so we don't have to constantly do the lookup on the content everytime it tries to make deals

### DIFF
--- a/contentmgr/replication.go
+++ b/contentmgr/replication.go
@@ -1089,6 +1089,13 @@ func (cm *ContentManager) ensureStorage(ctx context.Context, content util.Conten
 		}
 	}
 
+	bserv := blockservice.New(cm.Blockstore, cm.Node.Bitswap)
+	dserv := merkledag.NewDAGService(bserv)
+	_, err = dserv.Get(ctx, content.Cid.CID)
+	if err != nil {
+		return errors.New("will not make deal as the content is unretrievable")
+	}
+
 	go func() {
 		// make some more deals!
 		cm.log.Debugw("making more deals for content", "content", content.ID, "curDealCount", len(deals), "newDeals", dealsToBeMade)


whyrusleeping:
hrm, this doesnt guarantee the whole dag is available, and if this error ever happens that means we've done something very wrong. Are we seeing this on our nodes right now?

jcace:
The intention is to stop these errors:

Screenshot 2022-11-24 at 2 31 03 PM

As I understand it, this error is caused because we have a `corrupted LMDB blockstore` on some shuttles, but Estuary continues to try and make deals with that content. It results in a lot of `error reading from http response stream: unexpected EOF` errors on the Boost side. So SP's are getting spammed with these deals where the content can't be downloaded.

en0ma:
@jcace @whyrusleeping I am already working on this. @jcace based on our conversation along with @jlogelin, we definitely have corrupt data that will never make a deal, no need to send them over to SPs.

The PR I am landing in today does a sanity check (re-commp) before deals are attempted. It will be appropriately abstracted from the current deal given that this quick hack reduces the SP deal failure pains. 

For queued contents to be sealed, we call a sanity check on it (using a new table to track these checks)

en0ma:
> @jcace @whyrusleeping I am already working on this. @jcace based on our conversation along with @jlogelin, we definitely have corrupt data that will never make a deal, no need to send them over to SPs.
> 
> The PR I am landing in today does a sanity check (re-commp) before deals are attempted. It will be appropriately abstracted from the current deal given that this quick hack reduces the SP deal failure pains.
> 
> For queued contents to be dealt with, we call a sanity check on it (using a new table to track these checks). It will handle both primary and shuttle contents.
> 
> Your current approach will not suffice.

I just came up with an optimistic approach, using a sanity blockstore wrapper. It will detect missing blocks, mark the contents and such contents will not deal again

jcace:
closing in favour of @en0ma pr: https://github.com/application-research/estuary/pull/593

PR #590, opened 2022-11-25
fix collection get content response

anjor:
Right now the swagger generated clients return a string for this endpoint, which means clients need to go through an extra parsing function.


en0ma (approved):
LGTM

snissn:
thank you! great change! Possibly a lot more of these we need to do

PR #593, opened 2022-11-25
sanity check for bad data

en0ma:
An optimistic approach to marking bad content, so that there are not sent to SPs (as they will fail during data transfer).

The first time `commp` or boost fails to pull blocks due to missing blocks, this will mark all contents affected by that block, so we stop sending deals for them and manually address such contents. This will reduce spamming SPs with deals that will fail due to bad content.

**verification**
I have tested it by trying to generate `commp`, trying the health check endpoint, and trying to access the content via gateway, in all 3 cases, the check worked!

- Estuary requesting commp from a shuttle for content with missing block, the shuttle sent back a sanity check (look at the time difference between the logs - its performant, cc @alvin-reyes ) to Estuary and Estuary marks the content.
Screenshot 2022-11-26 at 01 20 26

- Next, calling gateway and content health endpoint on the shuttle for same data, shuttle sends sanity check to Estuary to mark it
Screenshot 2022-11-26 at 01 20 43

- In all 3 cases, the sanity table only records 1 entry as expected
Screenshot 2022-11-26 at 01 45 17

- Lastly, trying to make deal, the content is ignored since it has missing blocks.
Screenshot 2022-11-26 at 01 30 53

-


jcace:
nice! I like this abstraction

jlogelin (dismissed):
LGTM

jcace (approved):

contentmgr/sanity_check.go
```diff
@@ -0,0 +1,28 @@
...
+	if err := cm.DB.Find(&cnts, where, cc.Bytes()).Error; err != nil {
+		cm.log.Errorf("sanity check failed to get content(s) for cid: %s, err: %w", cc.String(), err)
+		return
+	}
+
+	// mark all contents affected by this missing block
+	var marks map[uint]bool
+	for _, cnt := range cnts {
+		if _, ok := marks[cnt.ID]; !ok {
+			if err := cm.DB.Model(&util.Content{}).Where("id = ?", cnt.ID).Update("failed_sanity_check", true).Error; err != nil {
```
alvin-reyes, line 21: Are we introducing a new field? I suggest you create a roll forward and back SQL script with transactions enabled.

contentmgr/shuttle.go
```diff
@@ -220,6 +220,13 @@ func (cm *ContentManager) processShuttleMessage(handle string, msg *drpc.Message
 			cm.log.Errorf("handling split complete message from shuttle %s: %s", handle, err)
 		}
 		return nil
+	case drpc.OP_SanityCheck:
+		sc := msg.Params.SanityCheck
+		if sc == nil {
+			return ErrNilParams
+		}
+		cm.HandleSanityCheck(sc.CID, sc.Err)
```
alvin-reyes, line 228: I think this is going to be an expensive operation (it looks up obj and obj ref tables). I’m thinking if we should decouple this as part of the GC.

util/content.go
```diff
@@ -99,6 +99,8 @@ type Content struct {
 	// them (unlike with aggregates)
 	DagSplit  bool `json:"dagSplit"`
 	SplitFrom uint `json:"splitFrom"`
+
+	FailedSanityCheck bool `json:"-"`
```
alvin-reyes, line 103: Introducing a new column in a massive table like Content needs to have a roll forward and back script. We will need to execute this SQL script on API node first before we deploy the app.

contentmgr/shuttle.go
en0ma: I do not think so, we do this all the time when we do commp, plus missing blocks do not happen often (and should not)

contentmgr/sanity_check.go
alvin-reyes: I'll add this to the instructions for the next release.

file_name: contents_add_failed_sanity_check_column_roll_forward.sql
```
\echo 'start - adding failed_sanity_check on contents table'
// Roll forward
alter table contents add failed_sanity_check boolean;
\echo 'end - adding failed_sanity_check on contents table'
```

file_name: contents_add_failed_sanity_check_column_roll_back.sql
```
\echo 'start - adding failed_sanity_check on contents table'
alter table contents drop column failed_sanity_check;
\echo 'end - adding failed_sanity_check on contents table'
```

en0ma: @alvin-reyes I think it best to use another table, so we can persist the missing block and the error message - which might come in handy for manually fixing the data, plus this won't strain the contents table

util/content.go
en0ma: I have changed it to use a new table

en0ma:
@jcace @jlogelin @alvin-reyes I have tested this with one of the bad data - it works!

jcace (approved):
Looks good!

alvin-reyes (approved):
LGTM

alvin-reyes:
Great work on this. Let's get this included on v0.2.0.

PR #596, opened 2022-11-27
Chore/extract api package

en0ma:
This PR moves all API logic into an `apiV1` package and sets the ground for API v2 work which is `apiv2` both in the `api` directory. It is also a step toward improving the code for readability, testing, and identifying service boundaries (which will help us build a better deterministic platform).

**verification**
- I have tested `content/add`, `content/add-car`, `pinnging/pin`, `staging-zones`, and `deal-making`, `shuttles` - all works

Please have a look and provide feedback


gmelodie (commented):
It makes very little sense to me to add the API changes alongside other refactors on the same PR. Can you decouple these into two or more PRs? If so I'll approve the API changes

en0ma:
> It makes very little sense to me to add the API changes alongside other refactors on the same PR. Can you decouple these into two or more PRs? If so I'll approve the API changes

I totally understand @gmelodie, But some of those changes are required to bring the API abstraction to life. I know there are some changes that could have been avoided mostly around the content manager (moving files around - I needed to do that as I tried to solve the pinner cyclic dependence error)

gmelodie (approved):

alvin-reyes:
Let’s not merge this until we get v0.2.1 out.

gmelodie:
@en0ma since this is blocking APIv2 development, let's merge it to `api-v2`

contentmgr/peice_commp.go
```diff
@@ -0,0 +1,134 @@
+package contentmgr
```
neelvirdy, line 1: nit: filename spelling > `piece_commp.go`

api/v1/handlers.go
```diff
@@ -3902,7 +3654,7 @@ func (s *Server) handleDeleteContentFromCollection(c echo.Context, u *util.User)
 	return c.NoContent(http.StatusOK)
 }
 
-func (s *Server) tracingMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
+func (s *apiV1) tracingMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
```
neelvirdy, line 3678: you can remove this since you have it in util/tracing_middleware.go and configure it in api_engine already

elijaharita:
microscopic nit: could we bring the package names more inline with the go style guide? https://go.dev/blog/package-names

my recommendations: 
`api_engine` -> `apiengine` or `apieng`
`apiV1` -> `apiv1` or just `api` (if  two versions are imported, they can be renamed locally at import to `apiv1`, `apiv2`, etc.)

pinner/progress/progress.go
```diff
@@ -0,0 +1,3 @@
+package pinning_progress
```
neelvirdy, line 1: is this worth having as its own package? or own file? seems like declaring a type in pinmgr.go would be ok, or we could have a types.go file in the pinner package to declare this

pinner/pinning_op/pinning_op.go
```diff
@@ -0,0 +1,77 @@
+package pinning_op
```
neelvirdy, line 1: also wondering why this has to be its own package, seems more convenient to have it be part of the pinner package

pinner/pinning_op/pinning_op.go
```diff
@@ -36,24 +36,42 @@ type PinningOperation struct {
...
 	po.LastUpdate = time.Now()
 }
+
+func (po *PinningOperation) Complete() {
+	po.lk.Lock()
+	defer po.lk.Unlock()
+
+	po.EndTime = time.Now()
+	po.LastUpdate = time.Now()
+	po.Status = types.PinningStatusPinned
```
neelvirdy, line 68: nit: can we update the status before these timestamps and use the same timestamp for EndTime and LastUpdate?

pinner/pinning_op/pinning_op.go
```diff
@@ -36,24 +36,42 @@ type PinningOperation struct {
...
 
 func (po *PinningOperation) Fail(err error) {
-	po.Lk.Lock()
+	po.lk.Lock()
+	defer po.lk.Unlock()
+
 	po.FetchErr = err
 	po.EndTime = time.Now()
 	po.Status = types.PinningStatusFailed
 	po.LastUpdate = time.Now()
```
neelvirdy, line 51: nit: can we update the status before these timestamps and use the same timestamp for EndTime and LastUpdate?

neelvirdy:
another nit: it would be nice if we could have subfolders, packages, and API groups match when feasible.

example: `pinner/` -> `pinning/` for folder name, contains a `pinning` package, implements the `/pinning` API group.

miners may be another feasible case, but totally fair to include that with the storage providers switch later in API v2

pinner/progress/progress.go
neelvirdy: nvm - i see the cycle with cm now

en0ma:
> another nit: it would be nice if we could have subfolders, packages, and API groups match when feasible.
> 
> example: `pinner/` -> `pinning/` for folder name, contains a `pinning` package, implements the `/pinning` API group.
> 
> miners may be another feasible case, but totally fair to include that with the storage providers switch later in API v2

Yeah, this will come in a different PR, I did not want to make too much unrelated changes. Ideally, the pinner package should hold all pinning logic. This PR is a first step to better abstractions.

en0ma:
> microscopic nit: could we bring the package names more inline with the go style guide? https://go.dev/blog/package-names
> 
> namely the no underscores rule and no capitals rule
> 
> my recommendations: `api_engine` -> `apiengine` or `apieng` `apiV1` -> `apiv1` or just `api` (if two versions are imported, they can be renamed locally at import to `apiv1`, `apiv2`, etc.)

Nice, done.

pinner/pinning_op/pinning_op.go
neelvirdy, line 1: nit: `pinning_op` -> `operation` for idiomatic package name and so it can be imported as `/pinner/operation` and renamed locally as oneof: `op` or `po` or `pinop`

also should update folder and filename to operation

pinner/progress/progress.go
neelvirdy, line 1: nit: `pinning_progress` -> `progress` for idiomatic package name and so it can be imported as `/pinner/progress`

neelvirdy:
LGTM. also manually verified i'm still able to upload, consolidate, and aggregate a deal with an API node + one shuttle setup

neelvirdy (approved):

PR #597, opened 2022-11-28
stop multiple pin status update

en0ma:
While extracting codes and testing, I noticed pin status update is happening twice (it should happen only once).


alvin-reyes (approved):
LGTM

PR #598, opened 2022-11-28
change pin retry duration to 6 months

en0ma:
@alving re: https://github.com/application-research/estuary/pull/526#issuecomment-1327514531, I think 6 months is reasonable


alvin-reyes (approved):
LGTM

Issue #583, opened 2022-11-24
Restify /contents endpoint


gmelodie:
- /content/add -> /contents/ POST
    - new field "type", which can be
        - "file": raw data on the request; (needs input and alternatives);
        - "cid" (replaces /content/ipfs-add): JSON {"cid": "my-cid-here"};
        - or "car" (replaces /content/add-car): JSON (same json we currently use for /content/add-car;

- delete /content/create endpoint (redundant with /content/add)
- remove /contents/importdeal (see /deals endpoint issue)
- /content/read/{cont} -> /contents/{contentid}/read (use contentid instead of cid}
- /content/sealed -> /contents?sealed=true
- remove /contents/{contentid}/read -> use /contents/{contentid} instead
- merge /contents/{contentid}/failures + /contents/{contentid}/bandwidth ->
  /contents/{contentid}/status
- remove /contents/staging-zones (see /staging-buckets issue) 
- remove /contents/deals (see /deals issue)
- /public/cid/{cid} -> /contents?cid=mycid
- remove /public/stats -> use /contents/stats for this purpose

PR #600, opened 2022-11-28
Restify /contents API

gmelodie:
Addresses #583

## Verification
1. Add CAR, add to bucket (new name for collections) and dir too
image
1.2. List that bucket on that dir to see the file
image

2. Add CID, add to bucket and dir too
image
2.2. List that bucket on that dir to see the pinned file
image

3. Getting the status of one of the contents
image

4. Get content by CID
image

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -160,16 +160,11 @@ func (s *Server) ServeAPI() error {
 	e.POST("/contents", withUser(s.handleAdd), s.AuthRequired(util.PermLevelUpload))
 
 	contents := e.Group("", s.AuthRequired(util.PermLevelUser))
-
 	contents.GET("/contents", withUser(s.handleListContent))
 	contents.GET("/contents/by-cid/:cid", s.handleGetContentByCid)
 	contents.GET("/contents/:contentid", withUser(s.handleGetContent))
-	contents.GET("/contents/stats", withUser(s.handleStats))
 	contents.GET("/contents/ensure-replication/:cid", s.handleEnsureReplication)
 	contents.GET("/contents/:contentid/status", withUser(s.handleContentStatus))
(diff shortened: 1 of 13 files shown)


kelindi:
@gmelodie Could we have the car files be passed through formData rather than through the body to be more consistent with how regular files are added

handlers.go
```diff
@@ -157,32 +155,24 @@ func (s *Server) ServeAPI() error {
...
-	content.GET("/bw-usage/:content", withUser(s.handleGetContentBandwidth))
-	content.GET("/staging-zones", withUser(s.handleGetStagingZoneForUser))
-	content.GET("/aggregated/:content", withUser(s.handleGetAggregatedForContent))
-	content.GET("/all-deals", withUser(s.handleGetAllDealsForUser))
+	// to upload contents you only need an upload key
+	// to see info about contents you need a user-level key (see contents group)
+	e.POST("/contents", withUser(s.handleAdd), s.AuthRequired(util.PermLevelUpload))
+
+	contents := e.Group("", s.AuthRequired(util.PermLevelUser))
+	contents.GET("/contents", withUser(s.handleListContent))
```
anjor, line 163: Shouldn't these be behind /api/v2 ?

anjor (changes requested):
We should not delete the v1 endpoints. They should still be around. The new endpoints should be under a v2 path.

Separately, I think uploadType should be an enum

gmelodie:
@kelindi I'm trying not to change too much. I think it's about balance.

kelindi:
@gmelodie I think it would add some more consistency when uploading a file, which would be better for dx. If I am understanding it correctly, the current route you have is the same no matter the file and then it figures out what the case is based on the file? If that is the case it would make more sense to have the request be formatted the same when uploading a regular file or a car file.

gmelodie:
Closing in favor of #716

PR #605, opened 2022-11-28
604 increment subversion in estuary clients when we run our GitHub action

snissn:
github actions run on master branch


snissn:
too many files changed i must have pushed wrong

PR #608, opened 2022-11-28
increment subversion in estuary-clients when we run our github action

snissn:
github actions run against master, so this needs to update master


jlogelin (approved):
LGTM

snissn:
ty @anjor @jlogelin

PR #609, opened 2022-11-28
Track buckets in DB instead of in-memory

neelvirdy:


alvin-reyes (commented):
I was thinking if we can add a new buckets table which will have an bucket uuid andevery CID is then assigned to one of these bucket which can either be a dedicated bucket (per user) or a global bucket.

The bucket is then processed like a batch job with different trigger parameters:
1 - size of the bucket, meaning if it reaches a threshold (3.5GB), it'll get processed.
2 - time (we can schedule it everyday) - cron job.

neelvirdy:
Currently writing up a task to recompute staging zone sizes at startup and update the DB if they mismatch what is in the aggregate content's row. This is a migration task and is required to move to tracking size incrementally in DB, since there staging zones will already exist when this gets deployed but their sizes will not have been accounted for in the incremental tracking.

contentmgr/replication.go
```diff
@@ -297,17 +251,33 @@ func (cm *ContentManager) ToCheck(contID uint) {
 	}
 }
 
+func (cm *ContentManager) getReadyStagingZones() ([]util.Content, error) {
+	var readyZones []util.Content
+	if err := cm.DB.Model(&util.Content{}).
+		Where("not active and pinning and aggregate and size >= ?", constants.MinDealContentSize).
```
alvin-reyes, line 257: do we need to add a filter to only get content that are relatively new? created within 3 months?

contentmgr/gc.go
```diff
@@ -152,6 +151,17 @@ func (cm *ContentManager) UnpinContent(ctx context.Context, contid uint) error {
 		return err
 	}
 
+	if pin.AggregatedIn > 0 {
+		if cm.ZonesConsolidating[pin.AggregatedIn] {
+			return fmt.Errorf("unable to unpin content while zone is consolidating (pin: %d, zone: %d)", pin.ID, pin.AggregatedIn)
+		}
+		if err := cm.DB.Model(util.Content{}).
+			Where("id = ?", pin.AggregatedIn).
```
alvin-reyes, line 159: will there be a possible scenario that a "aggregated" content is older than 3 months? @en0ma

contentmgr/replication.go
en0ma: no, this should be fine, it has nothing to do with pinning, it's basically just taking contents ready for deal making

contentmgr/replication.go
```diff
@@ -297,17 +251,33 @@ func (cm *ContentManager) ToCheck(contID uint) {
...
 func (cm *ContentManager) runStagingBucketWorker(ctx context.Context) {
 	timer := time.NewTicker(cm.cfg.StagingBucket.AggregateInterval)
 	for {
 		select {
 		case <-timer.C:
 			cm.log.Debugw("content check queue", "length", len(cm.queueMgr.queue.elems), "nextEvent", cm.queueMgr.nextEvent)
 
-			zones := cm.popReadyStagingZone()
-			for _, z := range zones {
+			cm.log.Infof("checking for ready staging zones")
```
en0ma, line 271: make this `debugf`

contentmgr/replication.go
```diff
@@ -297,17 +251,33 @@ func (cm *ContentManager) ToCheck(contID uint) {
...
+			readyZones, err := cm.getReadyStagingZones()
+			if err != nil {
+				cm.log.Errorf("failed to get ready staging zones: %s", err)
+			}
+
+			cm.log.Infof("found ready staging zones: %d", len(readyZones))
```
en0ma, line 277: make this `debugf`

contentmgr/gc.go
en0ma: yeah, but there is no issue, as their contents are pinned, just that no deal is made yet

en0ma: @neelvirdy thinking about it, when a pin is unpinned, we should only remove it in these conditions;

- if it's not in an aggregate - remove
- if it's in an aggregate, but the aggregate is not pinned(not active) yet - remove it
- if it's in an aggregate and the aggregate is already pinned(active) - ignore it

contentmgr/pinning.go
```diff
@@ -428,6 +426,8 @@ func (cm *ContentManager) handlePinningComplete(ctx context.Context, handle stri
 		}).Error; err != nil {
 			return xerrors.Errorf("failed to update content in database: %w", err)
 		}
+
+		delete(cm.ZonesConsolidating, cont.ID)
```
en0ma, line 430: @neelvirdy add a comment above this - `if the content is a consolidated aggregate, it means all its children have been moved and the content pinned, so remove consolidation flag`.

Good catch!

en0ma: Also, let us stop consolidation altogether where we can avoid it - it's prone to bugs, and it is an expensive process. 

We can discuss this later.

contentmgr/replication.go
```diff
@@ -160,15 +160,7 @@ func (cm *ContentManager) newContentStagingZone(user uint, loc string) (*Content
...
-		MaxSize:    cm.cfg.Content.MaxSize,
-		User:       user,
-		ContID:     content.ID,
-		Location:   content.Location,
-		Readiness:  stagingZoneReadiness{false, "Readiness not yet evaluated"},
-	}, nil
+	return content, nil
 }
 
 func (cb *ContentStagingZone) updateReadiness() {
```
en0ma, line 166: do we still need this?

contentmgr/replication.go
```diff
@@ -687,42 +646,57 @@ func (cm *ContentManager) CreateAggregate(ctx context.Context, conts []util.Cont
...
+		Find(&storedZoneSizes).Error; err != nil {
 		return err
 	}
 
-	zones := make(map[uint][]*ContentStagingZone)
-	for _, c := range stages {
-		var inZones []util.Content
-		if err := cm.DB.Find(&inZones, "aggregated_in = ?", c.ID).Error; err != nil {
-			return err
-		}
```
en0ma, line 703: add an if len(storedZoneSizes) > 0

contentmgr/replication.go
```diff
@@ -687,42 +646,57 @@ func (cm *ContentManager) CreateAggregate(ctx context.Context, conts []util.Cont
...
+}
+
+func (cm *ContentManager) recomputeStagingZoneSizes() error {
+	cm.log.Info("recomputing staging zone sizes .......")
 
-	var stages []util.Content
-	if err := cm.DB.Find(&stages, "not active and pinning and aggregate").Error; err != nil {
+	var storedZoneSizes []zoneSize
+	if err := cm.DB.Model(&util.Content{}).
+		Where("not active and pinning and aggregate").
```
en0ma, line 659: only select does with size=0

en0ma: this way, recompute is safe to run multiple times

contentmgr/gc.go
```diff
@@ -147,6 +146,24 @@ func (cm *ContentManager) UnpinContent(ctx context.Context, contid uint) error {
...
+	if pin.AggregatedIn > 0 {
+		if cm.ZonesConsolidating[pin.AggregatedIn] {
+			return fmt.Errorf("unable to unpin content while zone is consolidating (pin: %d, zone: %d)", pin.ID, pin.AggregatedIn)
+		}
+		var zone util.Content
+		if err := cm.DB.First(&zone, "id = ?", pin.AggregatedIn).Error; err != nil {
+			return err
+		}
+		if zone.Active {
+			return fmt.Errorf("unable to unpin content belonging to a pinned aggregate (pin: %d, zone: %d)", pin.ID, pin.AggregatedIn)
```
en0ma, line 158: no need to error, return nil

contentmgr/replication.go
```diff
@@ -160,99 +160,36 @@ func (cm *ContentManager) newContentStagingZone(user uint, loc string) (*Content
...
-	// TODO: propagate reason for failing to add content to bucket to end user
-
+func (cm *ContentManager) tryAddContent(zone util.Content, c util.Content) (bool, error) {
 	// if this bucket is being consolidated, do not add anymore content
-	if cb.IsConsolidating {
-		return false, nil
-	}
-
-	if cb.CurSize+c.Size > cb.MaxSize {
+	if cm.ZonesConsolidating[zone.ID] {
```
en0ma, line 168: dont you need to lock before check?

en0ma: yeah, you need the lock for each content add

neelvirdy: tryAddContent is only called while the bucketLk is already locked (https://github.com/application-research/estuary/pull/609/files#diff-20923025e6ed8064607b87fb913169095c8e675a484359bd37581febefc8295aL841)

can comment it's assuming the lock is already locked

but i added a couple other references to ZonesConsolidating that i need to wrap with the lock, will do now

contentmgr/replication.go
```diff
@@ -841,24 +809,42 @@ func (cm *ContentManager) addContentToStagingZone(ctx context.Context, content u
 	cm.BucketLk.Lock()
 	defer cm.BucketLk.Unlock()
 
-	blist, ok := cm.Buckets[content.UserID]
-	if !ok {
-		b, err := cm.newContentStagingZone(content.UserID, content.Location)
+	var zones []util.Content
+	if err := cm.DB.Find(&zones, "not active and pinning and aggregate and user_id = ?", content.UserID).Error; err != nil {
```
en0ma, line 814: cant we do something like this `not active and pinning and aggregate and user_id =? and size + ? <= ?", user_id, content.Size, max content`

contentmgr/replication.go
```diff
@@ -841,24 +809,42 @@ func (cm *ContentManager) addContentToStagingZone(ctx context.Context, content u
...
 		}
-
-		cm.Buckets[content.UserID] = []*ContentStagingZone{b}
 		return nil
 	}
 
-	for _, b := range blist {
-		ok, err := cm.tryAddContent(b, content)
+	// find zones without enough capacity and exclude them as zone options
+	var inviableZoneIDs []uint
```
en0ma, line 831: no need for this if we can do https://github.com/application-research/estuary/pull/609/files#r1036184580

contentmgr/replication.go
en0ma: > tryAddContent is only called while the bucketLk is already locked

Ah, I see it now, nice - ignore my comment

contentmgr/gc.go
```diff
@@ -147,6 +147,7 @@ func (cm *ContentManager) UnpinContent(ctx context.Context, contid uint) error {
 	}
 
 	if pin.AggregatedIn > 0 {
+		cm.BucketLk.Lock()
```
en0ma, line 150: defer the unlock from here

contentmgr/replication.go
```diff
@@ -2521,7 +2522,9 @@ func (cm *ContentManager) migrateContentToLocalNode(ctx context.Context, cont ut
 		return err
 	}
 
+	cm.BucketLk.Lock()
```
en0ma, line 2525: no need for this lock, its safe

contentmgr/replication.go
neelvirdy: yep good catch! this could be simplified even further after the incremental size tracking

contentmgr/pinning.go
```diff
@@ -428,7 +428,9 @@ func (cm *ContentManager) handlePinningComplete(ctx context.Context, handle stri
 		}
 
 		// if the content is a consolidated aggregate, it means all its children have been moved and the content pinned, so remove consolidation flag
+		cm.BucketLk.Lock()
```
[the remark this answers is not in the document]
neelvirdy: isn't the delete a write operation though? i might prefer to have the lock there if we're not sure, until we have some better tests that help us validate whether or not it makes a difference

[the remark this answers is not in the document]
en0ma: ok

contentmgr/gc.go
```diff
@@ -147,6 +146,27 @@ func (cm *ContentManager) UnpinContent(ctx context.Context, contid uint) error {
 		return err
 	}
 
+	if pin.AggregatedIn > 0 {
```
en0ma, line 149: @neelvirdy move this logic down, it should be done after the blocks are removed

contentmgr/replication.go
```diff
@@ -297,17 +234,33 @@ func (cm *ContentManager) ToCheck(contID uint) {
...
+
 func (cm *ContentManager) runStagingBucketWorker(ctx context.Context) {
 	timer := time.NewTicker(cm.cfg.StagingBucket.AggregateInterval)
 	for {
 		select {
 		case <-timer.C:
 			cm.log.Debugw("content check queue", "length", len(cm.queueMgr.queue.elems), "nextEvent", cm.queueMgr.nextEvent)
 
-			zones := cm.popReadyStagingZone()
-			for _, z := range zones {
```
en0ma, line 308: @neelvirdy Initially, we pop the zone out so new content is not added when aggregating. Now that we do not pop, we need to track it, or else content will be added to an ongoing aggregation which might lead to an unpredictable outcome. 

I say we change the `ZonesConsolidating` map to something like `processingZones` and use that to track both aggregation and consolidation

contentmgr/gc.go
neelvirdy: i moved all contents table updates below blocks are removed, and put them in one transaction. lmk if you were thinking of something else. i think doing all the unpinning DB updates in one transaction then deleting blocks would be a bit cleaner so the end user sees the delete happen and we can always GC the deleted blocks later if that ends up failing. but would consider that for a separate PR

contentmgr/replication.go
neelvirdy: this was a bit tricky due to it being difficult to tell when consolidating is done if the destination is not the local node. i think there's still room for improvement there, but ideally we remove the concept of consolidation instead. 

"Processing" a ready zone can go through roughly 3 paths:
1. consolidation is done with the local node as destination
2. CMD_TakeContent will be sent to consolidate content to a non-local node
3. consolidation isn't needed and just aggregation will be done

In cases (1) and (2), another "processing" phase will have to happen and (1) or (2) might happen again, if more content was added to the zone. If not, (3) will happen and that will be the last time that zone is "processing" before it is set as active.

I only have a hotspot connection where I'm staying for the next few days so I wasn't able to test this thoroughly yet. But if it's something we can test in staging or any pre-release environment, would be worth digging into it to validate.

contentmgr/replication.go
```diff
@@ -837,8 +837,6 @@ func (cm *ContentManager) addContentToStagingZone(ctx context.Context, content u
 	}
 
 	cm.log.Debugf("adding content to staging zone: %d", content.ID)
-	cm.bucketLk.Lock()
-	defer cm.bucketLk.Unlock()
```
en0ma: why remove this?

en0ma: ah, I see you used another lock

contentmgr/replication.go
en0ma: Here are the scenarios (and why we need to track processing);

- A ready zone with all content in the same location - it will run aggregation. When aggregation is running, it is not safe to add new content and to have a concurrent run. This scenario currently uses the pin complete to tell when it is done.
- A ready zone with contents in different locations - it will run consolidation. When consolidation is running, it is safe to add new content, but not safe to have a concurrent run. The scenario can use the worker re-run to check the completion by checking the location of the content and remove the flag if all are in the same location (consolidated), then proceed to aggregation

In either case, we need to mark an ongoing run and react accordingly (prevent re-run and adding new content for aggregates, prevent re-run for consolidation).

contentmgr/replication.go
```diff
@@ -550,6 +549,16 @@ func (cm *ContentManager) processStagingZone(ctx context.Context, zone util.Cont
...
 
+	// if not already processing, consolidation wasn't needed
+	if !cm.IsZoneProcessing(zone.ID) {
+		cm.MarkStartedProcessing(zone)
+	}
+	// if already processing, consolidation was needed but is now completed.
+	// we can keep this mark set and jump right into aggregation
+
+	// once aggregation is done, mark as finished processing
+	defer cm.MarkFinishedProcessing(zone)
```
en0ma, line 560: nope, aggregation is done when the zone content is pinned - either locally or when the incomplete events comes in from shuttle

en0ma: so remove markers where the pincomplete event is handled and where local aggregation is done

en0ma: lastly, how does this prevent duplicate runs for aggregation?

contentmgr/replication.go
```diff
@@ -168,7 +169,7 @@ func (cm *ContentManager) newContentStagingZone(user uint, loc string) (*util.Co
 
 func (cm *ContentManager) tryAddContent(zone util.Content, c util.Content) (bool, error) {
 	// if this bucket is being consolidated, do not add anymore content
-	if cm.IsZoneProcessing(zone.ID) {
+	if cm.IsZoneConsolidating(zone.ID) {
```
en0ma, line 172: or aggregated

contentmgr/replication.go
neelvirdy: should be fixed - pending testing with shuttles

pinning.go
```diff
@@ -498,6 +498,14 @@ func (s *Server) handleDeletePin(e echo.Context, u *util.User) error {
 		return err
 	}
 
+	if content.AggregatedIn > 0 {
+		if s.CM.IsZoneConsolidating(content.AggregatedIn) || s.CM.IsZoneAggregating(content.AggregatedIn) {
+			return fmt.Errorf("unable to unpin content while zone is consolidating or aggregating (pin: %d, zone: %d)", content.ID, content.AggregatedIn)
+		}
+		// even if the zone is pinned (active), we can allow the content to be unpinned
```
en0ma, line 505: this is not true, if it is a child of a pinned aggregate, do not delete it, as it will damage the aggregate

neelvirdy:
@en0ma I've validated that I can successfully consolidate and aggregate buckets across an API node and one shuttle. Currently the API node is not selectable as a consolidation destination in the code, so all the consolidations I tested went onto the shuttle. Included some restarts in the tests and tried adding data between consolidation and aggregation and confirmed it re-consolidated then eventually aggregated.

en0ma (approved):
LGTM

PR #603, opened 2022-11-28
API V2: /storage-providers endpoint

jcace:
Remove the `/public/miners` group of endpoints and work them into `/storage-providers` endpoints.

Closes #578

PR #614, opened 2022-11-30
/keys endpoint

### LINKED ISSUE #573: Create /keys endpoint
- /user/api-keys -> /keys
- GET /keys (list user's keys)
- DELETE /keys/:keyid (revoke key)
- POST /keys (create key)

toastts:
removed the `user/api-keys` endpoints in favor of `/keys`, updated their doc strings

closes #573

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -140,11 +140,15 @@ func (s *Server) ServeAPI() error {
 	e.GET("/get/:cid", s.handleGetFullContentbyCid)
 	// e.HEAD("/get/:cid", s.handleGetContentByCid)
 
+	//move /user route to separate /keys route, keep auth level
+	keys := e.Group("/keys")
+	keys.Use(s.AuthRequired(util.PermLevelUser))
+	keys.GET("", withUser(s.handleUserGetApiKeys))
+	keys.POST("", withUser(s.handleUserCreateApiKey))
+	keys.DELETE("/:keyid", withUser(s.handleUserRevokeApiKey))
+
 	user := e.Group("/user")
 	user.Use(s.AuthRequired(util.PermLevelUser))
-	user.GET("/api-keys", withUser(s.handleUserGetApiKeys))
-	user.POST("/api-keys", withUser(s.handleUserCreateApiKey))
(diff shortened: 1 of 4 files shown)


handlers.go
```diff
@@ -140,11 +140,15 @@ func (s *Server) ServeAPI() error {
...
+	//move /user route to separate /keys route, keep auth level
+	keys := e.Group("/keys")
+	keys.Use(s.AuthRequired(util.PermLevelUser))
+	keys.GET("", withUser(s.handleUserGetApiKeys))
+	keys.POST("", withUser(s.handleUserCreateApiKey))
+	keys.DELETE("/:keyid", withUser(s.handleUserRevokeApiKey))
+
 	user := e.Group("/user")
 	user.Use(s.AuthRequired(util.PermLevelUser))
-	user.GET("/api-keys", withUser(s.handleUserGetApiKeys))
```
alvin-reyes, line 145: These are breaking changes, do we intend to change the current v1 endpoints? I thought we're going to create a new set but will be grouped together on `api/v2` instead?

if so, instead of removing, we should add the following:

```
v2Group  := e.Group("api/v2") // this should be global V2 group for all new v2 endpoints.

userV2 := v2Group.Group("/user")
userV2.GET("", withUser(s.handleUserGetApiKeys))
userV2.POST("", withUser(s.handleUserCreateApiKey))
userV2.DELETE("/:keyid", withUser(s.handleUserRevokeApiKey))
```

We don't need to remove the previous one, just introduce a new group for V2 and call the same handler.

neelvirdy: +1 ^ i had this same confusion on https://github.com/application-research/estuary/pull/603#discussion_r1034042625.  can we wait for https://github.com/application-research/estuary/pull/596 to merge so we can make these changes to `dev` on the v2 API group instead of using the `api-v2` branch? cc @gmelodie @en0ma

gmelodie: we'll rebase them after #596 is merged. For now let's leave it this way

gmelodie (approved):

PR #615, opened 2022-11-30
/users endpoint api-v2

### LINKED ISSUE #574: Create /users endpoint
- remove /user/api-keys (see /keys issue)
- /user/stats -> /users/stats
- /user/export -> /users/export

toastts:
changed `/user/stats` to `/users/stats`
changed `/user/export` -> `/users/export`
updated docstrings to reflect changes

changed the handler for stats to have consistent naming semantics with the other handlers

resolves #574

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -151,7 +151,7 @@ func (s *Server) ServeAPI() error {
 	//move resources to plural users <-- (notice the s!!)
 	users := e.Group("/users")
 	users.Use(s.AuthRequired(util.PermLevelUser))
-	users.GET("/stats", withUser(s.handleGetUserStats))
+	users.GET("/stats", withUser(s.handleUserGetStats))
 	users.GET("/export", withUser(s.handleUserExportData))
 
 	userMiner := user.Group("/miner")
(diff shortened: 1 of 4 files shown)


handlers.go
```diff
@@ -145,10 +145,14 @@ func (s *Server) ServeAPI() error {
 	user.GET("/api-keys", withUser(s.handleUserGetApiKeys))
 	user.POST("/api-keys", withUser(s.handleUserCreateApiKey))
 	user.DELETE("/api-keys/:key_or_hash", withUser(s.handleUserRevokeApiKey))
-	user.GET("/export", withUser(s.handleUserExportData))
 	user.PUT("/password", withUser(s.handleUserChangePassword))
 	user.PUT("/address", withUser(s.handleUserChangeAddress))
```
gmelodie, line 149: We don't want to keep other `/user` endpoints

gmelodie (changes requested):
In order for this to get merged, I'd like to see all `/user` endpoints gone, either because they became `/users` or because they got moved somewhere else

handlers.go
gmelodie: @toastts any news here?

PR #630, opened 2022-11-30
Fix/pin panic

snissn:
Swap out the GOB encoder for msg pack as this doesn't panic. 
Also change the working directory for the queue since the data format is different now, so we don't have collisions

### DIFF
--- a/pinner/pinmgr.go
+++ b/pinner/pinmgr.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/application-research/estuary/pinner/types"
 	"github.com/application-research/goque"
+	"github.com/vmihailenco/msgpack/v5"
 
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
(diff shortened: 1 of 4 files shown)


pinner/pinmgr.go
```diff
@@ -356,8 +366,11 @@ func (pm *PinManager) enqueuePinOp(po *PinningOperation) {
 	if po.SkipLimiter {
 		u = 0
 	}
-
-	_, err = pm.pinQueue.EnqueueObject(getUserForQueue(u), po)
```
en0ma, line 360: @snissn can we add some form of validation on the addrs before enqueuing, so we don't fail and lose the message? Let's validate, if it fails validation, set the address to `[]`, so we still process the message, and ignore the bad addrs.

alvin-reyes: Yes. Lets prioritize this today and get it merged asap. I’d like to do an emergency release with this fix.

alvin-reyes: I’m OK with that. @en0ma?

Let's get this merged.

en0ma: ok

en0ma (approved):

PR #633, opened 2022-12-01
disable ar by default

anjor:


alvin-reyes:
If we agree to use the AR-standalone, we can switch this estuary option to false by default. The AR-standalone is just the same AR from estuary, pulled into an isolated service. We tested it last night and seems to work on render (background process).

https://github.com/application-research/ar-standalone.

alvin-reyes:
> If we agree to that (which to me makes sense, but I don't know enough about AR to have a strong opinion yet), then it sounds like we should delete the AR code in estuary instead of simply disabling it.

Yes. We’ll need @elijaharita and @gmelodie to confirm. Cc: @jlogelin

jlogelin (approved):
Agreed - For now, out of box estuary should have AR disabled by default

elijaharita:
Yes, we can pull this code out permanently

PR #641, opened 2022-12-01
Sanitize PinningOperation.Peers (don't allow nil ma.Multiaddr)

snissn:
sanitize Peers ( not allow nil ma.Multiaddr) in pinmgr.Add and include several tests for coverage to validate the sanitize method

@en0ma / @alvin-reyes can you validate the sanitize method does what we intend? if it needs modifications please let me know


alvin-reyes (approved):
LGTM

PR #649, opened 2022-12-03
use deal prices from config for miner selection

en0ma:


model/miner_ask.go
```diff
@@ -1,28 +1,19 @@
...
+	"github.com/application-research/estuary/config"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
 	"github.com/filecoin-project/lotus/chain/types"
 	"gorm.io/gorm"
 )
 
-var priceMax abi.TokenAmount
-
-func init() {
```
en0ma, line 12: removed this has it is already set for config at bootup here https://github.com/application-research/estuary/blob/dev/constants/constants.go#L40

alvin-reyes (approved):
LGTM

PR #650, opened 2022-12-04
Snissn/pinmgrtestpanic

snissn:
fix for pin mgr panic by stashing addr in a string based datastructure before serializing


Zorlin (approved):
LGTM, merge if tests pass and build works.

PR #653, opened 2022-12-05
check content-type on handleAdd early

anjor:
I was trying to upload a large dataset and it failed after nearly 7 minutes because I had the wrong content-type in my request:

```
root@c3-small-x86-01:/data# time curl https://upload.estuary.tech/content/add -H "Authorization: Bearer $(cat /data/.estuary_auth)" -H "Content-Type: multipar/form-data" -F "data=@longitudinal-nutrient-deficiency.tar"
{"error":{"code":500,"reason":"Internal Server Error","details":"request Content-Type isn't multipart/form-data"}}

real	6m52.221s
user	0m15.953s
sys	0m18.861s
```

Checking it early to handle similar instances.

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -1250,10 +1250,6 @@ func (s *Shuttle) handleLogLevel(c echo.Context) error {
 func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 	ctx := c.Request().Context()
 
-	if err := util.WithMultipartFormDataChecker(c.Request().Header); err != nil {
-		return err
-	}
-
 	if err := util.ErrorIfContentAddingDisabled(s.isContentAddingDisabled(u)); err != nil {
 		return err
 	}
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -313,7 +313,7 @@ func main() {
 			Value: cfg.Dev,
 		},
 		&cli.StringSliceFlag{
-			Name:  "announce-addr",
+			Name: "announce-addr",
 			Usage: "specify multiaddrs that this node can be connected to	",
 			Value: cli.NewStringSlice(cfg.Node.AnnounceAddrs...),
 		},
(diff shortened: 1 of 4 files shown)


util/content.go
```diff
@@ -213,3 +214,10 @@ func GetContent(contentid string, db *gorm.DB, u *User) (Content, error) {
 	}
 	return content, nil
 }
+
+func CheckContentTypeIsMultipartFormData(header http.Header) error {
```
en0ma, line 218: I was hoping you make it an actual HTTP middleware, like the others

anjor: will that work though? Because /content endpoint also has `handleAddIpfs` which does not need a multipart/form-data. 

Same is the case for `handleAddCar`

en0ma: @anjor this is an example I did for `car` endpoint - https://github.com/application-research/estuary/blob/dev/handlers.go#L161

handlers.go
```diff
@@ -857,7 +857,7 @@ func (s *Server) handleAdd(c echo.Context, u *util.User) error {
 	ctx, span := s.tracer.Start(c.Request().Context(), "handleAdd", trace.WithAttributes(attribute.Int("user", int(u.ID))))
 	defer span.End()
 
-	if err := util.CheckContentTypeIsMultipartFormData(c.Request().Header); err != nil {
+	if err := util.WithMultipartFormDataChecker(c.Request().Header); err != nil {
```
en0ma, line 860: no need for this anymore, the middleware is sufficient

cmd/estuary-shuttle/main.go
```diff
@@ -1250,7 +1250,7 @@ func (s *Shuttle) handleLogLevel(c echo.Context) error {
 func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
 	ctx := c.Request().Context()
 
-	if err := util.CheckContentTypeIsMultipartFormData(c.Request().Header); err != nil {
+	if err := util.WithMultipartFormDataChecker(c.Request().Header); err != nil {
```
en0ma, line 1253: no need for this anymore, the middleware is sufficient

en0ma (approved):
LGTM

en0ma (approved):

PR #655, opened 2022-12-06
Handle pin failures during consolidation to prevent stuck staging zones

neelvirdy:
Handle pin failures during consolidations by removing the failed contents from the staging zone and unmarking it as consolidating, so the staging worker will retry consolidation without the failed contents on the next interval.

Also, prevent staging zones from getting stuck in consolidating when the take content command doesn't even get successfully sent to the shuttle.

### DIFF
--- a/cmd/estuary-shuttle/rpc.go
+++ b/cmd/estuary-shuttle/rpc.go
@@ -144,6 +144,7 @@ func (d *Shuttle) addPin(ctx context.Context, contid uint, data cid.Cid, user ui
 			}); err != nil {
 				log.Errorf("failed to send pin status update: %s", err)
 			}
+			return nil
 		}
 
 		if !existing.Pinning && existing.Active {
--- a/cmd/estuary-shuttle/rpc.go
+++ b/cmd/estuary-shuttle/rpc.go
@@ -129,18 +129,21 @@ func (d *Shuttle) addPin(ctx context.Context, contid uint, data cid.Cid, user ui
 		existing := search[0]
 
 		if existing.Failed {
-			// retry a failed existing pin
-			op := &pinner.PinningOperation{
-				Obj:         data,
-				ContId:      contid,
-				UserId:      user,
-				Status:      types.PinningStatusQueued,
-				SkipLimiter: skipLimiter,
-				Peers:       peers,
-			}
+			// being asked to pin a thing we have marked as failed means the
+			// primary node isnt aware that this pin failed, we need to resend
+			// that notification
 
-			d.PinMgr.Add(op)
-			return nil
+			if err := d.sendRpcMessage(ctx, &drpc.Message{
+				Op: drpc.OP_UpdatePinStatus,
+				Params: drpc.MsgParams{
+					UpdatePinStatus: &drpc.UpdatePinStatus{
+						DBID:   contid,
+						Status: types.PinningStatusFailed,
+					},
+				},
+			}); err != nil {
+				log.Errorf("failed to send pin status update: %s", err)
+			}
 		}
 
 		if !existing.Pinning && existing.Active {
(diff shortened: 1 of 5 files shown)


cmd/estuary-shuttle/rpc.go
```diff
@@ -128,21 +129,17 @@ func (d *Shuttle) addPin(ctx context.Context, contid uint, data cid.Cid, user ui
...
-				Op: drpc.OP_UpdatePinStatus,
-				Params: drpc.MsgParams{
-					UpdatePinStatus: &drpc.UpdatePinStatus{
-						DBID:   contid,
-						Status: types.PinningStatusFailed,
-					},
-				},
-			}); err != nil {
-				log.Errorf("failed to send pin status update: %s", err)
+			// retry a failed existing pin
```
en0ma, line 132: no, we do not retry a failed pin. Simply tell Estuary it has failed, and Estuary should mark it (and remove it from consolidation and aggregation zones) and never resend a pin request again.

cmd/estuary-shuttle/rpc.go
```diff
@@ -358,7 +368,7 @@ func (s *Shuttle) handleRpcAggregateStagedContent(ctx context.Context, aggregate
 			return err
 		}
 
-		if !cont.Active || cont.Failed {
```
en0ma, line 361: if the content is failed the blocks will not be on this node, so you wont be able to aggregate it, that is why the failed check was there in the first place

cmd/estuary-shuttle/rpc.go
```diff
@@ -300,22 +297,35 @@ func (d *Shuttle) handleRpcTakeContent(ctx context.Context, cmd *drpc.TakeConten
...
+			} else {
+				log.Errorf("error finding pin: %s", err)
+				return err
+			}
 		}
 
-		go func(c drpc.ContentFetch) {
-			if err := d.addPin(ctx, c.ID, c.Cid, c.UserID, c.Peers, true); err != nil {
-				log.Errorf("failed to pin takeContent: %d", c.ID)
+		if pin.Active {
```
en0ma, line 315: only advantage of resending this will be to update the location if it was not pointing to this node

cmd/estuary-shuttle/rpc.go
```diff
@@ -300,22 +297,35 @@ func (d *Shuttle) handleRpcTakeContent(ctx context.Context, cmd *drpc.TakeConten
...
+			// if this content is already in this shuttle, resend pin complete
+			if err := d.resendPinComplete(ctx, pin); err != nil {
+				log.Error(err)
 			}
-		}(c)
+		}
+		if pin.Failed {
```
en0ma, line 321: we do not retry failed pins, instead, send a pin failure RPC message, so Estuary can mark the content as failed (as it should have been in the first place, probably did not have the right state due to dropped RPC message) and remove this content from the zones

cmd/estuary-shuttle/rpc.go
neelvirdy: @en0ma but in the case of adding a pin for consolidation, how would we unblock the staging zone from completing consolidation? and in the case i hit, this failure wasn't deterministic - i simply had to get the PinOp code path to run again and it succeeded fine. that makes me think that we need to either make adding a pin more deterministic, or have this type of retry mechanism.

en0ma: > but in the case of adding a pin for consolidation, how would we unblock the staging zone from completing consolidation?

We already addressed this in the last PR (where the worker reruns at x interval, it should always re-check the content location, if all are in the same place, consolidation is done).

> and in the case i hit, this failure wasn't deterministic

just like we do for pin completion, let us handle pin failure - remove this pin from the consolidation set, so that the next time worker runs, it will be aware of this change

neelvirdy: ok, i can close this PR out and try the failure handling approach. right now that's not in place so if a pin fails during consolidation the bucket does get stuck indefinitely in "consolidating", waiting for the consolidation to finish.

cmd/estuary-shuttle/rpc.go
```diff
@@ -143,7 +144,6 @@ func (d *Shuttle) addPin(ctx context.Context, contid uint, data cid.Cid, user ui
 			}); err != nil {
 				log.Errorf("failed to send pin status update: %s", err)
 			}
-			return nil
```
en0ma, line 146: why remove this?

en0ma (approved):
LGTM

PR #663, opened 2022-12-06
cleaner flag for splitContent

anjor:
I am not sure what this means for the existing data model though. Do we need to write a migration?

### DIFF
--- a/contentmgr/replication.go
+++ b/contentmgr/replication.go
@@ -1147,7 +1147,7 @@ func (cm *ContentManager) splitContent(ctx context.Context, cont util.Content, s
 		return fmt.Errorf("failed to load contents user from db: %w", err)
 	}
 
-	if !u.FlagSplitContent() {
+	if !u.SplitContent {
 		return fmt.Errorf("user does not have content splitting enabled")
 	}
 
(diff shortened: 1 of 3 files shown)


en0ma:
> I am not sure what this means for the existing data model though. Do we need to write a migration?

not sure what you want to achieve with this PR, but the `flags` is meant to be used for other things besides content splitting.  I do not think it should be removed

anjor:
I looked through the code (well, golang did) and didn't find any other usage for it.

Regardless, I think the code should be more readable than ```return u.Flags&8 != 0```

en0ma:
> I looked through the code (well, golang did) and didn't find any other usage for it.
> 
> Regardless, I think the code should be more readable than `return u.Flags&8 != 0`

We need the method to determine if split flag is set. I agree how it is determined needs to be readable. I don't think this is 

> I looked through the code (well, golang did) and didn't find any other usage for it.

it is currently only used for splitting, but it looks like a feature that will hold other flags

en0ma:
>  Regardless, I think the code should be more readable than `return u.Flags&8 != 0`

 We need the method to determine if split flag is set. I agree how it is determined needs to be readable. I think we should keep things as is, until we define what flags are and how we want to use them

> I looked through the code (well, golang did) and didn't find any other usage for it.
 
it is currently only used for splitting, but it looks like a feature that will hold other flags

anjor:
Yeah, I guess I am arguing we shouldn't overload Flags like that. It makes the code not readable.

en0ma:
> Yeah, I guess I am arguing we shouldn't overload Flags like that. It makes the code not readable.

totally, in its current form, I doubt it can handle more use cases

PR #671, opened 2022-12-06
don't exclude createdAt in content json responses

neelvirdy:
include `createdAt` in json so `/content/deals` includes it and the deals page can render the content's created at timestamp

related thread: https://filecoinproject.slack.com/archives/C016APFREQK/p1670347934023049


LucRoy (approved):
LGTM

PR #672, opened 2022-12-06
Move flags into flags.go and extract Setup/Run logic

neelvirdy:
this is a small start to making estuary setup and run cmds accessible in tests, as well as de-bloating main.go
1. move flags into a flags.go file
2. move bulk of Setup and Run logic out of action and into dedicated methods (for `./estuary setup` and `./estuary` cmds)


en0ma (approved):
LGTM

PR #674, opened 2022-12-06
properly handle peer addrInfo for pinner

en0ma:


snissn (approved):
looks good to me! handling the MultiAddr data types like they're toxic waste

anjor:
I haven't been paying close attention to this, but did this bug always exist? Why are we only seeing it now?

alvin-reyes (approved):
LGTM

en0ma:
> I haven't been paying close attention to this, but did this bug always exist? Why are we only seeing it now?

So the pinner queue was refactored to be persistent (save pin requests to disk as we receive them), and process them queued pins sequentially. Some pins' payload has peer values which are wonky formats, and because the peer struct is a bit complex (serializing it in its raw form can give unpredictable outcome when deserializing), so when the pinner manager encodes them and tries to decode them, it panics. 

This PR is to make the peer use a simple struct that will be consistent for both encode and decode

Issue #699, opened 2022-12-07
Strange behaviour with collections api


anjor:
here’s something funny that happens with the collections api:
created a collection, added some data to the collection with the directory path /dir1/dir2
```
anjor@seven ~ $ curl -k http://localhost:3004/collections/ -H "Authorization: Bearer $(cat ~/.estuary_local)" -d '{"name":"test-1", "description": ""}'
{"createdAt":"2022-12-07T23:00:26.767666Z","uuid":"b9f372de-0301-4fc3-b70d-53a140b06143","name":"test-1","description":"","userId":1,"cid":""}
anjor@seven ~ $ curl -k "http://localhost:3004/content/add?coluuid=b9f372de-0301-4fc3-b70d-53a140b06143&dir=/dir1/dir2" -H "Authorization: Bearer $(cat ~/.estuary_local)" -F "data=@a"
{"cid":"bafkreidame73g74ml6273rlrqp5vmii5s3e37nps6guduwlguei3k4ia54","retrieval_url":"https://dweb.link/ipfs/bafkreidame73g74ml6273rlrqp5vmii5s3e37nps6guduwlguei3k4ia54","estuary_retrieval_url":"https://api.estuary.tech/gw/ipfs/bafkreidame73g74ml6273rlrqp5vmii5s3e37nps6guduwlguei3k4ia54","estuaryId":85,"providers":["/ip4/192.168.86.28/tcp/6744/p2p/12D3KooWG7it68ksNLmK9SjSBEwTM68NHzgaKHUXcm21DhUAuo2k","/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWG7it68ksNLmK9SjSBEwTM68NHzgaKHUXcm21DhUAuo2k","/ip4/192.168.0.26/tcp/12210/p2p/12D3KooWG7it68ksNLmK9SjSBEwTM68NHzgaKHUXcm21DhUAuo2k","/ip4/86.14.245.42/tcp/12210/p2p/12D3KooWG7it68ksNLmK9SjSBEwTM68NHzgaKHUXcm21DhUAuo2k"]}
```

Now when I try to list all contents in the collection, i get this —

```
anjor@seven ~ $ curl -k "http://localhost:3004/collections/$coluuid" -H "Authorization: Bearer $(cat ~/.estuary_local)" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
[...]

PR #704, opened 2022-12-08
fix collections duplicate content bug

en0ma:
closes https://github.com/application-research/estuary/issues/699


anjor:
Test:

Before:

en0ma:
> If you only look for `/dir1/dir2` it doesn't show `a`, I would expect there to be a `/dir1/dir2/a` entry here as well, no?:

yeah

en0ma:
@anjor I have addressed it, and tested it;
- tested dir path that contains dir and file - works
- tested dir path that contains only file - works
- direct file path - works
- none existing/empty path - works (returns empty array now instead of null)

en0ma:
closing this for https://github.com/application-research/estuary/pull/705

PR #707, opened 2022-12-08
make pinner peer addr to be string

en0ma:
I initially used bytes for the peer ID as I could not get it to work with string. I figured we did not init the peer ID correctly that was stringifying it was failing. This PR changes it to a string, so now, the operation payload is a simple object.

I have verified pinning works on both API and shuttle nodes - peers are encoded and decoded correctly.

### DIFF
--- a/pinner/pinmgr_test.go
+++ b/pinner/pinmgr_test.go
@@ -2,6 +2,7 @@ package pinner
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"sync"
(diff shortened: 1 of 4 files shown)


pinner/pinmgr_test.go
```diff
@@ -67,41 +68,47 @@ func TestConstructMultiAddr(t *testing.T) {
...
+
+		var originsStr string
+		if origins != nil {
+			b, _ := json.Marshal(origins)
+			originsStr = string(b)
 		}
 
-		peer := []*peer.AddrInfo{{ID: peer.ID("12D3KooWCsxFFH242NZ4bjRMJEVc61La6Ha4yGVNXeEEwpf8KWCX"), Addrs: []ma.Multiaddr{addr}}}
-		po := &operation.PinningOperation{Peers: operation.SerializePeers(peer), Name: "pinning operation name"}
+		var origins2 []*peer.AddrInfo
```
neelvirdy, line 81: nit: `origins2` -> `originsUnmarshalled`

alvin-reyes (approved):
LGTM

neelvirdy (approved):

PR #718, opened 2022-12-08
feat: shuttle libp2p ping

### LINKED ISSUE #677: Add libp2p ping function to shuttles
Add a new RPC method to the Shuttles where they can be instructed to ping a given multiaddr/peerID, and return back an average RTT to reach that node.

jcace:
closes #677 

This PR adds the `PeerPingManager`, which will be used to ping SP's. The actual ping service / communication with API node / caching of data will be done in a separate PR.

### DIFF
--- a/cmd/estuary-shuttle/ping.go
+++ b/cmd/estuary-shuttle/ping.go
@@ -12,6 +12,11 @@ import (
 	"github.com/multiformats/go-multiaddr"
 )
 
+type PeerPingManager struct {
+	*Shuttle
+	Result PingManyResult
+}
+
 type Libp2pHost struct {
 	addr   multiaddr.Multiaddr
 	peerID peer.ID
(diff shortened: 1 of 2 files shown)


cmd/estuary-shuttle/ping.go
```diff
@@ -0,0 +1,76 @@
...
+
+	sort.SliceStable(result, func(i, j int) bool {
+		return p[result[i]] < p[result[j]]
+	})
+
+	return result[0:n]
+}
+
+func netPing(ctx context.Context, p peer.ID, h host.Host) (time.Duration, error) {
+	result, ok := <-ping.Ping(ctx, h, p)
```
neelvirdy, line 71: nit: we should be consistent with the parameter orders (not need to switch peer and host when calling Ping)

cmd/estuary-shuttle/ping.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &t, nil
+}
+
+// Returns a slice of the top-performing (lowest-latency) peers in the ping result.
+// Output will be truncated to the top `n`
+func GetTopPeers(p PingManyResult, n int) []peer.ID {
```
neelvirdy, line 56: can you make this callable on a `PingManyResult`? so in the end you would use this package like `s.PingMany(ctx, hosts).GetTopPeers(10)`

cmd/estuary-shuttle/ping.go
```diff
@@ -0,0 +1,76 @@
...
+	"time"
+
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/peerstore"
+	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
+	"github.com/multiformats/go-multiaddr"
+)
+
+type Libp2pHost struct {
```
neelvirdy, line 15: do you know where we're going to be constructing this shuttle-side yet? wondering if there's an existing struct that might be more convenient to use instead of us having to construct this on the caller. something like `AddrInfo` but seems like not exactly that

cmd/estuary-shuttle/ping.go
neelvirdy: actually, we probably want a PeerPingManager instance that each shuttle and the MinerManager tracks. so like `s.ppm.PingMany(hosts)` could be running on an interval as a subroutine spun off of each shuttle. and the ppm would store its own PingManyResults as the latest results, then you can send the ppm instance to the MinerManager so it can call `ppms[shuttleHandle].GetTopPeers(n)` in PickMiners

cmd/estuary-shuttle/ping.go
jcace: hmm, these two variables are present together in the [MinerInfo](https://github.dev/filecoin-project/lotus/blob/master/api/types.go#L292) struct, but it has a lot of other stuff too so it's kinda heavy. 

I can't see anywhere where the `peerID` and `MultiAddress` are sent to the shuttle via RPC, so we'd be adding something new for that anyways. Maybe we should define this struct elsewhere so it's more generally accessible in the Shuttle package?

en0ma (approved):
This is nice @jcace, I cant wait for the next PR

PR #733, opened 2022-12-09
move content markers and queue out of shuttle handler

en0ma:
This is the first PR to get queue messaging for IPC between shuttles and the API node to work. The shuttle handlers will no longer be aware of the content markers' state and content queue state. 

It also takes the markers closer to where there are needed, and they are driven by the staging worker that uses them.

### DIFF
--- a/contentmgr/pinning.go
+++ b/contentmgr/pinning.go
@@ -161,7 +161,7 @@ func (cm *ContentManager) PinContent(ctx context.Context, user uint, obj cid.Cid
 	if err != nil {
 		return nil, nil, 0, err
 	}
-	return ipfsRes, pinOp, 0, nil
+	return ipfsRes, pinOp, cont.ID, nil
 }
 
 func (cm *ContentManager) GetPinOperation(cont util.Content, peers []*peer.AddrInfo, replaceID uint, makeDeal bool) *operation.PinningOperation {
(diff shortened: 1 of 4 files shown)


contentmgr/pinning.go
```diff
@@ -89,24 +89,24 @@ func (cm *ContentManager) PinDelegatesForContent(cont util.Content) []string {
...
 	if err != nil {
-		return nil, nil, xerrors.Errorf("selecting location for content failed: %w", err)
+		return nil, nil, 0, xerrors.Errorf("selecting location for content failed: %w", err)
 	}
 
 	if replaceID > 0 {
 		// mark as replace since it will removed and so it should not be fetched anymore
 		if err := cm.db.Model(&util.Content{}).Where("id = ?", replaceID).Update("replace", true).Error; err != nil {
-			return nil, nil, err
+			return nil, nil, 0, err
```
alvin-reyes, line 101: Why did we need 0 here?

contentmgr/pinning.go
```diff
@@ -317,23 +317,26 @@ func (cm *ContentManager) UpdatePinStatus(location string, contID uint, status t
 			return errors.Wrap(err, "failed to look up content")
 		}
 
+		// if content is already active, ignore it
 		if c.Active {
-			return fmt.Errorf("got failed pin status message from location: %s where content(%d) was already active, refusing to do anything", location, contID)
+			return nil
```
alvin-reyes, line 322: Did we remove this because of the logs or we just really want to ignore the active ones? Is there a way we can add this on the query filter before it gets here?

alvin-reyes (approved):
LGTM

contentmgr/pinning.go
en0ma: null version of int

contentmgr/pinning.go
en0ma: This should not be happening, it means we sent a duplicate RPC message. If the pin is already active, no need to reactivate it. Better to ignore than complain, even though it is not actionable.

PR #736, opened 2022-12-10
fix: by-cid endpoint to check version explicitly

jcace:
The `/verify/by-cid` endpoint now properly determines the CID version based on the prefix, instead of checking the codec and length. 

The previous approach caused it to mistakenly determine that it was V1 if the length was `32` as the condition was `|| length = 32`


alvin-reyes (approved):
LGTM

PR #744, opened 2022-12-11
refactor to use nsq and websocket

en0ma:
to use a queue for RPC, just enable and specify the host and number of consumers per topic like so;

```
--queue-eng-enabled --queue-eng-host=ip:4150 --queue-eng-consumers=1
```

The queue engine is disabled by default and must be explicitly enabled on both API and shuttle to use it. Websocket is enabled by default and will work as is, except a queue engine is enabled on both shuttle and API. 

For backward compatibility, if;
- queue engine is enabled on api and disabled on shuttle, websocket be used for rpc calls
- queue engine is enabled on api and enabled on shuttle, queue(nsq by default) will be used for rpc calls
- queue engine is disabled on api and disabled on shuttle, websocket will be used for rpc calls
- queue engine is disabled on api and enabled on shuttle, websocket will be used for rpc calls

So queue is only used when both shuttle and API have it enabled.


alvin-reyes (approved):
Lets change nsq ip to nsq.estuary.tech.

LGTM

PR #746, opened 2022-12-11
Refactor pin estuary native: part 1

anjor:


anjor:
Check https://github.com/application-research/estuary/pull/746/commits/e2fe0ea031361717b5bd126a5d819cca51d258ea to see that the two codepaths are the same, at which point I decided to delete `handleAddIpfs` altogether.

anjor:
Hmm, looks like swagger just deletes the whole thing if I don't keep the function + godocs. I am happy to go either way -keep `handleAddIpfs` that just makes a call to `handleAddPin`, or delete it the way I have. Would love some feedback.

anjor:
Discussed offline with @en0ma, we are deleting this endpoint in api v2. In the meantime, just delegating the call to handleAddPin.

en0ma (approved):

PR #762, opened 2022-12-13
fix/chore: PinManager instance check

alvin-reyes:
this causes panics on the shuttles. if we are checking the op, we should also check the pm.


en0ma:
> this causes panics on the shuttles. if we are checking the op, we should also check the pm.

It is weird though, I don't see how it wont be initialized

PR #764, opened 2022-12-13
Enforce idea that there is only one connection between api node and a shuttle at a time

snissn:
Based on a thread in slack about a memory usage issue, I perceived a problem that could occur if a shuttle tries to connect multiple times to an api node. This PR enforces the idea that there's only one connection per shuttle+api node pair.

Please carefully think about this PR as this part of the code is less familiar to me.

Thanks

### DIFF
--- a/shuttle/rpc/engines/websocket/websocket.go
+++ b/shuttle/rpc/engines/websocket/websocket.go
@@ -2,6 +2,7 @@ package websocket
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/url"
 	"sync"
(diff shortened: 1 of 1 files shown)


shuttle/rpc/engines/websocket/websocket.go
```diff
@@ -107,9 +108,18 @@ func (m *manager) Connect(ws *websocket.Conn, handle string, done chan struct{})
 		QueueEngEnabled:       hello.QueueEngEnabled,
 	}
 
+	err = nil
 	m.shuttlesLk.Lock()
-	m.shuttles[handle] = sc
```
en0ma, line 111: if it already exists, would this not replace it? also, we do not want to fail, we want to replace it with the latest connection. Except we are saying `m.shuttles[handle] = sc` is not garbage collected

snissn: Makes sense. Instead of this code change should we call a clean up command on the old connection we are replacing?

en0ma: something like that, I was of the assumption just setting the same handle does garbage collection

en0ma: lets try out bytes first

en0ma:
@snissn we can close this

PR #765, opened 2022-12-13
use bytes for websocket message

en0ma:
This should address the WebSocket cid mem leak issue


en0ma:
closing this, need to separate the fixes from websocket

PR #772, opened 2022-12-14
Add /content/contents endpoint, include PinningStatus in Content

neelvirdy:
1. Add `/content/contents` endpoint that is like `/content/stats` but returns actual `Content`s instead of `statsResp`
2. Include `PinningStatus` in the Content struct so it is accessible from both the `/content/contents` endpoint and the `/content/staging-zones/:staging_zone/contents` endpoint
3. Adds "offloaded" `PinningStatus`

enables https://github.com/application-research/estuary-www/pull/136


en0ma (approved):
LGTM

PR #773, opened 2022-12-14
fix: call the queueManager checker from a go routine

alvin-reyes:


jlogelin (approved):
LGTM

en0ma:
Nice! additionally, we should only add to the queue if the size is above min deal size. No point adding anything less. Lastly, we should not be using this queue at all, but that will come with the content improvement work

PR #782, opened 2022-12-15
handle nsq pin update, pin complete errors

en0ma:


shuttle/rpc/engines/queue/drivers/nsq/nsq.go
```diff
@@ -67,8 +67,9 @@ func registerEstuaryConsumers(log *zap.SugaredLogger, cfg *config.Estuary, handl
 
 				if err := handlerFn(msg, "queue"); err != nil {
 					log.Errorf("failed to process message: %s with error: %s", msg.Op, err)
+				} else {
+					nsqMsg.Finish()
```
en0ma, line 71: this will allow for retry

shuttle/rpc/rpc.go
```diff
@@ -239,7 +238,8 @@ func (m *manager) SendRPCMessage(ctx context.Context, handle string, cmd *rpceve
 		m.log.Debugf("sending rpc message: %s, to shuttle: %s using websocket engine", cmd.Op, handle)
 		return d.SendMessage(ctx, cmd)
 	}
-	return websocketeng.ErrNoShuttleConnection
+	m.log.Warnf("failed to sending rpc message: %s, to shuttle: %s - shuttle is offline", cmd.Op, handle)
```
en0ma, line 241: no need to fail here

shuttle/rpc/rpc.go
```diff
@@ -293,7 +293,11 @@ func (m *manager) handlePinUpdate(location string, contID uint, status types.Pin
 
 		var c util.Content
 		if err := m.db.First(&c, "id = ?", contID).Error; err != nil {
-			return errors.Wrap(err, "failed to look up content")
+			if !xerrors.Is(err, gorm.ErrRecordNotFound) {
+				return xerrors.Errorf("failed to look up content: %d (shuttle = %s): %w", contID, location, err)
+			}
+			m.log.Warnf("content: %d not found for pin update from shuttle: %s", contID, location)
```
en0ma, line 299: if no content, warn but do not fail - it is not actionable

shuttle/rpc/rpc.go
```diff
@@ -327,7 +331,11 @@ func (m *manager) handlePinningComplete(ctx context.Context, handle string, pinc
 
 	var cont util.Content
 	if err := m.db.First(&cont, "id = ?", pincomp.DBID).Error; err != nil {
-		return xerrors.Errorf("got shuttle pin complete for unknown content %d (shuttle = %s): %w", pincomp.DBID, handle, err)
+		if !xerrors.Is(err, gorm.ErrRecordNotFound) {
+			return xerrors.Errorf("failed to look up content: %d (shuttle = %s): %w", pincomp.DBID, handle, err)
+		}
+		m.log.Warnf("content: %d not found for pin complete from shuttle: %s", pincomp.DBID, handle)
```
en0ma, line 337: if no content, warn but do not fail - it is not actionable

shuttle/rpc/rpc.go
neelvirdy: is this assuming that any RPC message we send has a retry mechanism elsewhere, and none of the callers would want to error handle?

i'm thinking of the consolidation commands where we would want to know that the message failed to send so we can mark the zone as not-consolidating anymore. otherwise it wouldn't retry because it's perpetually "in consolidation" already

en0ma: @neelvirdy it is retried. I will revert this part, we can tackle it later.

en0ma:
closing this as it will come with the staging zone pr

PR #785, opened 2022-12-15
Backwards compat for /add-ipfs filename param, cid as fallback

neelvirdy:
Addresses https://filecoinproject.slack.com/archives/C016APFREQK/p1671110132518029?thread_ts=1671097233.969249&cid=C016APFREQK

### DIFF
--- a/api/v1/pinning.go
+++ b/api/v1/pinning.go
@@ -255,6 +255,17 @@ func (s *apiV1) handleAddPin(e echo.Context, u *util.User) error {
 		return err
 	}
 
+	// backwards compatibility to accept filename queryParam if name is not provided
+	if filename := e.QueryParam("filename"); filename != "" {
+		if pin.Name == "" {
+			pin.Name = filename
+		}
+	}
+	// if name is still empty, use cid as name
+	if pin.Name == "" {
+		pin.Name = pin.CID
+	}
+
 	var cols []*collections.CollectionRef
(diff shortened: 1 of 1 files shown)


api/v1/pinning.go
```diff
@@ -255,6 +255,17 @@ func (s *apiV1) handleAddPin(e echo.Context, u *util.User) error {
 		return err
 	}
 
+	// backwards compatibility to accept filename queryParam if name is not provided
```
anjor, line 258: Let's make this change in handlers.go in `handleAddIpfs` instead of in the pinning code.

neelvirdy: i want this in `handleAddPin` so the `/add-ipfs` and `/pins` endpoints remain on the same codepath. and generally we should be trying to move handlers out of `handlers.go` and into `api/v1/<group>.go`

anjor: handleAddIpfs is legacy and needs to be backward compatible. Adding this codepath here introduces new functionality to handleAddPin -- I don't think we should do that.

anjor: Given that in api v2 handleAddIpfs will be removed anyway I think it makes sense to have the backward compat logic there instead of the pinning logic.

neelvirdy: i'd rather not do any backend change then and just let the frontend PR address the issue of a blank name in UI. gonna close this out

PR #786, opened 2022-12-15
Composite index for primaryStagingLocation on contents

neelvirdy:
~90% of our DB time is spent on this query:
image

Link to specific query analytics: https://app.pganalyze.com/databases/-675815701/queries/5815925939?t=7d

Query itself:
image

This gets called by `primaryStagingLocation`: https://github.com/application-research/estuary/blob/a427628ba8dfd0cc71cf6f8099d539e885513091/shuttle/location.go#L82

Gorm composite index docs: https://gorm.io/docs/indexes.html#Composite-Indexes

By adding this composite index, we can avoid doing db scans on this query and it should become much faster, freeing up DB time.

### DIFF
--- a/util/content.go
+++ b/util/content.go
@@ -59,7 +59,7 @@ type Content struct {
 
 	Cid         DbCID       `json:"cid"`
 	Name        string      `json:"name"`
-	UserID      uint        `json:"userId" gorm:"index"`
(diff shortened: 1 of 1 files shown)


neelvirdy:
converting to draft since adding this index to a large table will likely require manual intervention first, instead of an auto migration.

it should still eventually be put into code so other nodes can pick up the composite index

en0ma:
@neelvirdy has this been addressed?

PR #787, opened 2022-12-15
simplify and defer primaryStagingLocation

neelvirdy:
1. query StagingZone table for primaryStagingLocation before querying contents table
2. primaryStagingLocation doesn't need to use an array since it calls First
3. avoid calling primaryStagingLocation until needed

(1) should address excessive db time spent on this query https://app.pganalyze.com/databases/-675815701/queries/5815925939?t=7d

### DIFF
--- a/cmd/estuary-shuttle/data.go
+++ b/cmd/estuary-shuttle/data.go
@@ -8,7 +8,7 @@ import (
 )
 
 type Pin struct {
-	ID        uint      `gorm:"primarykey" json:"id"`
+	ID        uint      `gorm:"primaryKey" json:"id"`
 	CreatedAt time.Time `json:"-"`
 	UpdatedAt time.Time `json:"-"`
 
@@ -35,7 +35,7 @@ type Pin struct {
 }
 
 type Object struct {
-	ID   uint       `gorm:"primarykey"`
+	ID   uint       `gorm:"primaryKey"`
 	Cid  util.DbCID `gorm:"index"`
 	Size int
 	//Reads      int
@@ -43,7 +43,7 @@ type Object struct {
 }
 
 type ObjRef struct {
-	ID     uint `gorm:"primarykey"`
+	ID     uint `gorm:"primaryKey"`
 	Pin    uint `gorm:"index"`
 	Object uint `gorm:"index"`
 	//Offloaded bool
--- a/cmd/estuary-shuttle/data.go
+++ b/cmd/estuary-shuttle/data.go
@@ -8,7 +8,7 @@ import (
 )
 
 type Pin struct {
-	ID        uint      `gorm:"primaryKey" json:"id"`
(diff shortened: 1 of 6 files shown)


cmd/estuary-shuttle/data.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 type Pin struct {
-	ID        uint      `gorm:"primarykey" json:"id"`
+	ID        uint      `gorm:"primaryKey" json:"id"`
```
en0ma: does this have any side effects on already created tables?

content/pinning.go
```diff
@@ -117,6 +112,11 @@ func (cm *ContentManager) PinContent(ctx context.Context, user uint, obj cid.Cid
 		originsStr = string(b)
 	}
 
+	loc, err := cm.shuttleMgr.GetLocationForStorage(ctx, obj, user)
```
en0ma, line 114: nice

shuttle/location.go
```diff
@@ -84,17 +84,14 @@ func (m *manager) GetLocationForStorage(ctx context.Context, obj cid.Cid, uid ui
...
-		return ""
-	}
-
 	// TODO: maybe we could make this more complex, but for now, if we have a
 	// staging zone opened in a particular location, just keep using that one
-	for _, z := range zones {
-		return z.Location
+
+	var zone model.StagingZone
+	if err := cm.db.First(&zone, "user_id = ? and status = ?", uid, model.ZoneStatusOpen).Error; err != nil {
```
en0ma, line 93: this is nice, but not backward compatible, also, the zone is still being reconstructed. Best way would be to get the location of the most current content (desc). Also, we need to handle - not found

en0ma:
@neelvirdy I think this is a good fix, lets create a separate PR for the `simplify and defer primaryStagingLocation` part and merge first

cmd/estuary-shuttle/data.go
neelvirdy: it shouldn't, but will separate into another PR

shuttle/location.go
neelvirdy: i updated this to fallback to the old location logic when a StagingZone record is not found - does that address the backwards compatibility? i figure at some point we'll want to remove this fallback since it's relatively expensive and shouldn't be needed once the reconstruction process is done and those zones are aggregated

if there's no existing zone, i think it's correct to return empty string, since the caller will then pick the preferred shuttle from the shuttles table for which location to use next

neelvirdy:
> @neelvirdy I think this is a good fix, lets create a separate PR for the `simplify and defer primaryStagingLocation` part and merge first

@en0ma i removed the gorm changes from this PR instead of creating a separate PR, it should be ready for re-review now

en0ma (approved):
LGTM

PR #789, opened 2022-12-15
Use ExpirableLRU to cache http requests with TTL

neelvirdy:
1. Adds two `ExpirableLRU` caches for caching certain http requests with two options of TTL (10 seconds, 60 minutes).
2. Uses the 10 second cache for: `/content/contents`, `/content/staging-zones/*`, part of `/public/stats`
3. Uses the 60 minute cache for: rest of `/public/stats`, `/public/metrics/deals-on-chain`

Note: I only included the 10 second cache for apiV2 because the extended cache endpoints should be moved to metrics API, and I'm not expecting them to be rewritten under apiV2. So, there's no need for the extended cache there.


alvin-reyes (dismissed):
LGTM

alvin-reyes (approved):
LGTM

PR #797, opened 2022-12-16
refactor shuttle connection

en0ma:
This will persist shuttle connection details so that commp can be calculated and deals can flow. Also, it is much better in flaky websocket connections.

### DIFF
--- a/api/v1/handlers.go
+++ b/api/v1/handlers.go
@@ -59,7 +59,6 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/multiformats/go-multiaddr"
 	"github.com/pkg/errors"
-	"golang.org/x/net/websocket"
 	"golang.org/x/sys/unix"
 	"golang.org/x/xerrors"
 	"gorm.io/gorm"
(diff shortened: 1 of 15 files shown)


10d9e (dismissed):
LGTM

api/v1/handlers.go
```diff
@@ -5114,7 +5112,12 @@ func (s *apiV1) checkGatewayRedirect(proto string, cc cid.Cid, segs []string) (s
 		return "", nil
 	}
 
-	if !s.shuttleMgr.IsOnline(cont.Location) {
+	isOnline, err := s.shuttleMgr.IsOnline(cont.Location)
```
anjor, line 5115: nice

cmd/estuary-shuttle/main.go
```diff
@@ -943,12 +942,6 @@ func (d *Shuttle) runRpc(conn *websocket.Conn) (err error) {
 		return err
 	}
 
-	var hi rpcevent.Hi
```
anjor, line 946: this is no longer necessary?

en0ma: yeah

content/replication.go
```diff
@@ -205,8 +205,14 @@ func (cm *ContentManager) ensureStorage(ctx context.Context, content util.Conten
 
 	// after reconciling content deals,
 	// check If this is a shuttle content and that the shuttle is online and can start data transfer
-	if content.Location != constants.ContentLocationLocal && !cm.shuttleMgr.IsOnline(content.Location) {
-		cm.log.Warnf("content shuttle: %s, is not online for deal making", content.Location)
+	isOnline, err := cm.shuttleMgr.IsOnline(content.Location)
+	if err != nil {
+		done(time.Minute * 15)
```
anjor, line 210: Not sure what 15 minutes is, and looks like we use it below as well. Should we extract it as a variable?

en0ma: lol, this part of content manager will be extracted to a different package. That timing will not be used anymore. I did not just added it, so I don't want to modify it

10d9e (approved):
LGTM

PR #799, opened 2022-12-16
getStagedContentsGroupedByLocation improvement

en0ma:


content/staging.go
```diff
@@ -394,25 +354,58 @@ func (cm *ContentManager) CreateAggregate(ctx context.Context, conts []util.Cont
...
-		out[c.Location] = append(out[c.Location], c)
+	if dstLocation == "" {
+		return "", nil, 0, fmt.Errorf("zone: %d failed to consolidate as no destination location could be determined", zoneContID)
 	}
 
-	if len(out) == 0 {
-		return nil, fmt.Errorf("no location for staged contents")
+	var toMove []util.Content
+	var toMoveBatch []util.Content
+	if err := cm.db.Where("active and aggregated_in = ? and location <> ?", zoneContID, constants.ContentLocationLocal).FindInBatches(&toMoveBatch, 500, func(tx *gorm.DB, batch int) error {
```
neelvirdy: i dont think we want the location <> ? part here. we still need to move local contents to the shuttle that is going to be aggregating and making the deal

neelvirdy (changes requested):

content/staging.go
en0ma: ah right, not from shuttle to local, but local to shuttle

neelvirdy (approved):

10d9e (approved):
Looks good to me.

PR #808, opened 2022-12-17
exclude cmd directory for swagger entyr points so just the api endpoi…

### LINKED ISSUE #807: Build swagger docs from api endpoint and ignore shuttle endpoints
shuttle endpoints have duplicate but slightly different endpoints/swagger notation. this pr just builds for the main api node.

snissn:
…nts are used to generate documentation

previously we were using both the shuttle endpoints and the api endpoints to make swagger annotations.

I think generally the only issue is that there were duplicate annotations, however it could be possible that this has some unexpected side effects in the swagger.json generation.

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -865,19 +865,67 @@ const docTemplate = `{
         },
         "/content/add": {
             "post": {
-                "description": "This endpoint uploads a file.",
+                "description": "This endpoint is used to upload new content.",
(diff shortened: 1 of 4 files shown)


snissn:
the diff looks small, i think it's good, but a review would be appreciated!

alvin-reyes (approved):
LGTM

Issue #679, opened 2022-12-06
Create sp-locator scheduled task to run on shuttles


jcace:
Set up a task to run on the shuttles, that uses the PeerPingManager to ping all SP's and store the results in a list that's cached in memory. 

This task should run
- Whenever the shuttle first starts up
- On a regular, configurable interval (once every 12h should be a sensible starting point) 

If it makes sense, modify the SP Addition process so that new SPs are automatically pinged  when they are onboarded.

PR #813, opened 2022-12-17
feat: sp selection apis

### LINKED ISSUE #678: SP Selection - Shuttle <-> API Node build out communication / RPCs
Build out the RPCs / communication interfaces for the shuttles and API node to share SP preference data:

- Shuttles need a way to request a list of all SPs from the api node, so they know what to ping in the `PeerPingManager`
- API node needs a way to query the shuttle for its list of `preferred SPs`

jcace:
closes #678 

Adds an endpoint to the shuttle, and the API node to allow communication for the SP selection mechanism


cmd/estuary-shuttle/main.go
```diff
@@ -2442,3 +2446,15 @@ func (s *Shuttle) handleGetSystemConfig(e echo.Context) error {
...
 }
+
+func (s *Shuttle) handleStorageProviderList(e echo.Context) error {
+	n, err := strconv.Atoi(e.Param("n"))
+
+	if err != nil {
+		return err
+	}
+
+	resp := s.PPM.Result.GetTopPeers(n)
```
jcace, line 2505: right now, this might return an empty list if it's not populated yet, but will be addressed in #679

api/v1/handlers.go
```diff
@@ -1773,20 +1774,32 @@ type minerResp struct {
...
+// @Param        params           path      bool  false   "set true to include on-chain miner info"
+// @Router       /public/miners/{lookup_chain} [get]
 func (s *apiV1) handleAdminGetMiners(c echo.Context) error {
+	ctx := context.TODO()
 	var miners []model.StorageMiner
 	if err := s.DB.Find(&miners).Error; err != nil {
 		return err
 	}
 
+	lookupChain, _ := strconv.ParseBool(c.Param("lookup_chain"))
```
jcace: what do you guys think of this approach?

I think the `GetMinerChainInfo` call is fairly expensive as it looks up the addresses on-chain, so I added a param to prevent it being ran unless necessary

api/v1/handlers.go
```diff
@@ -1773,20 +1774,32 @@ type minerResp struct {
...
+
 	for i, m := range miners {
 		out[i].Addr = m.Address.Addr
 		out[i].Suspended = m.Suspended
 		out[i].SuspendedReason = m.SuspendedReason
 		out[i].Name = m.Name
 		out[i].Version = m.Version
+
+		if lookupChain {
+			ci, err := s.minerManager.GetMinerChainInfo(ctx, m.Address.Addr)
```
jcace: should we run in multiple threads to improve performance?

miner/miner.go
```diff
@@ -93,3 +103,28 @@ func (mm *MinerManager) GetDealProtocolForMiner(ctx context.Context, miner addre
 	}
 	return proto, nil
 }
+
+func (mm *MinerManager) GetMinerChainInfo(ctx context.Context, maddr address.Address) (*MinerChainInfo, error) {
```
jcace, line 107: should we cache the `miner - chainInfo` somewhere to reduce the on-chain lookups required?

Do we have an existing cache in place where we can put it?

api/v1/api.go
```diff
@@ -207,7 +207,7 @@ func (s *apiV1) RegisterRoutes(e *echo.Echo) {
 	netw.GET("/addrs", s.handleNetAddrs)
 
 	miners := public.Group("/miners")
-	miners.GET("", s.handleAdminGetMiners)
+	miners.GET(":lookup_chain", s.handleAdminGetMiners)
```
alvin-reyes: is this supposed to be under `/admin/miners`? My concern with putting this on /public is that someone can make unlimited calls to this endpoint which can affect the API's performance. We should instead cache the results and have an expiration of 1 hour.

api/v1/handlers.go
```diff
@@ -1773,20 +1774,32 @@ type minerResp struct {
 // @Success      200  {object}  string
 // @Failure      400           {object}  util.HttpError
 // @Failure      500           {object}  util.HttpError
-// @Router       /public/miners [get]
+// @Param        params           path      bool  false   "set true to include on-chain miner info"
+// @Router       /public/miners/{lookup_chain} [get]
 func (s *apiV1) handleAdminGetMiners(c echo.Context) error {
```
alvin-reyes, line 1750: I think this is for admin? If so we need to change the annotation.

api/v1/handlers.go
alvin-reyes: We can run this in a go but we should cache the results as well. We can wrap this on the cacher (the server object has an instance of this) and return the result. This way, we only need to get the miner chain info of the specific address directly from chain once every hour.

cmd/estuary-shuttle/main.go
alvin-reyes: We should evaluate if we can just cache this as well or not.

api/v1/api.go
jcace: Good catch, yes I meant to put it on `/admin`, will fix!

estuary.env
```diff
@@ -0,0 +1,8 @@
+set -a
+FULLNODE_API_INFO=wss://api.chain.love
+# FULLNODE_API_INFO=http://10.32.32.1:10000
+
+set +a
+
+
+# EST3edfbe00-2bf0-4c4f-ba5d-81bd0cb5ffd6ARY
```
neelvirdy: should we add this to `.gitignore`? also guessing this key is for your local estuary but flagging anyways

test-ar.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh
```
neelvirdy: should this be file here? maybe add it to the test scripts folder in a separate PR if you think it should be pushed

api/v1/handlers.go
neelvirdy: we can just use the cache instead of using a param

test-ar.sh
jcace: accidentally included this file - removed!

api/v1/handlers.go
jcace: Cached in `ExtendedCacher`

api/v1/handlers.go
```diff
@@ -1758,35 +1758,62 @@ func (s *apiV1) handleGetSystemConfig(c echo.Context, u *util.User) error {
...
 		out[i].Addr = m.Address.Addr
 		out[i].Suspended = m.Suspended
 		out[i].SuspendedReason = m.SuspendedReason
 		out[i].Name = m.Name
 		out[i].Version = m.Version
+
+		var ci *miner.MinerChainInfo
+
+		// Cache the Chain Lookup for this miner, looking up if it doesnt exist / is expired
+		key := cacheKey(c, nil, m.Address.Addr.String())
```
neelvirdy: i'd prefer to cache the whole request instead of per miner. that lets us avoid the db query and the loop per request, at the expense of new miners having to wait for the next cache expiry to be included (max 1 hour)

api/v1/handlers.go
```diff
@@ -1758,35 +1758,62 @@ func (s *apiV1) handleGetSystemConfig(c echo.Context, u *util.User) error {
...
+		cached, ok := s.extendedCacher.Get(key)
+		if ok {
+			cachedCi, ok := cached.(*miner.MinerChainInfo)
+			if ok {
+				ci = cachedCi
+			} else {
+				newCi, err := s.minerManager.GetMinerChainInfo(ctx, m.Address.Addr)
```
neelvirdy: this through adding to the cache should happen outside of the top-level "if ok" block

api/v1/handlers.go
```diff
@@ -1758,35 +1758,62 @@ func (s *apiV1) handleGetSystemConfig(c echo.Context, u *util.User) error {
...
+				if err != nil {
+					ci = newCi
+				}
+				// If the ChainInfo is `nil`, (not published), we should still cache it to ensure we don't keep looking it up
+				s.extendedCacher.Add(key, newCi)
+			}
+		}
+
+		if ci != nil {
```
neelvirdy: this check is unnecessary, it should be ok to set a nil chain info

api/v1/handlers.go
jcace: was going back and forth on this... yeah, good point! will update

gmelodie (commented):
These changes should be on `api/v2`. We should stop breaking api v1 since we have v2 setup already

jcace:
@gmelodie updated to include both v1 and v2, plus a minor refactor of the `cacheKey` util func.

api/v1/api.go
```diff
@@ -204,7 +204,6 @@ func (s *apiV1) RegisterRoutes(e *echo.Echo) {
 	netw.GET("/addrs", s.handleNetAddrs)
 
 	miners := public.Group("/miners")
-	miners.GET("", s.handleAdminGetMiners)
```
gmelodie: Shouldn't we leave the v1 as it was? Why are we removing this?

api/v2/storageproviders.go
```diff
@@ -288,35 +300,19 @@ type storageProviderChainInfo struct {
 // @Param        sp  path      string  true  "Filter by storage provider"
 // @Router       /storage-providers/stats/{sp} [get]
 func (s *apiV2) handleGetStorageProviderStats(c echo.Context) error {
-	ctx, span := s.tracer.Start(c.Request().Context(), "handleGetStorageProviderStats")
```
gmelodie: Shouldn't this be ` handleGetStorageProviderStats` on v2? Unless we want to use legacy names code-wise.

api/v2/storageproviders.go
```diff
@@ -288,35 +300,19 @@ type storageProviderChainInfo struct {
...
+	ctx, span := s.tracer.Start(c.Request().Context(), "handleGetMinerStats")
 	defer span.End()
 
-	maddr, err := address.NewFromString(c.Param("sp"))
+	maddr, err := address.NewFromString(c.Param("miner"))
```
gmelodie: On v2 this should be `sp`, not `miner`

cmd/estuary-shuttle/main.go
```diff
@@ -1210,6 +1210,10 @@ func (s *Shuttle) ServeAPI() error {
 	admin.GET("/net/rcmgr/stats", s.handleRcmgrStats)
 	admin.GET("/system/config", s.handleGetSystemConfig)
 
+	storageProvider := e.Group("/storage-provider")
+	storageProvider.Use(s.AuthRequired(util.PermLevelAdmin))
+	admin.GET("/list/:n", s.handleStorageProviderList)
```
gmelodie: I think this should be `storageProvider.GET`, not `admin.GET`

gmelodie (changes requested):

api/v1/api.go
jcace: good catch - shouldn't do this in case it breaks anything.

cmd/estuary-shuttle/main.go
jcace: good catch. No idea why I wrote admin here 😅

api/v1/handlers.go
gmelodie: If this is the endpoint below: https://github.com/application-research/estuary/blob/1ca72e29b4bcb36e22d66ec43bdacca7ba248bb8/api/v1/api.go#L210 Then the annotation should say `/public/miners` but the naming is confusing yes.

api/v1/handlers.go
```diff
@@ -2163,27 +2163,11 @@ func (s *apiV1) handleGetMinerStats(c echo.Context) error {
 		return err
 	}
 
-	minfo, err := s.Api.StateMinerInfo(ctx, maddr, types.EmptyTSK)
+	ci, err := s.minerManager.GetMinerChainInfo(ctx, maddr)
```
gmelodie, line 2172: Nice

api/v2/storageproviders.go
```diff
@@ -125,37 +124,58 @@ func (s *apiV2) handleStorageProvidersSetInfo(c echo.Context, u *util.User) erro
...
-// @Description  This endpoint returns all storage providers
+// @Description  This endpoint returns all storage providers. Note: Value may be cached
 // @Tags         sp
 // @Produce      json
 // @Success      200  {object}  []storageProviderResp
 // @Failure      400           {object}  util.HttpError
 // @Failure      500           {object}  util.HttpError
 // @Router       /v2/storage-providers [get]
 func (s *apiV2) handleGetStorageProviders(c echo.Context) error {
+	ctx := context.TODO()
```
gmelodie: Let's use a tracer here as well

```
ctx, span := s.tracer.Start(c.Request().Context(), "handleGetStorageProviders")
defer span.End()
```

cmd/estuary-shuttle/main.go
```diff
@@ -2483,3 +2487,15 @@ func setupMetrics(metCtx context.Context) Metrics {
 		SendingPeersPending:            metrics.NewCtx(metCtx, "graphsync_sending_peers_pending", "number of peers we can't send more data to cause of pending allocations").Gauge(),
 	}
 }
+
+func (s *Shuttle) handleStorageProviderList(e echo.Context) error {
```
gmelodie, line 2499: We should have swagger annotations for this

cmd/estuary-shuttle/main.go
```diff
@@ -2483,3 +2487,15 @@ func setupMetrics(metCtx context.Context) Metrics {
...
+	n, err := strconv.Atoi(e.Param("n"))
+
```
gmelodie: nit: remove blank line here

util/cache.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/labstack/echo/v4"
+)
+
+// cacheKey returns a key based on the request being made, the user associated to it, and optional tags
+// this key is used when calling Get or Add from a cache
+func CacheKey(c echo.Context, u *User, tags ...string) string {
```
gmelodie, line 12: nit/totally optional: maybe GenerateCacheKey would be a good name

api/v2/storageproviders.go
```diff
@@ -125,37 +124,58 @@ func (s *apiV2) handleStorageProvidersSetInfo(c echo.Context, u *util.User) erro
...
+
+	// Cache the Chain Lookup for this miner, looking up if it doesnt exist / is expired
+	key := util.CacheKey(c, nil)
+	cached, ok := s.extendedCacher.Get(key)
+	if ok {
+		out, ok := cached.([]storageProviderResp)
+		if ok {
```
gmelodie, line 152: I think we might want to return an error if this is `!ok`, since the key was found but I can't be cast (why? something cheeky might be happening)

api/v1/handlers.go
```diff
@@ -1743,37 +1730,58 @@ func (s *apiV1) handleGetSystemConfig(c echo.Context, u *util.User) error {
...
+// @Router       /admin/miners/ [get]
 func (s *apiV1) handleAdminGetMiners(c echo.Context) error {
+	ctx := context.TODO()
+
+	// Cache the Chain Lookup for this miner, looking up if it doesnt exist / is expired
+	key := util.CacheKey(c, nil)
+	cached, ok := s.extendedCacher.Get(key)
+	if ok {
+		out, ok := cached.([]minerResp)
+		if ok {
```
gmelodie, line 1758: Same as below, we might want to return an error if `!ok` here

api/v1/handlers.go
```diff
@@ -1743,37 +1730,58 @@ func (s *apiV1) handleGetSystemConfig(c echo.Context, u *util.User) error {
...
 }
 
 // handleAdminGetMiners godoc
 // @Summary      Get all miners
-// @Description  This endpoint returns all miners
-// @Tags         public,net
+// @Description  This endpoint returns all miners. Note: value may be cached
+// @Tags         admin,net
 // @Produce      json
 // @Success      200  {object}  string
```
gmelodie: I think this should say `// @Success 200 {object} minerResp` instead (better annotations)

gmelodie (changes requested):
Nice one! I just think we should be using `// @Success      200       {object}  actual_object_name` instead of `// @Success      200       {object}  string`. We should start adding those little by little to make docs better

api/v2/storageproviders.go
```diff
@@ -143,6 +143,8 @@ type storageProviderResp struct {
 // @Router       /v2/storage-providers [get]
 func (s *apiV2) handleGetStorageProviders(c echo.Context) error {
 	ctx := context.TODO()
```
gmelodie: I think you forgot to delete this line

gmelodie (changes requested):

gmelodie (approved):

Issue #822, opened 2022-12-28
GET /collections/{col-uuid} Excruciatingly Slow


pythonicode:
**Describe the bug**
Hi there, when I'm running queries to get contents of a specific path of a collection in the Estuary DB, I get highly varied and extremely slow response times ranging from 1 to 15 seconds.

**To Reproduce**

```js
        const url = dir ? `https://api.estuary.tech/collections/${col_id}?dir=${dir}`
                        : `https://api.estuary.tech/collections/${col_id}`;
        const start = Date.now();
        const result = await fetch(url, {
            method: 'GET',
            headers: {
                "Authorization": "Bearer " + process.env.ESTUARY_API_KEY,
            }
        });
        const data = await result.json();
        const end = Date.now();
        console.log(`GET ${url} took ${end - start}ms`);
```

```
GET https://api.estuary.tech/collections/5bf508d7-34c0-4ef7-8853-6dda75be491c?dir=/ took 853ms
GET https://api.estuary.tech/collections/5bf508d7-34c0-4ef7-8853-6dda75be491c?dir=/ took 1395ms
GET https://api.estuary.tech/collections/5bf508d7-34c0-4ef7-8853-6dda75be491c?dir=/ took 8364ms
GET https://api.estuary.tech/collections/5bf508d7-34c0-4ef7-8853-6dda75be491c?dir=/ took 3951ms
GET https://api.estuary.tech/collections/5bf508d7-34c0-4ef7-8853-6dda75be491c?dir=/ took 13101ms
```

**Additional context**

I see that in the following file there is a TODO which seems to imply that there is an unoptimized query being run to fetch the content of the collection in [handlers.go](https://github.com/application-research/estuary/blob/3e0a3d3ccd0e5742d012f099c9e1ba2f0bb01503/api/v1/handlers.go#L3316-L3324)

```go
	// TODO: optimize this a good deal
	var refs []util.ContentWithPath
	if err := s.DB.Model(collections.CollectionRef{}).
		Where("collection = ?", col.ID).
		Joins("left join contents on contents.id = collection_refs.content").
		Select("contents.*, collection_refs.path as path").
		Scan(&refs).Error; err != nil {
		return err
	}
```

I was wondering if it was as simple as improving the query or if the underlying models need to be changed to make this lookup more efficient? I don't know enough about Estuary's underlying system to try to fix this but if anyone has some insight that would be amazing.

Thanks.

pythonicode:
[...]

PR #823, opened 2022-12-28
Optimize GET /collection/{coluuid}

pythonicode:
See #822 

Do not merge, I haven't tested and just brainstorming solutions.


gmelodie (dismissed):
I love this! Thank you so much for doing it. There's a lot of duplicate code in estuary and this helps unify codepaths. The only thing worth pointing out is that I'm not sure this resolves the slowness mentioned on #822 (see [my comment there](https://github.com/application-research/estuary/issues/822#issuecomment-1384253897))

alvin-reyes (changes requested):
Please fix the builds and make sure it passes the checks.

PR #830, opened 2023-01-03
increase rate limit

anjor:
rclone integration tests getting 429ed


neelvirdy:
@anjor you probably want to do this in shuttle.go too. also curious if we know that 200 is the test requirement or if we are just 10x'ing because we have reason to believe it's much too low

anjor:
It was the test requirement. I tested it locally.

anjor:
It was the test requirement. I tested it locally.

LucRoy:
This is the default value right? You can still override this with `rate-limiter` params right? Just for my knowledge :)

anjor:
Correct.

anjor:
The build is failing because dev is broken, not because of the changes made in this PR.

en0ma (approved):

PR #839, opened 2023-01-05
feat: sp endpoints v2 (Revision B)

### LINKED ISSUE #578: Create /storage-providers endpoint
- /public/miners/storage/query/{miner}: description is weird
- remove /public/miners (goes into /miners)
- remove /public/miners/{miner}/failures (goes into miners)
- remove /public/miners/{miner}/storage/query

jcace:
Closes #578 

This is revision B as the original PR was reverted, since it broke V1. (https://github.com/application-research/estuary/pull/831)

In this new PR, V1 has been left completely untouched to avoid anything breaking.


util/auth.go
```diff
@@ -50,3 +52,17 @@ func GetTokenHash(token string) string {
 	// needs to be URL-encodable to send revoke token requests by hash
 	return b64.RawURLEncoding.EncodeToString(tokenHashBytes[:])
 }
+
+func WithUser(f func(echo.Context, *User) error) func(echo.Context) error {
```
jcace, line 56: Copy-paste out of `/api/v1/handlers.go`. In the interest of keeping V1 untouched, I left the old one there. Should we make the refactor on V1, or leave as-is?

util/content.go
```diff
@@ -215,3 +215,38 @@ func GetContent(contentid string, db *gorm.DB, u *User) (Content, error) {
 	}
 	return content, nil
 }
+
+func ConstructDirectoryPath(dir string) (string, error) {
```
jcace: Copy-paste out of `/api/v1/handlers.go`. In the interest of keeping V1 untouched, I left the old one there. Should we make the refactor on V1, or leave as-is?

util/auth.go
neelvirdy: you can do the refactor on V1 to prevent duplicating code here. if we need to change the middleware in the future and only apply it to newer api versions then we can make a separate func in this util and only apply it to the newer api versions.

util/content.go
neelvirdy: you can do the refactor on V1 to prevent duplicating code here

jcace: Good thinking.. Actually this was already refactored in V1 in the mean time - even better!

PR #848, opened 2023-01-09
Add postman testing instructions on CONTRIBUTING.md

gmelodie:


jcace:
I think we should mention that they have to build the solution first `make` in order to generate the swagger files.

For me, I had to delete my existing swagger docs in `docs/` as it would not create new ones. Could we make it automatically overwrite?

jcace:
Also, there is an issue with the `run_tests.bash` script on macOS, it will not run as the `sed` command is slightly different. 

So we should provide instructions - this worked for me

```
brew install gsed
brew info gsed
# <copy the `export` command, i.e:
PATH="/opt/homebrew/opt/gnu-sed/libexec/gnubin:$PATH"
# to persist the change, put it in .rc file
```

https://stackoverflow.com/a/65546076/9504276

alvin-reyes (dismissed):
LGTM

jcace:
LGTM

CONTRIBUTING.md
```diff
@@ -136,3 +136,25 @@ We use swagger to document all of our endpoints and we require developers to ann
...
+
+# Testing
+## Postman
+Run postman tests (these test API changes against swagger documentation)
+
+```
+export APIKEY=your_api_key
+./estuary &
+cd tests/postman
+npm install
```
jcace: maybe should do `npm ci` here instead of `npm install`, to ensure compatibility and avoid overwriting `package-lock.json`

jcace: actually, just noticed there is no `package-lock.json` committed. Any reason for that? If no then it's good your committing it here but we should make sure any other pipeline code does `npm ci` so it won't break in the future

jcace:
add an additional piece of documentation:

To override a test, copy the generated test from `tests/postman/jstests` to `/tests/postman/jsoverrides`, and change the contents. 

To "comment out" or ignore the test, simply replace its contents with:

```
#!/usr/bin/env node
console.log(process.argv[1])
```

Note: the filename is important here.

snissn:
i've been using this little script here running from the estuary repo main folder:

cat ./easysetup.sh
```
rm estuary.db
./estuary setup --username admin --password password  |  sed -e's/.*EST/EST/g' | tail -n1 > /root/estuary/tests/postman/APIKEY
./estuary
```

and then in a second tab running the postman tests via 

```
APIKEY=$(cat APIKEY) ./run_tests.bash
```

jcace (approved):

alvin-reyes (approved):
LGTM

PR #850, opened 2023-01-09
Cleanup postman tests

gmelodie:
- [x] simplify script
- [x] remove  `foo` file
- [ ] better error messages ([todo](https://stackoverflow.com/a/51768316/6032037))


jcace:
looks good! will definitely be good to process the errors better, raw axios dump is hard to figure out what's going on. Not entirely sure where to do that though, any idea?

snissn:
Awesome! Only failure is that we are currently expecting a file named "foo" with content "foo" in the folder. I think we also use it's cid somewhere so can we revert the `git rm foo` part of the commit? @gmelodie

tests/postman/foo
```diff
@@ -1 +0,0 @@
-foo
```
jcace, line 1: need to put this file back as it's used for the tests 🙂

PR #860, opened 2023-01-10
chore: set the collectionref path as nullable

alvin-reyes:
There are existing records in our DB that has `path` that are set to null and when gorm initializes the CollectionRefs, it throws a panic which prevents the API node from starting.

This is a quick fix to set the path column as nullable.


10d9e (approved):
LGTM

Issue #815, opened 2022-12-20
Admin login credentials incompatible with estuary-www


neelvirdy:
See: https://filecoinproject.slack.com/archives/C016APFREQK/p1671516580176459

the salt and the codepath used to generate the credentials via `./estuary setup` differ from how the register user path does it on estuary-www. the /login endpoint only supports the register user path from estuary-www, so logging in as an admin user fails

alvin-reyes:
Yeah this is because of the fixed salt used by the www and the bcrypt-go incompatibility with how bcrypt-js wraps the randomization of the salt, more specifically, bcrypto-go doesn't allow fixed salt.

My solution for this is to have a toggle on the frontend to indicate if the user is admin or not (a checkbox). If it's an admin, it'll pass the "password" over the wire (HTTP) instead of salting it cause this allows the backend to salt it using bcrypt-go in the same manner as when you created the admin account. 

if it's not admin, it uses the same process e.i salt it with a fix value and process the credentials as is.

https://github.com/application-research/estuary-www/pull/68/commits/1827ffa46a3194ad09e95a82a09921a16bdeb234
https://github.com/application-research/estuary-www/pull/66

My suggestion:
-  is either we go with my changes (reopen and merge) - **this is ok.**
- OR remove the "fixed" salt on www and unify both methods - **this is the best.**
- OR modify bcrypt-go to allow specifying a fixed salt to make it consistent with the fixed salt from the www - **I don't recommend this** simply because (1) - fixed salt is not secure and (2) - forking bcrypt-go just to change the access modifier is a stretch for us. We don't want to maintain an encryption library.

neelvirdy:
@alvin-reyes I opened the following PRs to address this issue and https://github.com/application-research/estuary-www/issues/99, which was blocked on admin login working:
- https://github.com/application-research/estuary/pull/861
- https://github.com/application-research/estuary-www/pull/164

PR #861, opened 2023-01-10
IsSession on auth token and validate admin creds on setup

### LINKED ISSUE #99: Treat browser tokens and service tokens separately
1. Browser tokens obtained via login should not appear on the API keys page. They should not be revokable through that UI - and should instead be revoked only via sign out
2. You should not be able to authenticate with an API key. Instead you must login with username/password to get a browser token, which will be revoked on signout. Browser tokens should not be treated like service tokens, which are created via the API keys page, and vice versa

- this prevents you from revoking your session's token via the API key page
- it also prevents you from setting up a token for a service that is also used for your browser session, and then signing out and causing that service token from being unintentionally revoked

cc @jimmylee

neelvirdy:
Along with https://github.com/application-research/estuary-www/pull/164, fixes https://github.com/application-research/estuary-www/issues/99 and https://github.com/application-research/estuary/issues/815

1. Adds `IsSession` field on AuthToken table and populates with `false` for tokens created via api admin page, and `true` for logins/signups. Session tokens will not be rendered in the api admin page so users cannot mix up their service and session tokens. The frontend will also no longer support authenticating via API key to prevent using a service token as a session token. That feature is also unnecessary when admin login is fixed.
2. When running `./estuary setup`, validates admin username and password to be compatible with bcrypt's requirement of 1-32 alphanumeric characters on usernames and 8+ alphanumeric characters on passwords.

image
image

I got a success with a valid creds, but omitted it from the screenshot.

### DIFF
--- a/.github/workflows/postman-tests.yml
+++ b/.github/workflows/postman-tests.yml
@@ -41,7 +41,7 @@ jobs:
         npm install
 
     - name: Setup
-      run: echo "APIKEY=$(./estuary setup --username admin --password password  |  sed -e's/.*EST/EST/g' | tail -n1)" >> $GITHUB_ENV
+      run: echo "APIKEY=$(./estuary setup --username admin --password password1  |  sed -e's/.*EST/EST/g' | tail -n1)" >> $GITHUB_ENV
 
     - name: Run Estuary
       run: |
--- a/main.go
+++ b/main.go
@@ -570,21 +570,6 @@ func main() {
 				if err := db.Create(newUser).Error; err != nil {
 					return fmt.Errorf("admin user creation failed: %w", err)
 				}
-
-				token := "EST" + uuid.New().String() + "ARY"
-				authToken := &util.AuthToken{
-					Token:     token,
-					TokenHash: util.GetTokenHash(token),
-					Label:     TOKEN_LABEL_ADMIN,
-					User:      newUser.ID,
-					Expiry:    time.Now().Add(constants.TokenExpiryDurationAdmin),
-					IsSession: false,
-				}
-				if err := db.Create(authToken).Error; err != nil {
-					return fmt.Errorf("admin token creation failed: %w", err)
-				}
-
-				fmt.Printf("Auth Token: %v\n", authToken.Token)
 				return nil
 			},
 		}, {
--- a/main.go
+++ b/main.go
@@ -575,6 +575,20 @@ func main() {
 				if err := db.Create(newUser).Error; err != nil {
 					return fmt.Errorf("admin user creation failed: %w", err)
 				}
+
+				token := "EST" + uuid.New().String() + "ARY"
+				authToken := &util.AuthToken{
+					Token:     token,
+					TokenHash: util.GetTokenHash(token),
+					Label:     TOKEN_LABEL_ADMIN,
+					User:      newUser.ID,
+					Expiry:    time.Now().Add(constants.TokenExpiryDurationAdmin),
+				}
+				if err := db.Create(authToken).Error; err != nil {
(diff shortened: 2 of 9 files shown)


en0ma:
@neelvirdy how did you come up with the term `session` token for the admin API key?

en0ma:
> Session tokens will not be rendered in the api admin page so users cannot mix up their service and session tokens

Makes sense, since it's a one-time setup key (session?) no need to show it in the UI. My question now is, does this mean the admin after logging in will have to then create a service API key if they need a key?

en0ma:
> When running ./estuary setup, validates admin username and password to be compatible with bcrypt's requirement of 1-32 alphanumeric characters on usernames and 8+ alphanumeric characters on passwords.

Would this play nicely with already existing credentials?

neelvirdy:
> @neelvirdy how did you come up with the term `session` token for the admin API key?

The session field actually doesn't correspond to admin vs regular user. It's purely whether or not the token was generated through login/signup, or if it was a manually provisioned key. In otherwords, session tokens will live in browser cookies exclusively. Service tokens are for all other use cases exclusively (scripts, curls, and other services calling estuary).

The admin creds are relevant to this change because admin login on estuary-www was not working, which forced us to authenticate via API key instead. But, authenticating via API key introduces a risk of mixing up session vs service tokens which defeats some of the purpose of the distinction. This is why the linked estuary-www PR fixes the admin login in addition to making the change to not render session tokens.




> > Session tokens will not be rendered in the api admin page so users cannot mix up their service and session tokens
> 
> Makes sense, since it's a one-time setup key (session?) no need to show it in the UI. My question now is, does this mean the admin after logging in will have to then create a service API key if they need a key?

Login creates a key, so the admin will get their first key simply by logging in with the credentials they ran `./estuary setup` with.

> > When running ./estuary setup, validates admin username and password to be compatible with bcrypt's requirement of 1-32 alphanumeric characters on usernames and 8+ alphanumeric characters on passwords.
> 
> Would this play nicely with already existing credentials?

Bcrypt already throws an error if these requirements are not met, we are simply catching it earlier so the user can't create a user in DB with invalid credentials, only to get a login failure down the road.

main.go
```diff
@@ -565,20 +575,6 @@ func main() {
 				if err := db.Create(newUser).Error; err != nil {
 					return fmt.Errorf("admin user creation failed: %w", err)
 				}
-
```
en0ma: please update the README that running `estuary setup` will no longer return an API key

en0ma: @neelvirdy after the readme is updated, I think this is ready to go

neelvirdy: updated

en0ma (dismissed):
LGTM

.github/workflows/postman-tests.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         npm install
 
     - name: Setup
-      run: echo "APIKEY=$(./estuary setup --username admin --password password  |  sed -e's/.*EST/EST/g' | tail -n1)" >> $GITHUB_ENV
+      run: echo "APIKEY=$(./estuary setup --username admin --password password1  |  sed -e's/.*EST/EST/g' | tail -n1)" >> $GITHUB_ENV
```
en0ma, line 44: @neelvirdy why this change?

en0ma (approved):
LGTM

alvin-reyes (approved):
LGTM

.github/workflows/postman-tests.yml
neelvirdy: password is an invalid password which would now fail the setup cmd

PR #867, opened 2023-01-13
feat: add batch pinning

alvin-reyes:
# Changes

we need to introduce a more efficient way requesting estuary re-pin a CID and this PR introduces a new endpoint that will allow the client to pass an array of IpfsPin object on the request body which will also respond with an array of IpfsPinResponseStatus.

# Verification

image

### Request body

Request body includes an array of `IpfsPin`
```
[
{
    "cid": "bafybeibdzl26proojvu4w5j5kxtvsgzvspiuavfcfr3owajqvehbbdl6z4",
    "name": "1",
    "origins": ["/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"] 

},
{
    "cid": "bafybeibdzl26proojvu4w5j5kxtvsgzvspiuavfcfr3owajqvehbbdl6z4",
    "name": "2",
    "origins": ["/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"] 

},
{
    "cid": "bafybeibdzl26proojvu4w5j5kxtvsgzvspiuavfcfr3owajqvehbbdl6z4",
    "name": "3",
    "origins": ["/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"] 

}]
```

### Response

After processing the request, it returns an array of response that has each of the request Pin status.
```
[
    {
        "requestid": "66",
        "status": "pinning",
        "created": "2023-01-18T14:37:18.780591Z",
        "delegates": [
            "/ip4/10.0.0.83/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/173.35.214.191/tcp/62435/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"
        ],
        "info": {},
        "pin": {
            "cid": "bafybeibdzl26proojvu4w5j5kxtvsgzvspiuavfcfr3owajqvehbbdl6z4",
            "name": "1",
            "origins": [
                "/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"
            ],
            "meta": {}
        }
    },
    {
        "requestid": "67",
        "status": "pinning",
        "created": "2023-01-18T14:37:18.781572Z",
        "delegates": [
            "/ip4/10.0.0.83/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/173.35.214.191/tcp/62435/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"
        ],
        "info": {},
        "pin": {
            "cid": "bafybeibdzl26proojvu4w5j5kxtvsgzvspiuavfcfr3owajqvehbbdl6z4",
            "name": "2",
            "origins": [
                "/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"
            ],
            "meta": {}
        }
    },
    {
        "requestid": "68",
        "status": "pinning",
        "created": "2023-01-18T14:37:18.782547Z",
        "delegates": [
            "/ip4/10.0.0.83/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/173.35.214.191/tcp/62435/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"
        ],
        "info": {},
        "pin": {
            "cid": "bafybeibdzl26proojvu4w5j5kxtvsgzvspiuavfcfr3owajqvehbbdl6z4",
            "name": "3",
            "origins": [
                "/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"
            ],
            "meta": {}
        }
    }
]
```

## Get the stats of the list of content ids

image

### Request
```
[
{
    "content_id": "1"
},
{
    "content_id": "2"
},
{
    "content_id": "3"
}
]
```

### Response
```
[
    {
        "requestid": "1",
        "status": "pinned",
        "created": "2022-11-24T15:53:49.734295Z",
        "delegates": [
            "/ip4/10.0.0.83/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/173.35.214.191/tcp/62435/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"
        ],
        "info": {},
        "pin": {
            "cid": "bafybeif4v2dpyqie2uw5dpygxvxa5jfgzdnkm55cwvdepvwsjwt6qfhzbm",
            "name": "this-is-value.png",
            "origins": [],
            "meta": {}
        }
    },
    {
        "requestid": "2",
        "status": "pinning",
        "created": "2022-11-24T15:53:49.739368Z",
        "delegates": [
            "/ip4/10.0.0.83/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/173.35.214.191/tcp/62435/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"
        ],
        "info": {},
        "pin": {
            "cid": "b",
            "name": "aggregate",
            "origins": [],
            "meta": {}
        }
    },
    {
        "requestid": "3",
        "status": "pinned",
        "created": "2022-11-24T15:53:51.458591Z",
        "delegates": [
            "/ip4/10.0.0.83/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/127.0.0.1/tcp/6744/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y",
            "/ip4/173.35.214.191/tcp/62435/p2p/12D3KooWAV3djoawNqV71NPdNv8pn1ZmNTqYW5fSQxLDKLBVxK6y"
        ],
        "info": {},
        "pin": {
            "cid": "bafybeif4v2dpyqie2uw5dpygxvxa5jfgzdnkm55cwvdepvwsjwt6qfhzbm",
            "name": "this-is-value.png",
            "origins": [],
            "meta": {}
        }
    }
]
```


api/v1/pinning.go
```diff
@@ -234,6 +234,112 @@ func filterForStatusQuery(q *gorm.DB, statuses map[types.PinningStatus]bool) (*g
...
+	}
+
+	var pins []types.IpfsPin
+	if err := c.Bind(&pins); err != nil {
+		return err
+	}
+
+	var pinStatuses []*types.IpfsPinStatusResponse
+	for _, pin := range pins {
+		filename := pin.Name
```
anjor: Instead of duplicating the code here can we delegate the call to handleAddPin?

At the very least extract out the duplicate logic in the same function.

anjor:
This is great and exactly what would benefit club NFT.

Could we also do add a batch version of handleGetPin? I believe they need to call that to get the status of their pins in order to ensure that it is safe to offload the data on their end.

api/v1/pinning.go
gmelodie: Agreed, but I think this new functionality should be on api v2 instead of v1. All new functionalities of API should go into v2

api/v1/pinning.go
```diff
@@ -234,6 +234,112 @@ func filterForStatusQuery(q *gorm.DB, statuses map[types.PinningStatus]bool) (*g
...
+			pathInCollection := collections.Contains(&srchCol, fullPath, s.DB)
+			if pathInCollection && !overwrite {
+				return &util.HttpError{
+					Code:    http.StatusBadRequest,
+					Reason:  util.ERR_CONTENT_IN_COLLECTION,
+					Details: "file already exists in collection, specify 'overwrite=true' to overwrite",
+				}
+			}
+		}
+
```
gmelodie: Use `collections.GetCollection` instead

api/v1/pinning.go
```diff
@@ -234,6 +234,112 @@ func filterForStatusQuery(q *gorm.DB, statuses map[types.PinningStatus]bool) (*g
...
+// @Tags         pinning
+// @Accept		 json
+// @Produce      json
+// @Success      202	{object}  []types.IpfsPinStatusResponse
+// @Failure      500    {object}  util.HttpError
+// @in           202,default  string  Token "token"
+// @Param        pin          body      []types.IpfsPin  true   "Pin Body {[cid:cid, name:name]}"
+// @Param        ignore-dupes  query     string                   false  "Ignore Dupes"
+// @Param        overwrite	   query     string                   false  "Overwrite conflicting files in collections"
+// @Router       /pinning/batched-pins [post]
```
gmelodie: nit: I personally don't like dashes on endpoint names. Maybe on v2 we could only have `POST /pinning/pins` and not two endpoints that essentially do the same thing.

gmelodie (dismissed):
Two major things besides my other comments:
1. This code should only live on v2 as per https://filecoinproject.slack.com/archives/C016APFREQK/p1673634938590189
2. There should be a function that adds pins to collections and such that is called both by /pinning/batched-pins and /pinning/pins, which ensures the same codepath and not needing to keep track of different endpoints when we change something


My preferred solution to both issues is: only have the batched pin add on v2

api/v1/pinning.go
alvin-reyes: it won't be directly compatible with the original `pinning/pins` apiV1. hyphens (dash) on names are the common practice. IMO, its ok to stick with `/pinning/batched-pins`

en0ma:
Nice @alvin-reyes I wished the pinning spec had support for this

en0ma:
> This is great and exactly what would benefit club NFT.
> 
> Could we also do add a batch version of handleGetPin? I believe they need to call that to get the status of their pins in order to ensure that it is safe to offload the data on their end.

I had hoped the pinning spec had support for this in the list pins, but no 😢

api/v1/pinning.go
en0ma: > Maybe on v2 we could only have POST /pinning/pins and not two endpoints that essentially do the same thing.

We can't do this, as it will break the pinning spec compliance

en0ma: > I personally don't like dashes on endpoint names

hyphens on URL is very common

api/v2/pinning.go
```diff
@@ -0,0 +1,172 @@
...
+// @Tags         pinning
+// @Produce      json
+// @Success      200	{object}  []types.IpfsPinStatusResponse
+// @Failure      404	{object}  util.HttpError
+// @Failure      500    {object}  util.HttpError
+// @Param        pin           body      []api.BatchedPinRequest  true   "Pin Body {[content_id:"content_id_to_pin"]}"
+// @Router       /v2/pinning/batched-pins/ [get]
+func (s *apiV2) handleGetBatchedPins(c echo.Context, u *util.User) error {
+
+	var pins []BatchedPinRequest
```
en0ma, line 34: nit: change `pins` -> `request`

api/v2/pinning.go
```diff
@@ -0,0 +1,172 @@
...
+	err := c.Bind(&pins)
+	if err != nil {
+		return err
+	}
+
+	var pinStatuses []*types.IpfsPinStatusResponse
+	for _, pinId := range pins {
```
en0ma, line 34: nit: change `pinId` to `pin`

api/v2/pinning.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	var pins []types.IpfsPin
+	if err := c.Bind(&pins); err != nil {
+		return err
+	}
+
+	var pinStatuses []*types.IpfsPinStatusResponse
+	for _, pin := range pins {
+		filename := pin.Name
```
en0ma: @alvin-reyes can we abstract these lines below and reuse them for both single and batch

api/v2/pinning.go
en0ma: also, can we abstract these lines below and reuse them for both single and batch

anjor: Going a bit further, can we just delegate the whole logic to the single endpoint and have the batch endpoint wrap it in a for loop?

en0ma: @anjor do you mean the batch endpoint should make an HTTP call to the single endpoint? If no, then we are saying the same thing, if yes, I do not think that is a good approach

api/v1/pinning.go
```diff
@@ -271,11 +271,7 @@ func (s *apiV1) handleAddPin(c echo.Context, u *util.User) error {
 
 	var cols []*collections.CollectionRef
 	if c, ok := pin.Meta["collection"].(string); ok && c != "" {
-		var srchCol collections.Collection
-		if err := s.DB.First(&srchCol, "uuid = ? and user_id = ?", c, u.ID).Error; err != nil {
-			return err
-		}
-
+		srchCol, err := collections.GetCollection(c, s.DB, u)
```
gmelodie: nice

api/v2/pinning.go
gmelodie: I think he means having a helper function that both /pinning/pins and /pinning/batched-pins call, only that batched-pins will call it several times.

api/v2/pinning.go
```diff
@@ -0,0 +1,172 @@
...
+			return err
+		}
+
+		if c.QueryParam("ignore-dupes") == "true" {
+			var count int64
+			if err := s.DB.Model(util.Content{}).Where("cid = ? and user_id = ?", obj.Bytes(), u.ID).Count(&count).Error; err != nil {
+				return err
+			}
+			if count > 0 {
+				return c.JSON(302, map[string]string{"message": "content with given cid already preserved"})
```
gmelodie:
```suggestion
				return c.JSON(http.StatusFound, map[string]string{"message": "content with given cid already preserved"})
```

gmelodie (dismissed):

api/v2/pinning.go
en0ma: > also, can we abstract these lines below and reuse them for both single and batch

that is what I meant by this

anjor: Yeah, except having multiple individual helper functions, just having a single one (which could then call separate individual helper functions).

so if the pinning endpoint is 
```
doPin()
```

and the batch pinnning endpoint is
```
for {
 doPin()
}
```


instead of pinning endpoint being
```
doOneThing()
doAnotherThing()
doYetAnotherThing()
```
and batch endpoint being
```
for {
doOneThing()
doAnotherThing()
doYetAnotherThing()

}
```

alvin-reyes: I added a 2 helper functions for this purpose. 1 for pinning and another for getting a pinned cid.

alvin-reyes:
I updated the verification steps to include

1. use of /v2/pinning endpoint
2. get the status of a given list of CIDs

en0ma (approved):
LGTM

PR #884, opened 2023-01-19
fix: change will use uint64 to all obj, obj refs and content Ids

alvin-reyes:
# Changes

This change will use uint64 to all obj, obj refs and content Ids, and all references

# Verification

## Forced to use an ID that's out of uint range

`pinning/pins`
image

`content/add`
image


en0ma (approved):
LGTM

PR #886, opened 2023-01-19
Add via queueing pin op on pinMgr for handleAdd and handleAddCar

neelvirdy:
1. Have `handleAdd` and `handleAddCar` leverage the existing pin op codepaths for pinning contents, instead of maintaining a separate pinning codepath for those handlers
2. Create contents in "queued" state instead of "pinning". Have pinmgr set the status to pinning once a worker is on it
3. Use `onPinStatusUpdate` for all types of pin status updates, not just failures


api/v1/handlers.go
```diff
@@ -548,17 +552,6 @@ func (s *apiV1) handleAdd(c echo.Context, u *util.User) error {
 
 	defer fi.Close()
 
-	replication := s.cfg.Replication
```
neelvirdy: the pin op pinning codepath doesn't support a custom replication setting, so removing this in favor of using what's configured by the content manager

api/v1/handlers.go
```diff
@@ -4947,6 +4922,29 @@ func (s *apiV1) handleShuttleRepinAll(c echo.Context) error {
 	return nil
 }
 
+func (s *apiV1) getOrigins() ([]*peer.AddrInfo, error) {
```
en0ma: the node now has `Origins` method, not sure this is required

api/v1/handlers.go
```diff
@@ -609,23 +602,22 @@ func (s *apiV1) handleAdd(c echo.Context, u *util.User) error {
 		}
 	}
 
-	content, err := s.CM.AddDatabaseTracking(ctx, u, dserv, nd.Cid(), filename, replication)
-	if err != nil {
-		return xerrors.Errorf("encountered problem computing object references: %w", err)
+	if err := util.DumpBlockstoreTo(ctx, s.tracer, bs, s.Node.Blockstore); err != nil {
+		s.log.Errorf("failed to move data from staging to main blockstore: %w", err)
```
en0ma: no need to log here since the return err is still logged downstream. maybe wrap the error with your context message

en0ma: Also, nice fix. We should really have the blocks before creating a db record. I was thinking about doing this later. I have a much better approach, but we can discuss it later after this PR

cmd/estuary-shuttle/main.go
```diff
@@ -1354,29 +1351,20 @@ func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
...
-		Content: contid,
-		Cid:     util.DbCID{CID: nd.Cid()},
-		UserID:  u.ID,
-		Active:  false,
-		Pinning: true,
-	}
-
-	if err := s.DB.Create(pin).Error; err != nil {
-		return err
+	if err := util.DumpBlockstoreTo(ctx, s.Tracer, bs, s.Node.Blockstore); err != nil {
```
en0ma: why not bring create the Estuary content after dumping the blocks

cmd/estuary-shuttle/main.go
```diff
@@ -1497,29 +1482,20 @@ func (s *Shuttle) handleAddCar(c echo.Context, u *User) error {
...
+	origins, err := s.Node.Origins()
 	if err != nil {
-		return xerrors.Errorf("encountered problem computing object references: %w", err)
+		return err
 	}
 
-	if err := util.DumpBlockstoreTo(ctx, s.Tracer, bs, s.Node.Blockstore); err != nil {
-		return xerrors.Errorf("failed to move data from staging to main blockstore: %w", err)
+	if err := s.addPin(ctx, contid, root, u.ID, origins, false); err != nil {
+		log.Errorf("failed to make pin op for content %d for user %d: %s", contid, u.ID, err)
```
en0ma: no need to log here, wrap the error instead with your message context

cmd/estuary-shuttle/main.go
```diff
@@ -1354,29 +1351,20 @@ func (s *Shuttle) handleAdd(c echo.Context, u *User) error {
...
+	origins, err := s.Node.Origins()
 	if err != nil {
-		return xerrors.Errorf("encountered problem computing object references: %w", err)
+		return err
 	}
 
-	if err := util.DumpBlockstoreTo(ctx, s.Tracer, bs, s.Node.Blockstore); err != nil {
-		return xerrors.Errorf("failed to move data from staging to main blockstore: %w", err)
+	if err := s.addPin(ctx, contid, nd.Cid(), u.ID, origins, false); err != nil {
+		log.Errorf("failed to make pin op for content %d for user %d: %s", contid, u.ID, err)
```
en0ma: no need to log here, wrap the error instead with your message context

cmd/estuary-shuttle/main.go
```diff
@@ -1497,29 +1482,20 @@ func (s *Shuttle) handleAddCar(c echo.Context, u *User) error {
...
-		Cid:     util.DbCID{CID: root},
-		UserID:  u.ID,
-		Active:  false,
-		Pinning: true,
-	}
-
-	if err := s.DB.Create(pin).Error; err != nil {
-		return err
+	if err := util.DumpBlockstoreTo(ctx, s.Tracer, bs, s.Node.Blockstore); err != nil {
+		return xerrors.Errorf("failed to move data from staging to main blockstore: %w", err)
```
en0ma: why not bring create the Estuary content after dumping the blocks

api/v1/handlers.go
en0ma: No, this endpoint allows users to submit a replication count, you can pass that into `PinContent`.  Pass default for the other places this is called

api/v1/handlers.go
```diff
@@ -440,17 +440,21 @@ func (s *apiV1) handleAddCar(c echo.Context, u *util.User) error {
...
+	origins, err := s.Node.Origins()
 	if err != nil {
 		return err
 	}
 
-	if err := util.DumpBlockstoreTo(ctx, s.tracer, sbs, s.Node.Blockstore); err != nil {
-		return xerrors.Errorf("failed to move data from staging to main blockstore: %w", err)
+	pinstatus, pinOp, err := s.CM.PinContent(ctx, u.ID, rootCID, filename, nil, origins, 0, nil, false)
+	if err != nil {
+		s.log.Errorf("failed to make pin op for content %d for user %d: %s", pinstatus.Content.ID, u.ID, err)
```
en0ma: no need to log here, return a wrapped error with your context message

api/v1/handlers.go
```diff
@@ -609,23 +602,22 @@ func (s *apiV1) handleAdd(c echo.Context, u *util.User) error {
...
 
-	content, err := s.CM.AddDatabaseTracking(ctx, u, dserv, nd.Cid(), filename, replication)
-	if err != nil {
-		return xerrors.Errorf("encountered problem computing object references: %w", err)
+	if err := util.DumpBlockstoreTo(ctx, s.tracer, bs, s.Node.Blockstore); err != nil {
+		s.log.Errorf("failed to move data from staging to main blockstore: %w", err)
+		return err
 	}
 
-	if col != nil {
```
en0ma, line 617: any reason why you dont want to save the collections anymore?

content/pinning.go
```diff
@@ -180,59 +180,56 @@ func (cm *ContentManager) GetPinOperation(cont util.Content, peers []*peer.AddrI
...
 
-		// if an aggregate zone is failing, zone is stuck
-		// TODO - revisit this later if it is actually happening
-		if c.Aggregate {
-			cm.log.Warnf("zone: %d is stuck, failed to aggregate(pin) on location: %s", c.ID, location)
+	updates := map[string]interface{}{
+		"active":    status == types.PinningStatusPinned,
+		"pinning":   status == types.PinningStatusPinning,
+		"failed":    status == types.PinningStatusFailed,
+		"offloaded": status == types.PinningStatusOffloaded,
```
en0ma: offloaded is a totally different flag. It should not be set here

content/queue/queue.go
```diff
@@ -140,7 +140,7 @@ func (qm *queueManager) processQueue() {
 
 func (qm *queueManager) ToCheck(contID uint64, contSize int64) {
 	// if DisableFilecoinStorage is not enabled, queue content for deal making
-	if !qm.isDisabled && contSize >= qm.dealSizeLimit {
```
en0ma: Any reason for removing the size check? The reason for the check is that we dont want to add content that is not ready for deal-making into the deal queue. Why queue them if they will be ignored downstream? This will only make the deal queue grow larger

content/pinning.go
```diff
@@ -180,59 +180,56 @@ func (cm *ContentManager) GetPinOperation(cont util.Content, peers []*peer.AddrI
...
+		return errors.Wrap(err, "failed to look up content")
+	}
 
-		// if content is already active, ignore it
-		if c.Active {
-			return nil
-		}
+	// if an aggregate zone is failing, zone is stuck
+	// TODO - not sure if this is happening, but we should look (next pr will have a zone status), ignore for now
+	if c.Aggregate && status == types.PinningStatusFailed {
```
en0ma, line 194: replace this with the one below;

```// if an aggregate zone is failing, zone is stuck
	// TODO - revisit this later if it is actually happening
	if c.Aggregate {
		cm.log.Warnf("zone: %d is stuck, failed to aggregate(pin) on location: %s", c.ID, location)

		return cm.db.Model(model.StagingZone{}).Where("id = ?", contID).UpdateColumns(map[string]interface{}{
			"status":  model.ZoneStatusStuck,
			"message": model.ZoneMessageStuck,
		}).Error
	}```

File: shuttle/rpc/rpc.go [hunk 13]
```diff
@@ -273,59 +274,57 @@ func (m *manager) handleRpcGarbageCheck(ctx context.Context, handle string, para
...
 
-		// if an aggregate zone is failing, zone is stuck
-		// TODO - revisit this later if it is actually happening
-		if c.Aggregate {
-			m.log.Warnf("zone: %d is stuck, failed to aggregate(pin) on location: %s", c.ID, location)
+	updates := map[string]interface{}{
+		"active":    status == types.PinningStatusPinned,
+		"pinning":   status == types.PinningStatusPinning,
+		"failed":    status == types.PinningStatusFailed,
+		"offloaded": status == types.PinningStatusOffloaded,
```
en0ma: do not set offloaded flag here

shuttle/rpc/rpc.go
```diff
@@ -273,59 +274,57 @@ func (m *manager) handleRpcGarbageCheck(ctx context.Context, handle string, para
...
+		return errors.Wrap(err, "failed to look up content")
+	}
 
-		// if content is already active, ignore it
-		if c.Active {
-			return nil
-		}
+	// if an aggregate zone is failing, zone is stuck
+	// TODO - not sure if this is happening, but we should look (next pr will have a zone status), ignore for now
+	if c.Aggregate && status == types.PinningStatusFailed {
```
en0ma, line 288: replace this with this below

```
// if an aggregate zone is failing, zone is stuck
	// TODO - revisit this later if it is actually happening
	if c.Aggregate {
		cm.log.Warnf("zone: %d is stuck, failed to aggregate(pin) on location: %s", c.ID, location)

		return cm.db.Model(model.StagingZone{}).Where("id = ?", contID).UpdateColumns(map[string]interface{}{
			"status":  model.ZoneStatusStuck,
			"message": model.ZoneMessageStuck,
		}).Error
	}
```

api/v1/handlers.go
en0ma: can you add a comment that communicates that file uploads block objects will not be created by the pinner, since  `AddDatabaseTracking` is no longer used here

en0ma (approved):
LGTM

PR #897, opened 2023-01-25
Refactor: break up replication into smaller modules

en0ma:
This PR changes the flow from content upload to deal-making. It decomposes the replication logic into separate smaller modules that can all work in an isolated manner making them non-blocking and overall, the platform becomes more stable. 

This also brings the added advantages of non-delayed deal-making as the deal queue is now persistent, with no need to rebuild it whenever the API is restarted.

image2

**Verifications**
- File upload (API and shuttle)
- Cid upload (API and shuttle)
- Splitting
- Aggregation
- Consolidation
- Deal


api/v1/api.go
```diff
@@ -23,21 +26,23 @@ import (
...
-	Node           *node.Node
-	FilClient      *filclient.FilClient
-	Api            api.Gateway
-	CM             *contentmgr.ContentManager
-	StagingMgr     *stagingbs.StagingBSMgr
+	nd             *node.Node
+	fc             *filclient.FilClient
+	api            api.Gateway
+	cm             content.IManager
+	stagingBsMgr   *stagingbs.StagingBSMgr
```
gmelodie, line 35: I like `node`, `fclient`, `contentMgr` better as variable names

10d9e (dismissed):
LGTM

anjor:
i tried to go through it as much as i was able to, but it's too difficult to properly review a PR of this size. The high level thought process and structure makes sense.

config/shuttle.go
```diff
@@ -180,7 +181,7 @@ func NewShuttle(appVersion string) *Shuttle {
 			Queue: QueueEngine{
 				Host:      "",
 				Enabled:   false,
-				Consumers: 5,
+				Consumers: 3,
```
alvin-reyes, line 184: Why did we change this to 3?

pinner/block/block.go
```diff
@@ -0,0 +1,196 @@
...
+
+		if c.Type() == cid.Raw {
+			return nil, nil
+		}
+
+		return util.FilterUnwalkableLinks(node.Links()), nil
+	}, root, cset.Visit, merkledag.Concurrent())
+
+	if err != nil {
+		return err
```
alvin-reyes, line 193: if there's error on walking the block, I think we should just log it and not return err.

content/stagingzone/worker.go
```diff
@@ -0,0 +1,234 @@
...
+			m.log.Debugf("found: %d ready staging zones, for aggregation", len(readyZones))
+
+			for _, z := range readyZones {
+				var zc *util.Content
+				if err := m.db.First(&zc, "id = ?", z.ContID).Error; err != nil {
+					m.log.Warnf("zone %d aggregation failed to get zone content %d for processing - %s", z.ID, z.ContID, err)
+					continue
+				}
+
+				err := m.processStagingZone(ctx, zc, z)
```
alvin-reyes, line 168: Instead of passing the zone as a pointer, we should pass it as the value unless we need to update the reference at some point (which I don't see it being in done on this flow).

alvin-reyes (approved):
A few minor comments but everything looks good to me.
 
i tested the following and it work as expected
- pinning
- content
- collections

Thank you @en0ma. We should target to deploy this as soon as the other reviews are done.

config/shuttle.go
en0ma: given we have too many open db connections, I figured we lower these a bit

content/stagingzone/worker.go
en0ma: ok, I will address this with the pinmgr refactor

pinner/block/block.go
en0ma: we will need to return upstream for splitting and in the coming pinmgr refactor, it will be helpful for pin retries

PR #898, opened 2023-01-25
Bump filclient and pass RemoveUnsealed = false

neelvirdy:
Picks up the `RemoveUnsealed` deal param changes from filclient so estuary deals include a flag in deal proposals that ask SPs to not remove the unsealed copy.


alvin-reyes (approved):
LGTM

Issue #905, opened 2023-01-30
/public/miners - chain_info is null


jcace:
**Describe the bug**
Calling the /public/miners API results in `chain_info` being null 

**Expected behavior**
`chain_info` should be populated for sp's where `suspended` = `false`. 

**Actual behavior**
`chain_info` is null for app SPs

PR #907, opened 2023-01-30
Feat: sp locator task

### LINKED ISSUE #679: Create sp-locator scheduled task to run on shuttles

jcace:
closes #679 , adding a scheduled task to run the PeerPingManager and wire up to the Estuary API, retrieving SP list to ping.

In addition, the following refactors were performed:
- Update API V2 `/storage-providers/` endpoint, fixing nil bug and make it use gothreads (was missed in #908 )
- Extract `ToMultiAddress` and `ToMultiAddresses` functions out of `node.go` and into `utils`, as they're being used in multiple places
- Extract the Shuttle code that makes HTTP requests into a separate file `ShuttleHttpClient`, reducing duplication 
- Rename tracer in `deal/check.go/checkContentDeals`, was missed in #897

### DIFF
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -2399,7 +2399,7 @@ func setupMetrics(metCtx context.Context) Metrics {
 // @Description  This endpoint returns a list of storage providers, sorted by lowest latency to this shuttle
 // @Tags         sp
 // @Produce      json
-// @Success      200  {object}  []peer.ID
+// @Success      200  {object}  PingManyResult
 // @Param        n  query      int  true  "Number of top SPs to return"
 // @Router       /list/{n} [get]
 func (s *Shuttle) handleStorageProviderList(e echo.Context) error {
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -480,7 +480,9 @@ func main() {
 
 		s.Node = nd
 		s.gwayHandler = gateway.NewGatewayHandler(nd.Blockstore)
-		s.PPM = NewPPM(nd)
+		s.PPM = NewPPM(nd, shtc)
+		// TODO @jcace parameterize
+		s.PPM.Run(time.Duration(12) * time.Hour)
 
 		// send a CLI context to lotus that contains only the node "api-url" flag set, so that other flags don't accidentally conflict with lotus cli flags
 		// https://github.com/filecoin-project/lotus/blob/731da455d46cb88ee5de9a70920a2d29dec9365c/cli/util/api.go#L37
--- a/cmd/estuary-shuttle/main.go
+++ b/cmd/estuary-shuttle/main.go
@@ -2395,12 +2395,12 @@ func setupMetrics(metCtx context.Context) Metrics {
 }
 
 // handleStorageProviderList godoc
-// @Summary      Get a list of the top storage providers
-// @Description  This endpoint returns a list of storage providers, sorted by lowest latency to this shuttle
+// @Summary      Get a list of the top storage providers, ordered by latency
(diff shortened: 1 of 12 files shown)


jcace:
Currently blocked from testing by #905 , will test and finish up once that is resolved.

en0ma:
@jcace is this still in draft or you want us to review as is?

jcace:
@en0ma not done yet, soon!

jcace:
Here's an example response from the new Shuttle `/storage-provider/list` endpoint:
```json
[
	{
		"address": "f010617",
		"latency": 33
	},
	{
		"address": "f0135078",
		"latency": 38
	},
	{
		"address": "f01886797",
		"latency": 42
	},
	{
		"address": "f08399",
		"latency": 43
	},
	{
		"address": "f058369",
		"latency": 68
	},
	{
		"address": "f09848",
		"latency": 72
	},
	{
		"address": "f02401",
		"latency": 94
	},
	{
		"address": "f022163",
		"latency": 202
	}
]
```
Screenshot 2023-02-01 at 8 26 19 PM

cmd/estuary-shuttle/main.go
```diff
@@ -1586,32 +1551,11 @@ func (s *Shuttle) shuttleCreateContent(ctx context.Context, uid uint, root cid.C
 		return 0, err
 	}
 
-	scheme := "https"
-	if s.dev {
-		scheme = "http"
-	}
-
-	req, err := http.NewRequest("POST", scheme+"://"+s.estuaryHost+"/shuttle/content/create", bytes.NewReader(data))
+	resp, closer, err := s.HtClient.MakeRequest("POST", "/shuttle/cojntent/create", bytes.NewReader(data), s.shuttleToken)
```
en0ma: `/content` not `cojntent`

cmd/estuary-shuttle/ppm.go
```diff
@@ -0,0 +1,162 @@
...
+
+	return sps, nil
+}
+
+// Ping a list of hosts, returning RTT for each of them. Errors will be ignored, unresponsive hosts will not be returned
+func (ppm *PeerPingManager) PingMany(ctx context.Context, hosts []SpHost) {
+	result := make(PingManyResult, 0)
+
+	// TODO: Batch them in go funcs for faster performance
+	for _, h := range hosts {
```
en0ma, line 105: why not do the pins in goroutines and wait for all result, so it will be faster, just like you did for the chaininfo?

10d9e (dismissed):
Looks good @jcace

cmd/estuary-shuttle/ppm.go
jcace: I'm hesitant to do it since the ping result is sensitive to time - we don't want a situation where it appears like an SP has bad latency when the issue is on our side, due to us running 100+ pings concurrently. I'm not sure if it could cause a queueing delay in seeing the response. Probably OK for a small number but as it scales up we might see that issue. 

I think the solution would be to only do 10 pings at a time, so the code is a little more complex. However, since we only run this routine once every 12h and it takes max. 1 second per provider, we really shouldn't need to optimize it much 🤔 

thoughts?

cmd/estuary-shuttle/main.go
jcace: 🤦‍♂️ thank you!!

cmd/estuary-shuttle/ppm.go
en0ma: thats ok then

en0ma (approved):
LGTM

PR #908, opened 2023-01-30
fix: sp info endpoint - chain info always null

### LINKED ISSUE #905: /public/miners - chain_info is null

jcace:
closes #905 
- Fixed incorrect error handling:
```
if err != nil { // we should do this if err *IS* nil!
	out[i].ChainInfo = ci
}
```
- Also took the opportunity to speed up the querying of ChainInfo, as the request was taking 30+ seconds to run in prod. In local testing, got it down from 10 seconds to under 2 seconds.

### DIFF
--- a/api/v1/handlers.go
+++ b/api/v1/handlers.go
@@ -1795,10 +1795,9 @@ func (s *apiV1) handleAdminGetMiners(c echo.Context) error {
 		out[i].Version = m.Version
 
 		ci, err := s.minerManager.GetMinerChainInfo(ctx, m.Address.Addr)
-		if err != nil {
+		if err == nil {
 			out[i].ChainInfo = ci
 		}
-
 	}
 
 	s.extendedCacher.Add(key, out)
--- a/api/v1/handlers.go
+++ b/api/v1/handlers.go
@@ -1786,6 +1786,7 @@ func (s *apiV1) handleAdminGetMiners(c echo.Context) error {
 	}
 
 	out := make([]minerResp, len(miners))
+	wg := new(sync.WaitGroup)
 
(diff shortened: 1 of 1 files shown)


api/v1/handlers.go
```diff
@@ -1794,13 +1795,19 @@ func (s *apiV1) handleAdminGetMiners(c echo.Context) error {
...
 		out[i].Version = m.Version
 
-		ci, err := s.minerManager.GetMinerChainInfo(ctx, m.Address.Addr)
-		if err != nil {
-			out[i].ChainInfo = ci
-		}
-
+		// Spawn a thread to fetch the Chain Info (Lotus RPC call - takes a few ms)
+		wg.Add(1)
+		go func(w *sync.WaitGroup, addr address.Address, i int) {
```
jcace, line 1800: This will spawn a new go thread per Active Storage Provider in our db. Currently about  ~100 in prod. Each call only takes a few ms to complete.

alvin-reyes: Great stuff man! Thanks for this.

One request-  can we add a verification test in this PR? Just a screenshot of postman response when calling `handleAdminGetMiners` endpoint.

Thank you!

alvin-reyes (approved):
LGTM

jcace:
API Request/response screenshot

Screenshot 2023-01-30 at 6 40 28 PM

PR #943, opened 2023-03-07
run shuttle commp sequentially

en0ma:
This PR will allow shuttles to run commp sequentially, so they don't get overwhelmed (which will cause them to crash).


alvin-reyes (dismissed):
LGTM

alvin-reyes:
Thanks. Lets fix the gosec issues.

alvin-reyes (approved):

PR #945, opened 2023-03-12
fix duplicate queue disk usage

snissn:
-- fix by essentially removing it - on shuttle load we go through the level db with the saved pins and reconstruct the duplicate queue into a map from uint64 to bool and use that while the program is running

### DIFF
--- a/pinner/pinmgr.go
+++ b/pinner/pinmgr.go
@@ -141,7 +141,7 @@ func newPinManager(pinfunc PinFunc, scf PinStatusFunc, opts *PinManagerOpts, log
 	if opts.QueueDataDir == "" {
 		log.Fatal("Deque needs queue data dir")
 	}
-	duplicateGuard := buildDupliceGuardFromPinQueue(opts.QueueDataDir, log)
+	duplicateGuard := buildDuplicateeGuardFromPinQueue(opts.QueueDataDir, log)
 	pinQueue := createDQue(opts.QueueDataDir, log)
 	//we need to have a variable pinQueueCount which keeps track in memory count in the queue
 	//Since the disk dequeue is durable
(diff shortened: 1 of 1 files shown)


10d9e (dismissed):
LGTM

pinner/pinmgr.go
```diff
@@ -143,7 +141,7 @@ func newPinManager(pinfunc PinFunc, scf PinStatusFunc, opts *PinManagerOpts, log
 	if opts.QueueDataDir == "" {
 		log.Fatal("Deque needs queue data dir")
 	}
-
+	duplicateGuard := buildDupliceGuardFromPinQueue(opts.QueueDataDir, log)
```
Zorlin: this should say duplicate, not duplice, unless there's some other meaning @snissn

pinner/pinmgr.go
```diff
@@ -297,33 +293,29 @@ func (pm *PinManager) closeQueueDataStructures() {
...
-	return db
+	iter := db.NewIterator(nil, nil)
+	for iter.Next() {
+		entry, err := decodeMsgPack(iter.Value())
+		if err != nil {
+			continue
+		}
+		ret[entry.ContId] = true
+	}
+	db.Close()
```
Zorlin: this is throwing CVE errors as we're not checking if we successfully close the database

Zorlin (changes requested):
I will be deploying this on shuttle-9 as-is, but please fix my two changes before we deploy it to all shuttles. Thanks for the awesome turnaround time!

pinner/pinmgr.go
snissn: lol no just bad @ spell

snissn: https://github.com/application-research/estuary/pull/945/commits/7d1e3147a4d77ec667641511bde2b23837f29daa

snissn: fixed

Zorlin (commented):
Duplicatee? Close ;)

Zorlin (approved):
Looks great, thank you!!!

Zorlin:
LGTM. Merged.

PR #992, opened 2023-07-14
fix the pinning api filtering

ohmpatel1997:
- This PR fixes the status based filtering on the list pins api.

### DIFF
--- a/api/v1/pinning.go
+++ b/api/v1/pinning.go
@@ -10,14 +10,15 @@ import (
 
 	"github.com/application-research/estuary/pinner"
 
-	"github.com/application-research/estuary/model"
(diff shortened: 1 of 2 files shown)


LucRoy:
I'm wondering if we shouldn't add either a new endpoint or new query param for this. We want to make sure we don't break any backward compatibility. @alvin-reyes what do you think?

ohmpatel1997:
> make sure we don't break any backw

I think the previous logic was flawed in itself. So even if someone was using it must be giving them the wrong results. also it wont break any existing api's schema, since we are not chaning any query param types.

ohmpatel1997:
Hi @LucRoy  can we please get this PR in?

